PR #84, opened 2014-10-28
changing default to on

bdolor:
On new book creation, setting the default to 'on' will allow more people to derive benefit from the built-in license module. Users would still have to choose a license in 'Book Info' as demonstrated in the youtube video https://www.youtube.com/watch?v=3Zz6fJm9cWQ&feature=youtu.be 

Checking both the 'display the copyright license' checkbox in 'Appearance->Theme Options' and choosing a license in 'Book Info' are what is required to display creative commons copyright information on the footer of every web page and make it through the export routines.

### DIFF
--- a/themes-book/pressbooks-book/functions.php
+++ b/themes-book/pressbooks-book/functions.php
@@ -529,7 +529,7 @@ function pressbooks_theme_copyright_license_callback( $args ) {
 	$options = get_option( 'pressbooks_theme_options_global' );
 
 	if ( ! isset( $options['copyright_license'] ) ) {
-		$options['copyright_license'] = 0;
+		$options['copyright_license'] = 1;
 	}
 	
 	$html = '<input type="checkbox" id="copyright_license" name="pressbooks_theme_options_global[copyright_license]" value="1" ' . checked( 1, $options['copyright_license'], false ) . '/>';


greatislander:
Thoughts on this, @hughmcguire?

hughmcguire:
@boldor I will commit to reviewing this. Currently its implementation seems unclear to me (and Pressbooks users). Possibly just a question of clarifying the explanatory language.

(also, to make @greatislander unhappy, if you have selected a non-English language for your book, then the copyright message comes out in a garble of [at least] two languages, one of them possibly English).

bdolor:
The description of the functionality would stand to benefit from using better explanatory language for the humans, I agree. Suggestions welcome; I would be happy to modify it. 

While attempts are made to match the language of the copyright statement with the declared language of the book, issues with incomplete translations stems from us running up against limitations of the API supplied by creative commons. I don't know what to suggest here. They are aware of the limitation: 

"All methods can return localized text. Note that in some cases a locale will have a mix of English and the native language – English is currently the fallback for all locales and is used in the case of partial translations."

https://api.creativecommons.org/docs/readme_15.html#locales

Although, documentation also states: "Future language translations may be added at any time in the future without breaking 1.0 compatibility."

Nevertheless, it is useful to us as feature despite its limitations and I feel that relying on creative commons to supply a creative commons license is wise. They may be open to specific feature requests: https://github.com/creativecommons/cc.api 

The list of available 'locales' can be seen at the bottom of the xml document here: 

http://api.creativecommons.org/rest/1.5/license/standard

greatislander:
@hughmcguire Giving you another nudge on this.

greatislander:
@bdolor In taking another look at this, the call to the Creative Commons API implemented in #62 seems to be failing. Can you look into this?

bdolor:
This looks to be temporary, but this type of scenario speaks to looking at what kind of backup can be built into the feature should the web service be down for maintenance again:
 
screen shot 2015-05-05 at 9 10 07 am

bdolor:
update: I've opened up a ticket with creative commons and they are aware of the issue and are actively working to fix it. They provided this link to check on the status: http://status.creativecommons.org/

bdolor:
...and they're back - http://status.creativecommons.org/

greatislander:
@bdolor @hughmcguire I'm wondering if we could address some of the confusion around this by splitting this functionality in two:

1. Global options would have a checkbox to control the display license module information in exported books.
2. Web options would have a checkbox to control the display of license module information in webpage footers.

Right now having this checkbox in Global options is a bit ambiguous as it simply refers to "Display[ing] the copyright license". If a user has entered their own custom license it would be displayed in exports already so I can see where @hughmcguire's confusion comes from. (Also: it might make it clearer to say "license" instead of "copyright license"?) Thoughts? Sorry to perhaps complicate this more but after playing around with the license module I do find it a bit confusing and I think these changes might clarify matters.

bdolor:
A couple of considerations that would result from implementing the suggestion: 
 1. Global options currently states that it applies to 'webbook, PDF and ebook exports'. The functionality of 'copyright license' is such that it does apply to all of those formats. Taking some of the functionality out of 'global options' and putting it in 'web options' runs the risk of being misleading, since checking the box in 'global options' would now only apply to PDF and ebook. 
 2. From a practical perspective, changing these option values now would require that we go through all the books in our collection and modify the theme options to recreate the behaviour that is currently set. An inconvenience at best, but I doubt we'd be the only ones. 

I looked back at the original problem — is the answer just better labeling? Including a more thorough description of what to expect with the functionality? 

In my mind there are two very different pieces of functionality with little overlap - the custom copyright notice (a text field that only applies to the book) , and the copyright license(Title, Author, Source, License that can be applied to an entire book and/or overridden for individual pages). 

For our use case, the custom copyright notice is often used to describe the exceptions to the license, an opportunity provided by the default text 'except where otherwise noted'.

Whereas the copyright license functionality describes copyright while adhering to [attribution best practices](https://wiki.creativecommons.org/Best_practices_for_attribution), which can't be guaranteed with a textbox, nor can it be overridden if the page license is different from a book license. The objective of the copyright license functionality was to offer a way to express attribution more formally, automatically and be overridden at the page level.  

 1. custom copyright notice (applies only to the book, can describe exceptions or whatever you want)
 2. copyright license (Following best practices for attribution — automatic generation of title, author, source, license at the book and page level)

greatislander:
I take your point re: changing options. I think clarifying the label would be helpful. Something like: "Display the copyright license (if set) in webbook footers and on PDF and ebook copyright pages"

greatislander:
That's a bit of a mouthful, I realize… thoughts, @hughmcguire?

It also occurs to me that this PR could resolve #108.

connerbw:
This pull request is 1 line of code and has been lingering for few months now. 

Can I suggest we change:

```php
if ( ! isset( $options['copyright_license'] ) ) {
    $options['copyright_license'] = 0;
}
```

To something like:

```php
if ( ! isset( $options['copyright_license'] ) ) {
    $options['copyright_license'] = apply_filters( 'pb_copyright_license_default', 0 );
}
```

Then, if the user has PB Textbook installed the default can be hooked/changed from there?

Thanks

bdolor:
If the user has PBT installed, the default is already modified on new book creation:

https://github.com/BCcampus/pressbooks-textbook/blob/9fe2d46dc225ed073990316d3ff5db30f0289305/pressbooks-textbook.php#L362

connerbw:
Alright! So... Won't fix?

greatislander:
Going to close this. I think this feature makes sense for textbook users but we haven't established that having it on by default is the best for core Pressbooks. Thanks for the input, all.

PR #88, opened 2015-01-06
modify logic for epub3 display, labels for api and epub3

bdolor:
Changing labels to 'Beta' which more accurately describes the state of both pieces of functionality. To encourage Beta testing of these two pieces of functionality, the logic for displaying EPUB3 output is now exclusive (of pressbooks.com) rather than inclusive of a handful of domains. This will keep EPUB3 off of pressbooks.com and make it available to others who will now have the opportunity to provide feedback which can inform further development.


greatislander:
Thanks @bdolor!

PR #89, opened 2015-01-09
add child theme support for toc collapse

bdolor:
`get_template_directory_ui()` will look for the js script in luther (parent theme) directory, instead of in a child theme's stylesheet directory (where it won't find it). Gives collapsing TOC functionality to child themes without having to copy the script to other directories.


greatislander:
Thanks!

Issue #91, opened 2015-01-22
catalogue page and webkit browsers


bdolor:
problem: css height isn't calculated properly in webkit browsers on `<div class="catalog-content' id='catalogue-content'>` for the initial page load. 

to reproduce: add a bunch of books to your catalogue, then view the page in safari or chrome. Book cover images will overlap each other:

![screen shot 2015-01-22 at 8 51 39 am](https://cloud.githubusercontent.com/assets/2048170/5860110/ffd414e4-a213-11e4-83da-82998649853b.png)

Toggling between an earlier commit `git checkout 448834b` and the latest commit `git checkout 7521ddfe` indicates webkit browsers don't like the js change introduced with this commit https://github.com/pressbooks/pressbooks/commit/ff794ba2091aaf3cd353e96110a51a9fffe9f2bf

I'll try and dive into this later today, but if you have a quick fix - feel free.

greatislander:
Thanks for spotting this, @bdolor. I had to replace the library we were using before because it was incompatible with GPL2—thought I had addressed this issue with the new one but evidently I didn't! If you can get to this it would be much appreciated, otherwise I'll take a look.

PR #92, opened 2015-01-22
fix webkit issue with catalogue page

bdolor:
I've tested this fix in the latest of FF, Safari and Chrome. fixes https://github.com/pressbooks/pressbooks/issues/91

Observations: `$container.equalizer()` wasn't firing inside `$container.isotope( 'on', 'layoutComplete'...`, in fact nothing was. Not even `console.log('is this thing on?');` Perhaps because `layoutComplete` is defaulted to true, `$container.equalizer()` didn't need binding to that event, explicitly. Just a guess.


greatislander:
Thanks!

PR #93, opened 2015-02-05
add web only options tab

bdolor:
TOC Collapse is an appearance option that affects only webbook. A new category of theme options affecting only web appearance is required. Leaving it under Global Options can be misleading since it does not affect either pdf or ebook output. 

screen shot 2015-02-05 at 1 50 27 pm


greatislander:
Thanks @bdolor.

PR #97, opened 2015-03-06
fix docx import, look for images in alternate directory

bdolor:
@hughmcguire - as discussed, this should resolve some images not being pulled in with the docx import. It was necessary to cast the image path as a string and to look for the images in an alternate directory. I've tested it minimally (with one or two docs) and seems to resolve it. Please let me know if it needs further tweaking.


hughmcguire:
thanks @bdolor ... @greatislander can you pull this into dev & I'll test?

greatislander:
Looks good to me— merged. Thanks @bdolor.

PR #110, opened 2015-03-30
add support for internal links on subdirectory installations

bdolor:
The quickest and dirtiest way to add support for internal linking on export routines for subdirectory installs as outlined in detail here: http://pressbookstest.trinfocafe.org/book/chapter/chapter-1/

Admittedly, I don't love this solution, but it works. Question for @greatislander — what risks are you aware of if we introduced a more robust function at this point in the routine, similar to what we see later, for instance, in epub export? https://github.com/pressbooks/pressbooks/blob/dev/includes/modules/export/epub/class-pb-epub201.php#L1819

### DIFF
--- a/includes/modules/export/xhtml/class-pb-xhtml11.php
+++ b/includes/modules/export/xhtml/class-pb-xhtml11.php
@@ -412,7 +412,9 @@ class Xhtml11 extends Export {
 	 * @return string
 	 */
 	protected function fixInternalLinks( $content ) {
-
+		// takes care of PB subdirectory installations of PB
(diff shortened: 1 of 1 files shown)


JackDougherty:
@bdolor  -- Many thanks for thinking up a potential solution to this problem of internal links for books created as subdirectories. Will wait to see what @greatislander says. But this sure beats rebuilding my entire site (or your site) in subdomain format.

greatislander:
Hi @bdolor—if you would like to develop a more robust routine to process internal links for PDF output, I see no risk in this. The `fixInternalLinks` function is only used within the XHTML generation routine, so PDF/XHTML exports are the only place where this will have any effect. As long as it doesn't break existing functionality.

That said, I would be happy to merge in this solution for the time being. Let me know what you think is best.

bdolor:
If this solves a problem now, then let's move ahead with the merge. Making `fixInternalLinks` more robust  will be noted as an enhancement for future development.

greatislander:
Sounds good, @bdolor.

Issue #109, opened 2015-03-29
SILE as PDF generator backend?


skreutzer:
You guys might consider to look into http://www.sile-typesetter.org (https://www.youtube.com/watch?v=5BIP_N9qQm4) as a way to get rid of the PrinceXML dependency, as the latter is proprietary software. Unfortunately, SILE is licensed under the X11 license, but as Pressbooks is under GPLv2, I guess you're not bothered too much regarding it's shortcomings in protecting users rights.

There are other freely licensed PDF generator backends as well, this is the latest one I got notified about.

bdolor:
For the last few months @jgraham909 and I have been working on an [alternative PDF backend](https://github.com/bdolor/pressbooks/tree/pdf) using [mPDF](http://www.mpdf1.com/mpdf/index.php) as a backend. This will provide an open source alternative to Prince — Prince is fantastic, but as mentioned, proprietary, which can be a barrier. I'll be submitting a Pull Request by the end of the week. mPDF is also released under [GPL](http://www.mpdf1.com/mpdf/index.php?page=Licence)

skreutzer:
Thank you very much! I'll look into mPDF as soon as I can find the time.

greatislander:
@bdolor @skreutzer Are we comfortable closing this issue now that #111 has been merged in? I’d like to focus development resources on enhancing mPDF as our open source PDF export option given the work that's already gone into it on @bdolor and @jgraham909's part.

bdolor:
If providing an alternative to PrinceXML was the intended goal, closing this sounds reasonable to me.

greatislander:
Thanks @bdolor.

PR #111, opened 2015-04-02
Open source alternative for generating PDF

bdolor:
mPDF is being used as a back end which is released under GPL. 
takes care of https://github.com/pressbooks/pressbooks/issues/109

I consider this PR a minimal viable product. It demonstrates that it is possible and though it is unfinished, it is at an acceptable level for evaluation. Maintainability of the code could be vastly improved upon if it was re-written to parse the XHTML output of the XHTML routine, instead of its current implementation. Indeed, this is recommended to be a priority moving forward, so as to not divorce itself from continuing development of PB or requiring duplication of code. Unlike Prince, mPDF doesn't like being given large XHTML files, and better control can be gained by giving it smaller HTML chunks.  

All open source themes will use the default parent 'mpdf export' theme until overridden by their own child theme export file. This will be another development task moving forward. 

It comes with it's own set of theme options, which are unique. Tying them into Prince theme options wasn't possible because the feature set and requirements are different. The only global theme option that hasn't been integrated is the 'Parse Sections - enable a two level TOC'

The export option is designed to not show up on pressbooks.com and also be the only PDF output option displayed to users should PrinceXML not be configured/installed on the system it finds itself on. Should PrinceXML be installed, both PDF output options are available to the user. 

screen shot 2015-04-02 at 1 55 03 pm


wexlerj:
I agree with this feature being a priority as PDF is still a big file standard, especially for print obviously.
We used XPDF about 12 years ago at Autodesk (the media and entertainment division), because we were using IRIX-based SGI machines, and then transitioned to Linux/IBM etc.

Since I am new to making any real contributions to the source code for PressBooks, I will research mPDF some more, and then add further comments when I understand how the attempt is making made to switch out Prince for mPDF.

My initial thought is to see if mPDF is the right backend for the conversion as I never heard of it.

wexlerj:
I did some research and thinking, and I don`t get why all this extra post-processing has to happen when PressBooks already outputs to HTML nicely in a browser.

Why can`t we just use the browser`s rendering engine to output PDFs after it is rendered by the browser for web.

I realize this solution will be convoluted for Internet Exploer (but who knows what`s in store for Window in general) but with Google Chrome and Mozilla FireFox, since they must use similar PDF conversion code, why not piggy-back off this, simplify it, and make the feature work so we are no longer reliant on the expensive and bloated Prince stuff.

hughmcguire:
The structure of the web output is slightly different from the xHTML structure used to generate the book outputs. In particular, it is less bookish.

So, for PDF output, we would want to stick with rendering the xHTML ... 

Pressbooks is built so that you get: 
Presssbooks --> xHTML --> [output engine] --> PDF 

For the plugin, we;re agnostic about what the output engine would be ... but not agnostic about the process by which the output is generated.

Hugh.

nciske:
> Why cant we just use the browsers rendering engine to output PDFs after it is rendered by the browser for web.

1. Because then books might look different depending on the browser that "rendered" it. PDFs are generally expected to look the same no matter what -- not have subtle differences based on how you access it.
2. Support for print/page specific CSS is very, very spotty. PressBooks uses these heavily, hence why Prince is used -- it supports them very well.
3. I don't know of any browsers that support outputting PDFs natively -- all rely on a 3rd party (OS, plugin, etc).
4. Using the "print to PDF" function of your OS, or an Acrobat plugin, etc. will result in varying outputs (e.g. colors spaces may shift, little control over image resolution/compression, limited to no control over PDF profile).
5. mPDF (or similar) as a drop-in replacement for Prince is far saner solution than juggling browser bugs as you can control the output.

My 2 cents.

bdolor:
@greatislander  - could you let me know where you're at with this PR? Thank you!

greatislander:
Hi @bdolor—really waiting for the all-clear from @hughmcguire before merging into dev. It looks good to me thus far.

hughmcguire:
OK @greatislander @bdolor ... this looks fine to me... (though @greatislander note that it should not appear in Pressbooks.com OR client installs (<-- !).

hughmcguire:
though generally I am worried about diverging dev trajectories for PDF outputs ... that is, new features now have a new codebase to be supported in.

@bdolor do you intend to focus on mPDF for your future PDF dev work whatever that may be?

bdolor:
It is a concern of mine as well. From a maintainability perspective it can't stay. Moving it to parse from the XHTML output needs to happen. The decision to merge this PR now and make changes to it later will depend on how important timelines are to get this out the door. There seems to be a lot of interest from the open source community in this feature. If it meets minimal requirements as is, this approach still aligns with an iterative development process. 

If there is clear direction on where development needs to go and a commitment from PB to include it in the code base,  I wonder if that could also translate into the community contributing to the development of this feature — crazier things have happened.

bdolor:
That is to say, this feature is important enough to our stakeholders that my involvement with this, in some capacity, is inevitable.

greatislander:
I am prepared to bring this into the development branch although I would like to establish a roadmap for moving to XHTML output parsing so that mPDF can track to the rest of the exports as we develop them. A tangential question, @bdolor: to your knowledge is there a standard means by which we are excluding certain features from pressbooks.com? I know you've worked with some such experimental features so I thought you might know offhand. If there isn't, I'd like to build a function that we can use everywhere in the plugin for this purpose.

@hughmcguire: shall I proceed with the merge?

bdolor:
The 'standard' thus far has been only to exclude by [host name](https://github.com/pressbooks/pressbooks/blob/dev/admin/templates/export.php#L126). So far there has only been one host name to exclude though, should the number of host names grow the need for a central function is paramount.

hughmcguire:
OK with me if OK with you Ned -- as long as this doesn't go into Pressbooks.com or Pressbooks client installs.

PR #113, opened 2015-04-14
Fixed variable-related warnings on RESTful API, when debugging mode is enabled

julienCXX:
When using the RESTful API (calling URL `/path/to/wp/installation/api/v1/books/2/?titles=a`), I was given some messages from PHP:
```
<br />
<b>Notice</b>:  Undefined offset: 2 in <b>/path/to/wp/installation/wp-content/plugins/pressbooks/includes/modules/api_v1/books/class-pb-booksapi.php</b> on line <b>446</b><br />
<br />
<b>Notice</b>:  Undefined variable: front_matter in <b>/path/to/wp/installation/wp-content/plugins/pressbooks/includes/modules/api_v1/books/class-pb-booksapi.php</b> on line <b>282</b><br />
```

I removed the mentioned variables, as they do not appear to be useful for anything. Even worse, `$front_matter` appears to delete data related to front matter chapters.
This breaks the inner workings of some plug-ins, like [Textbook](https://github.com/BCcampus/pressbooks-textbook)'s *Search and Import* feature.

Ironically, this only happens when the WordPress debugging mode is enabled, which sounds like this feature was developed without using it…

### DIFF
--- a/includes/modules/api_v1/books/class-pb-booksapi.php
+++ b/includes/modules/api_v1/books/class-pb-booksapi.php
@@ -279,8 +279,7 @@ class BooksApi extends Api {
 		foreach ( $book['front-matter'] as $fm ){
 			$chapters[$fm['post_id']] = $fm; 
 		}
-		$chapters = $front_matter;
-		
+
 		// parts
 		for ( $i = 0; $i < $parts_count; $i ++ ) {
 			// chapters
@@ -443,7 +442,6 @@ class BooksApi extends Api {
 			if ( ! in_array( $args['id'], $this->public_books ) ) {
 				return $this->apiErrors( 'empty' );
 			}
-			$book[$args['id']];
 			$book[$args['id']]['book_id'] = $args['id'];
 			$book[$args['id']]['book_url'] = get_blogaddress_by_id( $args['id'] );
 			$book[$args['id']]['book_meta'] = \PressBooks\Book::getBookInformation( intval( $args['id'] ) );


bdolor:
This looks fine to me.
 
I'm trying to remember if there was a reason I excluded front matter content from the results, but looking at how it is written, I'm going to assume for now that `$chapters = $front_matter` is an oversight rather than something intentional. 

I can't think of how building something with debugging disabled could benefit a developer or even be possible, but I am definitely guilty of ignoring PHP `Notice` messages on occasion. If it's not `Fatal` or `Warning` I tend to move on to other priorities and convince myself I'll come back to it later. Nevertheless, this is a valuable contribution — so thanks! 

It is worth nothing that another source of the problem with *Search and Import* not functioning however, is an extra `$` on [line 181 in PBT] (https://github.com/BCcampus/pressbooks-textbook/blob/dev/includes/modules/search/class-pbt-apisearch.php#L181)

`PHP Catchable fatal error:  Object of class WP_Error could not be converted to string in /Users/bdolor/Sites/pressbooks/wp-content/plugins/pressbooks-textbook/includes/modules/search/class-pbt-apisearch.php on line 181`

That's a doozy — If you're up for a Pull Request on that repo, I will happily merge.

julienCXX:
Well, I fixed the said issue in PressBooks.

As an advice about paying attention to `Notice` messages, I would say to do it at least on any component that prints data designed to be read by another component (like a REST API).

bdolor:
One of the great things about open source software is that anyone can review code. This sometimes means that you find areas to contribute where others haven't been able to put their attention to for any number of reasons, most likely this has to do with priorities. Fostering a community that everybody enjoys being a part of, often means being careful with the assumptions you make and refraining from offering unsolicited advice to people you don't know and who most likely won't benefit from it. A better assumption, for instance, is that each of us is a busy, competent developer and that we all have the same goals. That is, make it your goal to contribute good code and leave it at that. Thanks again, for your code contribution.

julienCXX:
Well, I have to say that I am not really used to the GitHub Pull Request system (and contributions to projects maintained by people I do not know IRL), so, that can explain my previous comment.

Have I to do anything else to have my contribution merged?

bdolor:
thanks! @julienCXX

hughmcguire:
hi @bdolor & @julienCXX ...first, thanks Julien for the contribution ... second, my guess is there was a bit lost in translation in the exchange above...Julien, I guess English is your second, or 3rd language -- my experience (for instance, at LibriVox) is that when you combine the contextless communication of text message boards like this, with slight idiosyncrasies in 2nd languages, you get awkward phrasing that sometimes doesn't sound to the reader's ears what the writer intended ...

anyway, thanks for this!

greatislander:
@julienCXX Thanks!

Issue #79, opened 2014-09-26
add_magic_quotes for wp_insert_post and wp_update_post


lukaiser:
Hey guys

I just found a bug in the code of my fork. It sometimes removed all the backslashes from a post. What is kind of annoying if you have a lot of formulas.

What I found, is that you need to call add_magic_quotes bevor you wp_insert_post or wp_update_post.
https://core.trac.wordpress.org/ticket/21767
There is at least one wp_update_post and several wp_insert_post not doing that in pressbooks (manly in the import modules). The function calls are in parts of the code I haven't really looked at, so I don't know if it ever affects any thing. But I think it happened to me while importing a WXR. I don't have the time to think it trough or test it. I will keep an eye on it if it ever happens again. But maybe somebody can test it or we add it anyway. It should not do any harm.

I guess one test case is to import a WXR with latex formulas with backslashes in it.

PR #116, opened 2015-04-19
Fix for #79

greatislander:
Hi @bdolor—as someone with a more thorough knowledge of the import routines than me, do you see any issues in adding `add_magic_quotes` to the `wp_insert_post` function  so as to address @lukaiser's issue #79? Let me know and I can merge or not accordingly.


bdolor:
Adding this benefits preserving LaTeX equations on import which are painful to have to re-create. No risks posed as far as I can tell. Seems to have recently cropped up on other forks of PB as well https://github.com/lumenlearning/candela/issues/23

greatislander:
Fantastic, thanks @bdolor! Merging now.

PR #118, opened 2015-04-21
add esc_url to add/remove_query_arg

bdolor:
https://blog.sucuri.net/2015/04/security-advisory-xss-vulnerability-affecting-multiple-wordpress-plugins.html 

https://yoast.com/coordinated-security-release/

### DIFF
--- a/admin/class-pb-catalog-list-table.php
+++ b/admin/class-pb-catalog-list-table.php
@@ -100,11 +100,11 @@ class Catalog_List_Table extends \WP_List_Table {
 	function column_status( $item ) {
 
 		$add_url = sprintf( ' ?page=%s&action=%s&ID=%s', $_REQUEST['page'], 'add', $item['ID'] );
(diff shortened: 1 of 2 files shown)


greatislander:
Thanks for being so prompt with this, @bdolor!

bdolor:
No problem, though in my haste, I wasn't thorough in examining the difference uses: 

https://make.wordpress.org/plugins/2015/04/20/fixing-add_query_arg-and-remove_query_arg-usage/

"The easiest way to fix this in your plugin is to escape the output of add_query_arg() and remove_query_arg(). When it’s being printed to a page (for example as a link), you should use esc_url(). When it’s being used in HTTP headers or as part of a HTTP request (for example, as part of a location redirect header or in a wp_remote_get() call), you should use esc_url_raw()."

perhaps you could verify that I've applied the patch correctly. 

thanks @greatislander

greatislander:
Ah okay. Looks like the change in `pb-latex-admin.php` should use `esc_url_raw()`. I'll change accordingly.

PR #126, opened 2015-05-20
Fallbacks for getBookInformation()

greatislander:
In instances where no book information has been entered, the `getBookInformation()` function and its theme alias, `pb_get_book_information()`, return empty arrays. This pull request instead returns a simple array with a title (the book/blog name), an author (the book/blog administrator's display name), and the default cover image.

### DIFF
--- a/includes/class-pb-book.php
+++ b/includes/class-pb-book.php
@@ -112,6 +112,7 @@ class Book {
 			$book_information[$key] = $val;
 		}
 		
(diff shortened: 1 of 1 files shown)


greatislander:
@hughmcguire Let me know what you think of this. Addresses issues where no book information has been entered as the current function returns a blank title, author and no cover image (the default cover image is only set once book information has been edited).

hughmcguire:
looks good to me @greatislander

PR #127, opened 2015-05-21
add specificity, preference for PB generated tags

bdolor:
Small tweaks to the XHTML import to better target content contained within HTML elements generated by PB. 

If the XHTML import source file is a PB site, this improves the title that gets pulled in, and also the content area.


hughmcguire:
seems like a good idea to me -- over to you ned.

PR #129, opened 2015-05-25
unset variable to accurately represent which chapters belong to which…

bdolor:
not destroying the value of this variable at the end of each loop has the unintended affect of adding chapters to parts to which they do not belong. The last part in a multi-part book is represented as having all the chapters in the entire book. This small change rectifies that.


greatislander:
Thanks, @bdolor.

PR #143, opened 2015-06-10
update readme.txt with php5-xsl notice

jflowers45:
Ran into this issue while trying to set up a self hosted instance on Ubuntu on EC2


greatislander:
Thanks @jflowers45!

PR #144, opened 2015-06-11
fix for about the author style, accessibility button

bdolor:
The content in this div class would remain at the enlarged font-size regardless of whether the accessibility button was enabled or disabled. This fixes that problem.


BakingSoda:
Thanks @bdolor

Issue #107, opened 2015-03-27
Add option to deactivate social media buttons


programmieraffe:
Hey there,

thanks for the great work on pressbooks. An option to deactivate the social media buttons beneath a chapter would be very good, because in some cases it is a privacy issue.

Thanks in advance!

Best regards,
Matthias

hughmcguire:
we're in the process of revamping the theme options pages ... so not bad timing.

PR #146, opened 2015-06-12
enable social media integration as an option

bdolor:
A potential fix for https://github.com/pressbooks/pressbooks/issues/107

Default setting is to 'enable' so no change will come to books that currently use it, however, this now becomes an option for those who don't want it. 

Passes tests with 'Ghostery' and 'Privacy Badger' Add-ons for Firefox


greatislander:
Thanks @bdolor—this looks great.

PR #151, opened 2015-06-19
include social media buttons if variable has yet to be set

bdolor:
If users have not yet visited Appearance -> Theme Options -> Web Options and hit 'save', the variable that the condition relies on for displaying buttons will not have been set. This more permissive conditional statement will display social media buttons without requiring any action on the part of the users.


greatislander:
Thanks!

PR #152, opened 2015-06-22
Fix for max import size

jflowers45:
Hi,
  Apologies in advance if I've broken any etiquette rules with an unsolicited pull request for code.  If you prefer to have an open issue first, please let me know!

We are hosting our own pressbooks instance and ran into an issue with importing ePubs.  Specifically, the file size listed on the import page only pays attention to php's upload_max_filesize setting.  We needed to increase both upload_max_filesize and post_max_filesize to allow imports over the default 2MB

So, to accurately show max import size, you need the minimum of those 2 values.  Unfortunately they're stored as string values that allow suffixes like M and K, so they have to be converted to bytes.  I came across this SO thread which had a drupal snippet that I adapted to suit this purpose.  I placed it in PB utilities as that seemed a logical place.

http://stackoverflow.com/questions/13076480/php-get-actual-maximum-upload-size

So, feel free to use/modify/ignore, but thought you may like to have it!

### DIFF
--- a/admin/templates/import.php
+++ b/admin/templates/import.php
@@ -127,7 +127,7 @@ $current_import = get_option( 'pressbooks_current_import' );
 			</script>
 		<p>
 			<?php _e( 'Supported file extensions:', 'pressbooks' ); ?> XML, EPUB, ODT, DOCX, HTML <br />
-			<?php _e( 'Maximum file size:', 'pressbooks' ); echo ' ' . ini_get( 'upload_max_filesize' ); ?>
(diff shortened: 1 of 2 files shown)


greatislander:
Thanks @jflowers45. This looks to be a helpful contribution. I'm going to merge in.

bdolor:
small request to modify the formatting of this PR (size of indents/tabs, space usage) to be consistent with what has been required in the past.
https://make.wordpress.org/core/handbook/coding-standards/php/

greatislander:
You got it, @bdolor. Please note for future reference, @jflowers45.

jflowers45:
Got it - will use real tabs and doublecheck the WP code formatting reference in the future.  Thanks for fixing it!

PR #155, opened 2015-06-23
ePub3 validation fix for users who change path to epubcheck

jflowers45:
We have a path for PB_EPUBCHECK_COMMAND defined in wp-config.php.  We needed a way for ePub2 files to validate against the ePub2 spec and ePub3 files to validate against ePub3.  There's only one constant, so this was my workaround ... an alternative would be to define a second constant but this seems to work for us so thought I'd put it out in case you like it.

I think that even if you didn't defined the command path yourself, if you were simultaneously exporting an ePub2 and an ePub3 in the current system, the validator would run against ePub v2 both times since the constant would already be defined by the time the second one came around.  I could be wrong though as I haven't looked into the internals of the export routine to verify that the loop all happens in one request.

### DIFF
--- a/includes/modules/export/epub/class-pb-epub201.php
+++ b/includes/modules/export/epub/class-pb-epub201.php
@@ -228,7 +228,11 @@ class Epub201 extends Export {
 	function validate() {
 
 		// Epubcheck command, (quiet flag requires version 3.0.1+)
-		$command = PB_EPUBCHECK_COMMAND . ' -quiet ' . escapeshellcmd( $this->outputPath ) . ' 2>&1';
+		$epubCheckCommand = PB_EPUBCHECK_COMMAND;
+		if ( $this->suffix == "_3.epub" && strpos( $epubCheckCommand, '-v 3' ) === false ) {
+			$epubCheckCommand = $epubCheckCommand . " -v 3.0";
+		}
+		$command = $epubCheckCommand . ' -quiet ' . escapeshellcmd( $this->outputPath ) . ' 2>&1';
 
 		// Execute command
 		$output = array();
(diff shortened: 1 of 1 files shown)


bdolor:
Can we look at [this line](https://github.com/pressbooks/pressbooks/blob/dev/includes/modules/export/epub3/class-pb-epub3.php#L52) — the logic may be flawed/redundant. There is little chance at that point that it hasn't already been defined and if you're not getting what you want from the epubcheck command it may be cleaner to see what happens/can happen there. There's a few other things that I can foresee happening, if that doesn't work.

greatislander:
Hi @jflowers45 and @bdolor, where are we with this PR? Do you plan further modifications?

bdolor:
My two cents: I'd rather see the function `validate()` overridden in `class-pb-epub3.php`. That strategy leverages the advantages of extending a class in the first place, is scalable, cleaner, etc.

jflowers45:
Sounds good.  I can submit a PR in the near future that moves the
modification there

greatislander:
Thanks! I'll look for the updated request.

jflowers45:
cool.  It will be either by end of day tomorrow or early next week

jflowers45:
Now that I've gone away and come back to this, I don't think this PR is necessary.  As of 3.0.1 (and maybe earlier), epubcheck is smart enough to detect what format the ePub is in rather than require the -v 3 flag to be passed into it.  Even if you pass -v 3 and you're validating an ePub 2, it's smart enough to validate against 2.  I verified this by running the command manually - it tells you

"Validating against EPUB version 2.0" 
or 
"Validating against EPUB version 3.0"

based on the file contents.

I hadn't picked up on that previously.  So, I think this PR is no longer needed and am closing it.  Sorry for the confusion.

greatislander:
Thanks @jflowers45!

PR #156, opened 2015-06-25
Fix books overlapping on initial catalog page rendering

bracken:
The Isotope library rendered the book boxes over each other
because the equalizer call was happening at the wrong time.

### DIFF
--- a/includes/pb-catalog.php
+++ b/includes/pb-catalog.php
@@ -249,6 +249,7 @@ $_current_user_id = $catalog->getUserId();
 	// <![CDATA[
 	jQuery.noConflict();
 	jQuery(function ($) {
+		$container.equalizer({ columns: '> div.book-data', min: 350, resizeable: false });
 		var $container = $('#catalog-content');
 		$('.filter-group-1').click( function () {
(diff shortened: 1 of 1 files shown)


bracken:
This is an example of what it looks like for us without this change: screen shot 2015-06-25 at 10 03 43 am

bracken:
Oh, and this only happens at certain unlucky widths. If you resize the window, or sometimes even just switch tabs it'll fix itself.

greatislander:
Thanks for the catch, @bracken!

bracken:
No problem, @changemachine fixed it, I just submitted the pull request. :)

connerbw:
This breaks the catalog.

    Uncaught TypeError: Cannot read property 'equalizer' of undefined

You need to declare `var $container` before using it...

bracken:
yeah, obvious problem looking at the diff. Happy to submit a fix really quick if you haven't already fixed it.

bracken:
Looks like there was a bit of a refactor that moved this around and exposed the hidden bug introduced by this pull request. This should fix the error: https://github.com/pressbooks/pressbooks/pull/282

connerbw:
Thanks!

PR #166, opened 2015-07-09
make book images clickable

kdv24:
(lines 222-225)  Moved href tags to make entire book image clickable instead of just the arrow being clickable


greatislander:
Thanks @kdv24!

PR #193, opened 2015-08-01
only look for sections if needed

bracken:
Section parsing can have some serious overhead
for large books, and if you're not using sections
on pages, it's unnecessary to do the work for them.

Test Plan:
 * TOC/Sidebar should work same as before

### DIFF
--- a/themes-book/pressbooks-book/page-cover-third-block.php
+++ b/themes-book/pressbooks-book/page-cover-third-block.php
@@ -1,5 +1,5 @@
 <?php global $blog_id; ?>
-			<section class="third-block-wrap"> 
+			<section class="third-block-wrap">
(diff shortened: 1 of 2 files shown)


bracken:
I think some work needs to be done to solve the section parsing performance problem, but since we're not using those I didn't take the time to dive in to solve it.

hughmcguire:
- is this true even in the case that Two-Level TOC is not selected in the Theme Options? 
- i wonder whether memcache (which we use on Pressbooks.com) might solve some of these problems?

greatislander:
@hughmcguire As many users don't have memcache I don't think that's a good solution. But this pull request addresses your first question by checking to see if section parsing is on before looking for sections in the current chapter. Thanks, @bracken.

PR #194, opened 2015-08-01
only get export meta value when needed

bracken:
The value of whether to export a page is only needed
when running exports. So don't fetch it for each page
on each page load. (TOC & because of navigation buttons)

The extra DB call for each page causes performance problems
with very large books

Test Plan:
 * TOC/sidebar TOC and navigation buttons should work as expected
 * Exporting anything should honer the export checkbox on a page

### DIFF
--- a/includes/class-pb-book.php
+++ b/includes/class-pb-book.php
@@ -145,7 +145,7 @@ class Book {
 	 * @see bottom of this file for more info
 	 * @return array
 	 */
-	static function getBookStructure( $id = '' ) {
+	static function getBookStructure( $id = '', $get_pb_export_meta=false  ) {
 
 		// -----------------------------------------------------------------------------
 		// Is cached?
@@ -197,6 +197,12 @@ class Book {
 
 				$post_name = static::fixSlug( $post->post_name );
 
+        if($get_pb_export_meta) {
+          $export = ( get_post_meta( $post->ID, 'pb_export', true ) ? true : false );
+        } else {
+          $export = false;
+        }
+
 				$book_structure[$type][] = array(
 					'ID' => $post->ID,
 					'post_title' => $post->post_title,
@@ -205,7 +211,7 @@ class Book {
 					'comment_count' => $post->comment_count,
 					'menu_order' => $post->menu_order,
 					'post_status' => $post->post_status,
-					'export' => ( get_post_meta( $post->ID, 'pb_export', true ) ? true : false ),
+					'export' => $export,
 					'post_parent' => $post->post_parent,
 				);
 			}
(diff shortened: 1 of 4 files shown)


bracken:
I looked at all the places that call `\PressBooks\Book::getBookStructure()` and made the ones I thought needed that export value still fetch it. I tested with our PDF export and a custom export plugin we have and it works as expected.

greatislander:
Thanks @bracken.

connerbw:
Hi. Are you using Memcached or Redis? Because this change breaks caching.

Steps to reproduce:

 + Restart caching engine
 + First, go to Web version of book (or any entry point that does not set `$get_pb_export_meta` to `true`)
   + With caching on, this will cache the book object with all `export` flags set to false
 + Navigate to Export page.
 + Export book

> Expected: Book has exported chapters
> Actual: Nothing is exported because all `export` flags are set to false.

This is a very odd solution for improving performance. You are falsifying the book object for short term gains. The book object is supposed to be in a true state at all times. That's the point of a persistent object.

Instead, you should try installing a cache and see if speed improves. It certainly does on Pressbooks dot com. :)  if caching is not enough I can work with you to improve performance.

More info:
http://kizu514.com/blog/wp-memcached-object-cache-leaks-memory/

Sorry, but we will be reverting this change.

bracken:
Yes, it was definitely a short term fix on my side and sorry it broke the expected caching. 

Even with caching this section could use an improvement to at least only do 1 call to the DB to get that meta field for all the relevant pages. As-is, it does a separate call for each page in the book, which in some of our cases is hundreds of extra calls for data that is only needed in one specific circumstance. So when there is an update, or this gets pushed out of the cache, some unlucky user is potentially waiting for a few seconds.

That's what I intended to do but didn't get around to it. And caching is an important piece of scaling a site, but we can still try to make it give bonus performance instead of required for a tolerable experience.

I think another solution would be to not cache that field and just do that DB query for the whole book when exporting. That probably happens infrequently enough that this would be a nice approach and it keeps the cache a bit slimmer. (probably not enough to make any practical difference though)

connerbw:
Try:

```php
global $wpdb;
$exportSettings = array();
$sql = $wpdb->prepare( "SELECT post_id, meta_value FROM {$wpdb->postmeta} WHERE meta_key = %s", 'pb_export' );
foreach ( $wpdb->get_results( $sql, ARRAY_A ) as $key => $val ) {
	$exportSettings[$val['post_id']] = $val['meta_value'];
}

// ... snip ...

'export' => ( isset( $exportSettings[$post->ID] ) && 'on' == strtolower($exportSettings[$post->ID]) ) ? true : false,
```

That being said, WordPress has done a lot of work around caching. It's not the most well documented or understood part of their code base but if you get Memcached or Redis working you will see huge gains.

I would be interested in the benchmarks.

Best regards,

PR #199, opened 2015-08-06
Update pressbooks-es_ES.po

colomet:
tipo corrections


greatislander:
Hi @colomet, thanks for this. In the future can I ask you to make any further translation submissions on [Transifex](https://www.transifex.com/pressbooks/pressbooks/language/es/) instead? Thank you.

Issue #115, opened 2015-04-15
audio files included in web output, not displaying in export


bdolor:
**To recreate**: Add an audio file from media library to your content area. Export the book. Audio file (url with link back to the audio file) and content proceeding audio file do not display.

**Description**: adding audio files to the content area invokes `wp_audio_shortcode()` which adds a default, inline `style` attribute with `visibility:hidden`. https://core.trac.wordpress.org/changeset/28182#file3

 Without the necessary CSS and JS present in the browser, this inline style remains in the XHTML output, preventing it from being displayed in any of the export routines. 

![screen shot 2015-04-15 at 11 38 37 am](https://cloud.githubusercontent.com/assets/2048170/7169702/cd49af34-e37c-11e4-920b-17804de95df2.png)

**Potential Resolutions**: It could be overridden with CSS, but that would mean going through each and every theme. Invoking HTMLLawed `deny_attribute` can target specific attributes on specific elements, but the `audio` tag is converted to a `div` tag in `\PressBooks\Sanitize\html5_to_xhtml11` so taking it out would mean targeting all `div` tags, instead of just the `audio` tag — is this such a bad thing? Likely a `preg_replace` or DOM traversal and node replacement is in order.

hughmcguire:
what does the div tag look like in the xHTML?

bdolor:
Note: `visibility: hidden` 
![screen shot 2015-05-14 at 12 02 54 pm](https://cloud.githubusercontent.com/assets/2048170/7639495/440a6738-fa31-11e4-8224-6fe2848fadd9.png)

Curious that  `<div id="audio-1-1">` also seems to envelop the rest of the content on the page, making any content that follows the audio clip not appear.

greatislander:
@bdolor That is very odd. In terms of fixing this—could targeting `<div>` elements whose `class="wp-audio-shortcode"` and removing the inline style be sufficient to address this primary issue?

bdolor:
seems like a plausible suggestion.

greatislander:
Okay, I can take a look at this.

greatislander:
Fix is underway.

greatislander:
@bdolor Looking for some `preg_replace` insight here. I can capture audio elements by `id` with no issue as follows:
```
$t = preg_replace('/(id=\"audio-[0-9]-[0-9]\")/u', "$1 data-replaced=\"true\"", $t);
```
However when I try to capture the `style` tag as well it doesn`t work:

```
[...]

PR #224, opened 2015-08-21
Display audio embeds in ePub 3 files

jflowers45:
I realized the ePub3 export is currently in beta, but the audio embeds aren't functional because they are hidden due to the way wordpress handles the audio shortcut.  In addition, the output of unclosed BR tags as well as the conditional IE9 comment break validation.  This PR fixes those issues.  It's possible you'll want a different approach to solve these problems (maybe it's possible to handle them further upstream) but this got things working on my end and at least brings the issues to your attention.

### DIFF
--- a/includes/modules/export/epub/class-pb-epub201.php
+++ b/includes/modules/export/epub/class-pb-epub201.php
@@ -1167,6 +1167,9 @@ class Epub201 extends Export {
 					Sanitize\decode( $title ),
 					$content,
 					'' );
+				
+				// Audio shortcodes in ePub3 exports (which inherit this method) result in the creation of <br> tags without closing tags, which breaks validation
+				$vars['post_content'] = str_ireplace( '<br>', '<br />', $vars['post_content'] ); 
 
 				$file_id = 'chapter-' . sprintf( "%03s", $j );
 				$filename = "{$file_id}-{$slug}.{$this->filext}";
--- a/includes/modules/export/epub3/class-pb-epub3.php
+++ b/includes/modules/export/epub3/class-pb-epub3.php
@@ -294,6 +294,10 @@ class Epub3 extends Epub\Epub201 {
 
 		libxml_use_internal_errors( true );
 
+		//these lines are produced by wordpress if Audio or Video are embedded and cause validation to fail
+		$html = str_replace( "<!--[if lt IE 9]><script>document.createElement('audio');</script><![endif] -->", '', $html );
+		$html = str_replace( "<!--[if lt IE 9]><script>document.createElement('video');</script><![endif] -->", '', $html );
+
 		// Load HTMl snippet into DOMDocument using UTF-8 hack
 		$utf8_hack = '<?xml version="1.0" encoding="UTF-8"?>';
 		$doc = new \DOMDocument();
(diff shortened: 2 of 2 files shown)


jflowers45:
it looks like you fixed the line break piece of the equation in 

https://github.com/pressbooks/pressbooks/commit/e6a52f9e01b1830d12ffb6e1bf05c8dd0238abf4

However the rest should still be valid

bdolor:
@jflowers45 We came across this and fixed it for the non-epub3 export formats in https://github.com/pressbooks/pressbooks/issues/115 — do you see any disadvantage putting a `preg_replace` line same as https://github.com/pressbooks/pressbooks/blob/dev/includes/pb-sanitize.php#L40 in the function `html5_to_xhtml5()`? 
I understand that there is also a validation task you are after (by stripping the IE9 hack), but the above will  take care of the same problem with audio being hidden, just at a different point in the code.

jflowers45:
@bdolor Nice!  I just tried it out - I removed my code and used the regular expression.  The audio player began showing up (in the ePub3 export) - the only difference is that the width is no longer 100%, but that could be done by overriding the wp-audio-shortcode CSS class, so no harm done.

So, I'd say your approach is definitely better.  Let me know if you'd like me to submit a PR with just that one line of code.  I could also try moving the hardcoded removal of those IE9 tags to the same function if you want.

bdolor:
Both suggestions you make sound good to me — `pb-sanitize.php` sounds like a good place to get ride of cruft we don't need. `html5_to_xhtml5()` also seems like a good place to strip out anything that's preventing it from validating. Let's close this PR and open a new one with the changes.

jflowers45:
@bdolor I tested moving the stripping of the script tags into the html5_to_xhtml11 function instead of having it inside of kneadHTML.  Once I moved it, the string replace stopped working.  After a little digging it appears that control characters are still in there at that point - if I paste the source from Chrome into Sublime Text here's what I see

controlchars

If I modify my string replace to be the following, it seems to work.  Feels a bit hacky, so thought I'd run by you before submitting it as part of a PR.  If you want I can also just submit the one liner for the audio stuff and this can be visited later.  Let me know your preference.

    $t= remove_control_characters($t);
    $t = str_replace( "!--[if lt IE 9]scriptdocument.createElement('audio');/script![endif] --", '', $t );
    $t = str_replace( "!--[if lt IE 9]scriptdocument.createElement('video');/script![endif] --", '', $t );

bdolor:
Could you try the following `preg_replace` which would both take care of audio and video, plus future proof it for versions of IE that would require a similar shiv:
 `$t = preg_replace( '/<!--\[if\slt\sIE(.*)-->/ui', "", $t );`

jflowers45:
couple notes

-/g was reported as an invalid regex modifier.  PHP docs seem to suggest that its PCRE implementation is greedy by default.  from here on out assume I'm not using the /g

-tried your regex inside of html5_to_xhtml5() - didn't pick anything up.  I think it's because of the issue with control characters being present at that point in the workflow

-tried moving your regex into kneadHTML and while it did pick up both the audio and video shims, it presented issues in chapters where I had both an audio and video element inserted.  I did some looking at the before/after html and couldn't figure out why, but extra html was getting sucked in by the regex and thus not displaying.

-I'm going to suggest for now that I just submit a PR with the one line that will un-hide the audio ePub3.   For audio/video shim removal I'll probably just do str_replace inside of the kneadHTML on our install, but I understand wanting a more robust solution in the official product whenever ePub3 goes live.  If I get a chance to play with the regex and figure out anything I'll let you know.  does that work for you?

bdolor:
sounds good — gets the job done, leaves validation as a task to be dealt with later.

PR #226, opened 2015-08-26
Fix audio shortcodes in ePub3 using same strategy as ePub2

jflowers45:
Modification of #224 - also see #115


jflowers45:
@bdolor let me know if this works ... I put a TODO: for IE9 shim since I noticed that a convention elsewhere in the pressbooks code

bdolor:
works for me! Over to you @greatislander

greatislander:
Thanks folks!

PR #251, opened 2015-09-16
resolves fatal errors on git submodule update init command

bdolor:
Using `git@github.com` generates a fatal error 'Permission denied (publickey). fatal: Could not read from remote repository' on the command `git submodule update --init` 

screen shot 2015-09-16 at 10 23 03 am


greatislander:
Thanks @bdolor!

PR #282, opened 2015-10-18
fix calling equalize before reference created

bracken:
This was introduced in pressbooks/pressbooks#156 and worked
by luck. Needs to be fixed in new context.


bracken:
I fixed the obvious problem but am not in a position to test it right now. I'd recommend a quick checkout and test locally to verify, I can get to it next week if needed though.

connerbw:
Thanks!

PR #330, opened 2016-01-12
remove date_default_timezone_set( 'America/Montreal' ) and use 24H clock

chrillep:


chrillep:
or maybe use
[date_default_timezone_get](http://php.net/manual/en/function.date-default-timezone-get.php)

greatislander:
Hi @chrillep, I'm closing this; instead opting to use the root site's timezone, date format and time format settings to determine how the export times are displayed (see https://github.com/pressbooks/pressbooks/commit/1bc3ad346834cebb71d9b3131d0fd57c89eac8f9). This is (I think) a better solution because the server's timezone may not be the one you want, and `date_default_timezone_get` is based on the server's information.

This functionality is partially blocked by https://github.com/pressbooks/pressbooks/issues/344, but you can access the general settings page by URL for now. Thanks for the report—I hope this will be a suitable solution for you.

chrillep:
yes, this will work better. Just didnt want ( 'America/Montreal' ) time..
Thanks!

PR #349, opened 2016-02-11
function to modify urls to protocol relative

bdolor:
We've recently forced our instance of PB to serve over SSL. To prevent a mixed media warning all `<img src` values for cover images need to be updated. 

screen shot 2016-02-11 at 10 27 50 am

PR introduces a simple utility function and a couple of wrappers to modify urls to become better versions of themselves, to be protocol relative. While the `src` value for cover images can be manually set/reset only when a new image is uploaded,  it means re-uploading hundreds of images. This avoids that manual task.

### DIFF
--- a/includes/class-pb-metadata.php
+++ b/includes/class-pb-metadata.php
@@ -287,8 +287,9 @@ class Metadata {
 		if ( is_object( $response ) ) {
 			$content = $response->asXML();
 			$content = trim( str_replace( array( '<p xmlns:dct="http://purl.org/dc/terms/">', '</p>', '<html>', '</html>' ), array( '', '', '', '' ), $content ) );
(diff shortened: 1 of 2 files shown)


greatislander:
@bdolor I'd prefer to use [set_url_scheme](https://developer.wordpress.org/reference/functions/set_url_scheme/) when populating the book information array. I'd rather not introduce a function that duplicates existing core functionality and requires writing additional tests. If that's okay with you I'll merge in https://github.com/pressbooks/pressbooks/commit/ecd2e409e9bbcca58369bec1501378688d3aa024 and you can update your pull request to just handle the Creative Commons license image.

bdolor:
works for me. The branch has been reset and forced pushed with the changes.

greatislander:
Cheers!

PR #372, opened 2016-03-14
access some specificity when interpreting transformation failed message

bdolor:
Generic 'Transformation failed' message that is generated if any one out of six files are not produced is unhelpful. This PR makes it only slightly more helpful for troubleshooting. 

screen shot 2016-03-14 at 10 34 45 am


greatislander:
Thanks @bdolor.

PR #402, opened 2016-04-28
parity between what is displayed and what is actually set

bdolor:
There was a discrepancy between default settings that were displayed and what was actually saved in the db. resolves https://github.com/BCcampus/pressbooks-textbook/issues/53


greatislander:
Thanks!

Issue #358, opened 2016-02-23
Change our namespace to \Pressbooks\


greatislander:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://wordpress.org/plugins/pressbooks/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

Our namespace is currently `\PressBooks\`, but we no longer use camelcase in our branding. For consistency, we should change our namespace to `\Pressbooks\`.

This will impact the use of our namespaced functions in other plugins, specifically https://github.com/bccampus/pressbooks-textbook, and some of our internal plugins, so coordination with @bdolor will be needed.

### Steps to Reproduce

Not applicable.

### Versions

Pressbooks: 3.2-dev
WordPress: 4.4.2
PHP: 5.6

PR #403, opened 2016-04-29
Add initial support for Pressbooks Plugins (closes #394).

### LINKED ISSUE #394: Enable limited plugin access for Pressbooks-specific plugins
### Prerequisites

* [x] Are you running the [latest version of Pressbooks](http://wordpress.org/plugins/pressbooks/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?

### Description

We need a robust system for Pressbooks-specific add-on modules. Instead of re-inventing the wheel, we should use the built in WordPress plugin system but filter plugins at the book level to only allow access to Pressbooks-specific plugins. So:

- All plugins are available for network activation at the Network level.
- Pressbooks-specific plugins (identified by a slug beginning with `pressbooks-`) which are not network activated can be enabled through the core plugin management by book administrators.

### Steps to Reproduce

Not applicable.

### Themes

Root Theme: Pressbooks Publisher
Book Theme: n/a

### Versions

Pressbooks: 3.4-dev
WordPress: 4.5
PHP: 5.6

greatislander:
1. Adds a tab to the plugin installer for network administrators which highlights plugins from the WordPress plugin directory that are tagged with `pressbooks`.
2. Enables access to the plugin interface for book administrators, restricting access to plugins prefixed with `pressbooks-`.

### DIFF
--- a/includes/admin/pb-plugins.php
+++ b/includes/admin/pb-plugins.php
@@ -15,9 +15,11 @@ namespace PressBooks\Admin\Plugins;
  */
 
 function filter_plugins( $plugins ) {
-	foreach ( $plugins as $slug => $value ) {
-		if ( false === strpos( $slug, 'pressbooks-' ) )
-			unset( $plugins[$slug] );
+	if ( ! is_super_admin() ) {
+		foreach ( $plugins as $slug => $value ) {
+			if ( false === strpos( $slug, 'pressbooks-' ) )
+				unset( $plugins[$slug] );
+		}
 	}
 
 	return $plugins;
--- a/includes/admin/pb-plugins.php
+++ b/includes/admin/pb-plugins.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * Control access to plugins.
+ *
+ * @author  Pressbooks <code@pressbooks.com>
+ * @license GPLv2 (or any later version)
+ */
+namespace PressBooks\Admin\Plugins;
+
+/**
+ * Hide plugins that aren't prefixed with `pressbooks-` (only applies to books).
+ *
+ * @param array $plugins
+ * @return array $plugins
+ */
+
+function filter_plugins( $plugins ) {
+	foreach ( $plugins as $slug => $value ) {
+		if ( false === strpos( $slug, 'pressbooks-' ) )
+			unset( $plugins[$slug] );
+	}
+
+	return $plugins;
+}
+
+/**
+ * Add a 'Pressbooks' tab to the plugin installer.
+ *
+ * @param array $tabs
+ * @return array $tabs
+ */
+
+function filter_install_plugins_tabs( $tabs ) {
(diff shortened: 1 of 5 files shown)


hughmcguire:
Nice work! ...

bdolor:
One assumption being made is that plugins will work if they are activated either at the book level or the network level. mPDF plugin will not work, for instance, at the book level. If this is a requirement, can it be explicitly stated, or is there a way to accommodate plugins/functionality that will only work at the network level? Also, as a network admin, with this change I'm losing the ability to activate other non pressbooks tagged plugins at the book level, though, this should be easy to take care of by checking what the current user can do? Finally, I'm wondering if a plugin template should be considered, in order to both establish guidelines and ensure that minimum requirements are taken care of. The inclusion of `compatibility.php` comes to mind?

bdolor:
Also, I see this — I gather that it's from PBT making a call to `add_menu_page()` 
screen shot 2016-04-29 at 9 15 13 am

greatislander:
A couple responses, @bdolor.

1. I think each plugin's installation instructions should indicate if it has to be network-enabled—so the network administrator, when installing plugins, would know that it was their responsibility to network-enable those that require it (much the way network administrators must network-enable Pressbooks itself). Any plugin that has been network-enabled will not be shown in the list available to book administrators (unless they are themselves network administrators).

2. Yes, we can certainly give network administrators the ability to activate all plugins, not just Pressbooks-specific ones, at the book level. I would just need to add a capability check to the filtering routine. (Okay, @hughmcguire?)

3. I absolutely agree that a plugin template needs to be considered, and your suggestion of including a compatibility check makes good sense. One prerequisite for releasing a plugin template, I think, is https://github.com/pressbooks/pressbooks/issues/358. I don't want to encourage the development of new plugins that use a namespace that should be changed.

4. I didn't test this PR with Pressbooks Textbook—but am I correct in assuming that you could remove that feature in Pressbooks Textbook once it is enabled in Pressbooks proper?

bdolor:
1. Okay - makes sense.
2. Great.
3. That seems fine, but this will only be a cosmetic requirement since namespace are not case-sensitive? 
4. Yes, absolutely.

bdolor:
3. see https://github.com/pressbooks/pressbooks/pull/405

includes/admin/pb-plugins.php
```diff
@@ -5,10 +5,12 @@
  * @author  Pressbooks <code@pressbooks.com>
  * @license GPLv2 (or any later version)
  */
-namespace PressBooks\Admin\Plugins;
+namespace Pressbooks\Admin\Plugins;
 
 /**
  * Hide plugins that aren't prefixed with `pressbooks-` (only applies to books).
+ * To show all plugins to all users, place the following in a plugin that loads before Pressbooks:
+ * `remove_filter( 'all_plugins', '\Pressbooks\Admin\Plugins\filter_plugins', 10 );`
```
greatislander: How to remove the filter of displayed plugin names (FYI @bdolor @bracken).

PR #405, opened 2016-05-02
case insensitive check for namespaces, to be consistent with how name…

bdolor:
…spaces function

would resolve https://github.com/pressbooks/pressbooks/issues/358 remove the risk of generating fatal errors.


greatislander:
Awesome. Thanks @bdolor!

PR #408, opened 2016-05-03
including wp rest-api

bdolor:
Initial commit to add wp-rest-api to pb, in preparation for more awesomeness.


greatislander:
Thanks!

PR #414, opened 2016-05-10
removing base href in xhtml output

bdolor:
please review https://github.com/pressbooks/pressbooks/commit/ea4a2f6690147ae658dc7d51bf4caed8cab02cd3 

I've tried to recreate the situation where 500 errors were generated after removing the `base href` declaration. I cannot, so maybe @connerbw could provide some context?

With `base href` undeclared/removed, anchor links work as expected in the XHTML file output. With `base href` declared, anchor/internal links (like in the TOC for instance `<a href="#chapter1">Some chapter</a>`) bring the user back to `http://somebasedomain.com/somebook/#chapter1`. Not only does this ignore the protocol (`is_ssl()`) of the site and exclude the necessary path (`/chapter/`,`/front-matter/`,`/back-matter/`) but it takes the user away from the file. 

Questions: 
1. Is this still necessary? 
2. Can it be removed without generating 500 errors - what am I missing? 
3. If it can't be removed, does it need to be modified to be accurate?

### DIFF
--- a/includes/modules/export/xhtml/class-pb-xhtml11.php
+++ b/includes/modules/export/xhtml/class-pb-xhtml11.php
@@ -187,7 +187,6 @@ class Xhtml11 extends Export {
 		echo "<head>\n";
 		echo '<meta content="text/html; charset=UTF-8" http-equiv="content-type" />' . "\n";
 		echo '<meta http-equiv="Content-Language" content="' . $this->lang . '" />' . "\n";
-		echo '<base href="' . trailingslashit( site_url( '', 'http' ) ) . '" />' . "\n";
 
 
 		$this->echoMetaData( $book_contents, $metadata );


connerbw:
We agree that the 500 errors wouldn't be in the XHTML output rather Prince and friends which use it, right?

I don't remember why I did this. I'm going to guess assets. Does your test book have relative path images?

bdolor:
I will put some relative path images in a test book and report back.

bdolor:
confirmed...500 errors are generated if relative paths to images exist.

`` assumes the base url and path (`/format/`) of the endpoint redirect for `xhtml`. 

screen shot 2016-05-10 at 9 45 04 am

If Prince is the only condition that needs this, is there any objection to removing the `base href` line in `$output` somewhere just prior to getting written to the file? Any other suggestions?

connerbw:
As described here, this is the "normal" behaviour of <base> (scroll down to drawbacks)

+ http://webdesign.tutsplus.com/articles/quick-tip-set-relative-urls-with-the-base-tag--cms-21399

If you take away base href then images will be broken in the XHTML. Why do you need to prioritize fixed anchor refs over broken images? If the anchor refs work as expected in Prince then that is our primary use case. It is also the expected behaviour.

Maybe you can override it with JavaScript?

+ https://rogerkeays.com/blog/using-base-href-with-anchors

bdolor:
Are images with relative paths broken in the XHTML output? They aren't on the test book I just produced. 
It's not necessary to think of it as a priority of one 'over' another, just as a problem that needs to be addressed. I'm not sure what the use case was for supporting a non WP standard way of linking to images, but I gather there was a reason. Nevertheless, it's important that internal links work in an XHTML output because the file isn't very useful otherwise. There was never any doubt it was expected behaviour, but the problem is that it's not desirable. 

Thanks for the JavaScript links.

connerbw:
I don't really have an opinion other than "that's how \<base\> works" and I remember we used it specifically to fix images in PDF outputs. If removing \<base\> after, on file export only, doesn't break Prince's behaviour then I'm OK with the change.

I'll let @greatislander and/or @hughmcguire have the final say.

greatislander:
So if I'm understanding this correctly, we have two issues:

1. Without `<base>`, relative images break exports based on XHTML (which would now include Prince and… ODT?).
2. With `<base>`, internal links break vanilla XHTML exports.

Seems to me the solution is to add `<base>` to the XHTML document during the export routines that need it (via a standard function of some sort) while leaving vanilla XHTML without a `<base>` tag? Or, as per @connerbw's proposal, do it the other way—add it to XHTML by default and only remove it when an XHTML file is written out in export? Would either of those solutions work?

(Part of my confusion here is that if you export an XHTML file without a `<base>` attribute, relative images are going to be broken no matter what, right? It's late. Please forgive.)

bdolor:
For clarity:
1. Relative images: Without `<base>`, relative images don't make it through a Prince export only. With *or* without `<base>` relative images don't make it through ODT, EPUB, MOBI and WXR (useless once imported). 
2. Internal links: With `<base>` internal links break XHTML file export only. 

Since this is much ado about a relative link to an image, is the use case to support relative links to images strong enough to warrant a work around, or since we know that relative links to images don't work in the majority of circumstances/is not WP standard/must be hand coded, is it the case that support for relative linked images can be phased out?

connerbw:
It's not necessarily hand coded. If I recall relative images happen on import, or in weird edge cases.  Maybe this happens less so nowadays with all the improvements in WP and PB over time. Whatever the reason was we added this code to make PDF exports just work™ :)

I vote remove it. (Ie. I vote approve this pull request)

Then, monitor customer feedback. (Or 500 errors?)

greatislander:
I'm inclined to agree with @connerbw and I think if we find that it's necessary we should be able to append the base to relative image URLs when we're preprocessing content for Prince export. I'm going to check that out on Monday and then I'll merge. Thanks for the input, all.

bdolor:
What's the status on this one?

Issue #429, opened 2016-06-20
HTML Import Error: Cannot use object of type WP_Error as array...


camuscando:
I did search and see a much older related issue in Pull Request #51.

### Description

Importing HTML results in a white screen with the following error:
PHP Fatal error:  Cannot use object of type WP_Error as array in /Applications/MAMP/htdocs/otl/wp-content/plugins/pressbooks/includes/modules/import/html/class-pb-xhtml.php on line 37

It appears that the wp_remote_get on line 22 of class-pb-xhtml.php is returning a WP_Error object but  the kneadandInsert method on line 37 expects an array element.

Not sure why wp_remote_get is failing. I've tested localhost html and remote files. I've reproduced the problem on my local, test, and production environments.

We are running RedHat Enterprise Linux and use automated deployment so we do have consistency across environments. Thanks for looking into this. I'm happy to test and help out in any way I can.

### Steps to Reproduce

1. From book site select import
2. Select HTML and enter url
3. Click import

**Expected behavior: Import html into chapter.

**Actual behavior:** Whitescreen. If you return to <book>/wp-admin after the whitescreen and select import again you will see the previous file there still waiting to continue but the import did not succeed.

I can bypass the error by putting a quick check on the return value of line 22. Either execution does yield the 'Tried to upload a file of type html and succeeded :)' in the log file.

### Themes

Root Theme: Pressbooks Publisher
Book Theme: Luther

### Versions

WP 4.5.2
PB 3.5.2
PHP 5.6.10
Luther and/or OpenTextbook Theme (duplicated with all plug-ins disabled including PB Textbook).

bdolor:
Introduced via https://github.com/pressbooks/pressbooks/commit/6ecb6ccaae1eb67fff0af5532dbc1dec754d9485 — lines 242-243 change the value of a url to a file path (`$current_import` before/`$current_import` after): 

![screen shot 2016-06-20 at 2 53 22 pm](https://cloud.githubusercontent.com/assets/2048170/16211562/26d0f552-36f7-11e6-8e6d-21b61df3bafd.png)

I will take a look at making an exception for html import.

PR #430, opened 2016-06-20
adding check for wp_error, conditional exclusion for upload path on h…

bdolor:
resolves https://github.com/pressbooks/pressbooks/issues/429


greatislander:
Thanks @bdolor!

PR #453, opened 2016-08-12
replacing string value with boolean

bdolor:
Previous values for options were stored as integers/boolean and were being blown away (reset to 0/false) when this function compared the previously stored value with a string.


greatislander:
Thanks for the catch, @bdolor!

PR #463, opened 2016-09-04
Use WP_DEFAULT_THEME for book activation.

greatislander:
Our book activation routine set the `stylesheet` and `current_theme` options to a hardcoded default (`pressbooks-book`), which meant that the `WP_DEFAULT_THEME` constant was ignored even if it was present (see [here](https://premium.wpmudev.org/blog/how-to-set-the-default-theme-for-wordpress-multisite/) for details on this constant). This PR removes the hardcoded defaults and, instead, sets `WP_DEFAULT_THEME` to `pressbooks-book` if it hasn't been defined in `wp-config.php`.

### DIFF
--- a/includes/class-pb-activation.php
+++ b/includes/class-pb-activation.php
@@ -25,9 +25,6 @@ class Activation {
 	 * @var array The set of default WP options to set up on activation
 	 */
 	private $opts = array(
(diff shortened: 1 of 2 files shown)


SteelWagstaff:
Thanks Ned!

PR #469, opened 2016-09-08
use to check if a site isnt a book besides the main site

chrillep:
need the option added somewhere in init

add_option( 'pressbooks_is_book', true );

### DIFF
--- a/includes/class-pb-book.php
+++ b/includes/class-pb-book.php
@@ -37,7 +37,12 @@ class Book {
 		// Currently, the main site is considered a "blog/landing page" whereas everything else is considered a "book".
 		// We might improve this in the future.
 
-		$is_book = ( is_main_site() === false );
+		if ( is_main_site() ) {
+			$is_book = false;
+		}
+		else {
+			$is_book = (bool) get_option( 'pressbooks_is_book' );
+		}
 
 		return $is_book;
 	}


greatislander:
@chrillep In my testing of this PR, `\Pressbooks\Book::isBook()` now returns false when called from blogs that don't have the `pressbooks_is_book` option set to true. This is a breaking change that strips the Pressbooks functionality from all existing books on my test network (e.g. undoes the admin changes, deregisters custom post types, et cetera). Is there a particular problem you are trying to solve with this PR? As it is we will not accept it.

greatislander:
@chrillep Closing as there has been no response.

chrillep:
sorry for not getting back at you! The problem i was having is that i wanted to be able to make a subsite not work as a "book". And since pressbooks turns every subsite into a "book", his was a quick solution to be able to use one installation of wordpress work for a book and make a subsite a complete site with domain mapping. Just so youd be able to tell the difference through an option.
If there is an obvious way ive missed it. Love to hear back from you.
Regards Christian

chrillep:
if you wanna make it backwards compatible it would probably be better if the option said pressbooks_not_book = true

greatislander:
I think the proper way to achieve this would be to apply a filter to the variable, like so:

`return apply_filters( 'pressbooks_is_book', $is_book );`

Then you could create a must-use plugin that does something like…

```
global $blog_id;

if ( 42 == $blog_id ) { // The ID of the blog that you want to use for something other than as a book
    add_filter( 'pressbooks_is_book', function() { return false; } );
}
```

That way you don't need to get into adding more options to the options table. If you want to update your PR accordingly I would strongly consider it for merge. Thanks.

greatislander:
@chrillep Also please note that any update to this PR must pass our CI tests before I will merge it. See link to Travis “Details”.

PR #473, opened 2016-09-12
fix logic for toc collapse

bdolor:
Presumably this condition checks for the existence of a TOC. 

If the TOC exists, the lengths of `.toc-btn` and `#toc` will never be less than 1, so the collapsible toc function will never fire if it checks for a value less than zero. 

screen shot 2016-09-12 at 12 26 32 pm

screen shot 2016-09-12 at 12 18 15 pm


greatislander:
Thanks for the catch @bdolor.

PR #475, opened 2016-09-13
adding action hook to register theme directories

bdolor:
Creating a theme for PB outside of PB involves some trickery. This action hook takes some of the magic out of actually getting a theme to display alongside PB core themes. Given all the recent refactoring on themes, this may make it easier for theme developers to create themes that PB doesn't have to maintain. 
A theme developer will still have to dance around the `allowedBookThemes` [function](https://github.com/pressbooks/pressbooks/blob/dev/includes/class-pb-pressbooks.php#L62) — anything added with this new action hook will be stripped away with [this] (https://github.com/pressbooks/pressbooks/blob/dev/includes/class-pb-pressbooks.php#L81) unless they append an array of themes to the `allowed_themes` filter and have it fire later (with a priority variable > 10). I'm not sure if the business needs are still present that would require all themes living in the `PB_PLUGIN_DIR` so I've left that alone. Needless to say, it would be nice if the condition were accommodating to themes that live in other directories.

### DIFF
--- a/includes/class-pb-pressbooks.php
+++ b/includes/class-pb-pressbooks.php
@@ -36,6 +36,8 @@ class Pressbooks {
 		register_theme_directory( PB_PLUGIN_DIR . 'themes-root' );
 		register_theme_directory( PB_PLUGIN_DIR . 'themes-book' );
 
+		do_action('pressbooks_register_theme_directory');
+
 		// Check for local themes-root directory
 		if ( realpath ( WP_CONTENT_DIR . '/themes-root' ) ) :
 			register_theme_directory( WP_CONTENT_DIR . '/themes-root' );


greatislander:
Hi @bdolor,

This isn't necessary. We load a bunch of our own extra book themes into the vanilla `/wp-content/themes` directory on Pressbooks.com and our other client networks. The key is this block [here](https://github.com/pressbooks/pressbooks/blob/dev/includes/class-pb-pressbooks.php#L70-L76).

Set `$GLOBALS['PB_SECRET_SAUCE']['BOOK_THEMES']` to an array of theme slugs either in your plugin or in `wp-config.php` and those themes will stay in the array of allowed book themes returned by `\Pressbooks\Pressbooks::allowedBookThemes()`. Let me know if you have any other questions about this.

bdolor:
Hi @greatislander, 
I am aware of that variable. Inside of a namespaced plugin, options to add theme exceptions are available only outside of the namespace, otherwise `$GLOBAL` gets limited by scope. Indeed `wp-config.php` is one place to declare it. You're right, it's not necessary, but it's also not as dev friendly as it could be. If theme proliferation is a desired goal then I'm looking for ways that make that process easier. If I want to develop a new theme for PB outside of PB, what are the barriers? This is one of them.

nciske:
Filters are the "WordPress way" to do this. 

Directly accessing globals is generally to be avoided, so I believe this is necessary and a good idea.

greatislander:
Fair points. Then—I would suggest you reconsider the PR to focus on the means by which the `$exceptions` array is determined in `\Pressbooks\Pressbooks::allowedBookThemes()`. Why not just add a filter to that array? Then you can explicitly enable themes that are placed in the standard themes directory, no?

bdolor:
One reason is that `allowedBookThemes` is a callback function given to the WP filter `allowed_themes`, so manipulating that array can already be achieved by hooking into the filter`allowed_themes` . The other reason is because of this [function](https://github.com/WordPress/WordPress/blob/master/wp-includes/theme.php#L36). 
`search_theme_directories()` returns a cached value if it's told not to force a new directory scan. In order to consistently return more than one theme that has been registered outside of PB with `register_theme_directory()` it needs to either force a new directory scan or hook into the PB function so that it is part of the cached value. 
Today I was telling you about some inconsistencies that I observed when loading multiple themes outside of PB. Modifying `allowedBookThemes` isn't enough for `switch_themes`, apparently. The `WP_Theme` object is modified depending on if a PB core theme is activated, or if a non-PB theme is activated. When a PB core theme is activated, everything works as expected; I watch the `WP_Theme` Object return the expected PB core themes and non-PB themes. This is not true when a non-PB theme is activated. The current theme (non-PB) is activated, but any others are omitted because of the cached value returned when `wp_prepare_themes_for_js` [calls](https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/theme.php#L525) `wp_get_themes()` which invokes `search_theme_directories()`.  You can observe the difference in behaviour if you set [this line](https://github.com/WordPress/WordPress/blob/master/wp-includes/theme.php#L36) to `$theme_directories = search_theme_directories(true);`

Hope that explains things thoroughly enough for my very revolutionary one-liner. :clap:

PR #480, opened 2016-09-16
resolve wonky float

bdolor:
Without a book description (not common), this area is left in an undesirable place. 

I believe this solves it. 

screen shot 2016-09-16 at 1 15 25 pm


greatislander:
Thanks for catching!

PR #485, opened 2016-09-20
remove conditions for fatal error

bdolor:
fatal error encountered due to an abstract class not implementing all methods. 

screen shot 2016-09-20 at 10 15 53 am

Don't know if this functionality is actually what is desired for `filterDefaults()`, but it moves things past a 500 error.


greatislander:
Thanks for catching this.

greatislander:
(@bdolor ultimately, the filterDefaults function will let themes or plugins override default option values—I forgot to put implement it in the export options class before merging.)

PR #492, opened 2016-09-29
Parts title web option

bdolor:
This option allows authors to choose whether or not to display the title of a part on each Chapter. Important feature to have for textbooks, currently a part of Open Textbook theme. Migrating over to PB. 

screen shot 2016-09-29 at 3 21 16 pm


greatislander:
Thanks @bdolor. This looks like an excellent addition. Would you be open to changing it to `part_title`, `renderPartTitle`, `Display Part Title`, `.part-title` etc. for consistency with the instructive text? I think it makes more sense to use the singular.

bdolor:
this is done.

PR #510, opened 2016-10-24
Standardize table button, add class button

greatislander:
This PR switches the JS used for the TinyMCE table button from our [forked version](https://github.com/pressbooks/mce-table-buttons/blob/master/tinymce4-table/plugin.js) to the current core plugin (TinyMCE 4.4.3). As this change removes the ability to assign a custom class to a table, a new button has been added which allow the user to add a custom class to the current selection or node.


greatislander:
FYI @thomasdumm, this changes the table editor behaviour slightly—the custom class field is no longer in the table properties window, but there is now a custom class button that lets you apply custom classes to any element. It's more broadly useful this way. Let me know your thoughts.

thomasdumm:
Fantastic. Looking forward to be able to use the latest version of the TinyMCE table editor which allows draging table colums to adjust column widths.

greatislander:
@thomasdumm Forthcoming!

greatislander:
Hi @thomasdumm—it appears that it is not possible to select the top-level table element with TinyMCE (it defaults to selecting the first editable node, i.e. the initial `<td>`) so the custom class button doesn't work for adding classes to `<table>` elements. The list of predefined table classes is still available within the table editor, and custom classes can be added in the code editor. Does this work for you?

thomasdumm:
Yes, if I can predefine the table classes somewhere?

greatislander:
Yep, we will set that up.

thomasdumm:
PS: In most situations that I can think of you do not want to add a class to the current element, but to an ancestor (div, ol, ul, table). Pitty the button does not allow to define a list of selectable elements.

PR #527, opened 2016-11-04
add filter hook 'pb_epub_has_dependencies' to epub export class

monkecheese:
this allows a user to enable the epub export option in the admin area under 'exports' by hooking into the 'pb_epub_has_dependencies' filter and without having to install the dependency EpubCheck, as installing system level dependencies may not be an option for some users.

### DIFF
--- a/includes/modules/export/epub/class-pb-epub201.php
+++ b/includes/modules/export/epub/class-pb-epub201.php
@@ -2178,7 +2178,7 @@ class Epub201 extends Export {
 			return true;
 		}
 
-		return false;
+		return apply_filters('pb_epub_has_dependencies', false);
 	}
 
 }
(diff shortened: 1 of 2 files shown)


monkecheese:
I did not implement a filter for .mobi exports, because when i tried to do that in the same way as above, the export would fail with the following error: `/opt/kindlegen/kindlegen: No such file or directory`.  So it didn't seem sensible to implement a filter that would allow a user to enable the option to export .mobi.

greatislander:
Hi @monkecheese, I would suggest filtering the output of `\Pressbooks\Utility\check_epubcheck_install()` in the `hasDependencies()` method instead—that way you'll get true for Epubcheck but false for Kindlegen. Make sense? Also, can you fix the [code standards](https://github.com/pressbooks/pressbooks/blob/dev/docs/coding-standards.md) error identified in the [Travis build](https://travis-ci.org/pressbooks/pressbooks/builds/173401207)?

greatislander:
Actually, maybe just filter the output of `check_epubcheck_install()` itself… I think that makes more sense.

PR #531, opened 2016-11-08
add filter hook 'pressbooks_session_configuration'

monkecheese:
some hosting platforms have requirements around sessions handling
that would require the modification of the way pressbooks handles
sessions. This filter allows users of pressbooks to add
custom configurations to session.


monkecheese:
Here is our specific use case:

https://github.com/lumenlearning/pressbooks/commit/f8ea4def8b213846fe22fac2da82b933638f3622

PR #532, opened 2016-11-10
add pb_append_chapter_content filter hook to pdf export

monkecheese:
this will enable users to append other chapter content on to
their mpdf exports.


monkecheese:
I don't understand what the unsuccessful checks are saying, but I'm happy to fix anything that needs it.  Cheers!

greatislander:
@monkecheese Basically just saying that your PR has added a single line of untested code (see [here](https://codecov.io/gh/pressbooks/pressbooks/compare/f63bf9805681321cca95871fde4fb61a795964d7...50a716fa6339aa637eacd2e800f5b70c3edf40ad)). However since that method doesn't have tests at this point anyway, I'm happy to merge.

PR #533, opened 2016-11-11
wrap all xhtml export book content in a div

monkecheese:
The div has class and id of all-parts-and-contents

### DIFF
--- a/includes/modules/export/xhtml/class-pb-xhtml11.php
+++ b/includes/modules/export/xhtml/class-pb-xhtml11.php
@@ -218,17 +218,8 @@ class Xhtml11 extends Export {
 		// Table of contents
 		$this->echoToc( $book_contents, $metadata );
 
-		// Front-matter
-		$this->echoFrontMatter( $book_contents, $metadata );
-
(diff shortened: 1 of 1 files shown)


greatislander:
Can you fix the coding standards error? https://travis-ci.org/pressbooks/pressbooks/jobs/175152604

greatislander:
So, if I'm understanding—you need to be able to force reset the page counter at a specified page number after the TOC (as per [this](https://pressbooks.slack.com/archives/opensource/p1478887338000317))? Am I wrong in thinking that you could just use this CSS to achieve that?

```
#toc + div {
  counter-reset: page 69;
}
```

See: https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Getting_started/Selectors#Information_Selectors_based_on_relationships

I'm just interested in seeing if your outcome can be achieved without adding any additional cruft to the XHTML. Thanks in advance.

monkecheese:
That looks like it might work.  I'll test it, hopefully sometime today or tomorrow.  Thank you @greatislander

PR #534, opened 2016-11-11
use original url to get image id instead of $cover

monkecheese:
users who use a plugin to host images outside of wordpress
(ex: Amazon S3) would find that cover images were broken in the
catalog. This fixes the broken images by using the original url
instead of the url set in the variable $cover.

### DIFF
--- a/includes/class-pb-catalog.php
+++ b/includes/class-pb-catalog.php
@@ -202,7 +202,7 @@ class Catalog {
 				$data[ $i ]['cover_url']['full'] = $cover;
 
 				// Cover Thumbnails
-				$cid = \Pressbooks\Image\attachment_id_from_url( $cover );
+				$cid = \Pressbooks\Image\attachment_id_from_url( $metadata['pb_cover_image'] );
 				foreach ( $cover_sizes as $size => $default ) {
 					$cid_thumb = wp_get_attachment_image_src( $cid, $size );
 					if ( $cid_thumb ) {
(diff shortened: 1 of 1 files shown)


monkecheese:
Here's a link to the original conversation and explanation regarding what this PR seeks to fix:

https://github.com/lumenlearning/pressbooks/commit/f584f3a350892d2702c4621d389dfb60be327407

greatislander:
Hi @monkecheese and @bracken, there's one problem that I see here. By substituting the `$metadata['pb_cover_image']` value as you've done, you'll now get a bunch of these errors if any books in your catalog don't have a cover image set:

screen shot 2016-11-14 at 1 18 20 pm

Is there any way you can make sure the logic accounts for this?

monkecheese:
I'll take another look at this.  thanks @greatislander

monkecheese:
I'm not getting this error @greatislander.  I've tried generating a book and not setting a cover image, then navigating to the catalog.  I've also tried generating a book, setting a cover image, then unsetting a cover image, then navigating to the catalog.  How might I replicate the error you're seeing?

greatislander:
@monkecheese It appears on the admin screen for 'My Catalog' for me…

monkecheese:
@greatislander hmm...i'm still not getting those errors.  i'll keep digging on this.

greatislander:
Always possible that there is something weird in my dev instance. I'll check again on my end as well.

monkecheese:
Either way, I think our solution was a little hacky to get us by.  I'm still not getting the errors you're getting, but the potential for that to come up for others isn't good.  So I'm closing this PR and opening a new one with a change that i think might work better for everyone.

https://github.com/pressbooks/pressbooks/pull/540

PR #538, opened 2016-11-15
RHEL 6 compatible

bdolor:
The latest RHEL6 package version of xmllint is 20706.


greatislander:
@bdolor I'll do it for this, but if you make a change that affects the versions in docs would love those to be updated in the same PR. Thanks!

PR #540, opened 2016-11-16
create new filter hook 'pb_cover_image'

monkecheese:
This will allow users who are not storing images locally on the
server to hook in to this filter and use either the data in the
$cover variable or the data in $metadata['pb_cover_image'], in
order to get cover images to display correctly depending on 
their specific use case.

### DIFF
--- a/includes/class-pb-catalog.php
+++ b/includes/class-pb-catalog.php
@@ -202,6 +202,19 @@ class Catalog {
 				$data[ $i ]['cover_url']['full'] = $cover;
 
 				// Cover Thumbnails
+				/**
+				 * Exposes $cover variable to be changed as-needed for cover images.
+				 *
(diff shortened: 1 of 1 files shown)


greatislander:
Hi @monkecheese, this is a much preferable solution to me. Can I get you to document the hook in the line preceding its first use? I was doing some research into inline documentation and I am going to try to do this for all existing filter and action hooks. https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

monkecheese:
sure thing, @greatislander!

monkecheese:
@greatislander, what version of pressbooks would this be rolled out in?  (i need to put something for the "@since" in the inline documentation)

greatislander:
@monkecheese 3.9.5.1. Thanks!

PR #562, opened 2016-11-28
Add hooks to support third-party imports (fixes #413)

### LINKED ISSUE #413: Import Modules
### Prerequisites
- [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
- [x] Are you running the [latest version of Pressbooks](http://wordpress.org/plugins/pressbooks/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
- [x] Are you experiencing this issue with one of our built-in root or book themes?
- [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?
### Description

Basically the same as https://github.com/pressbooks/pressbooks/issues/385, but for importing. This will allow us to support add-on importers like @lumenlearning's IMSCC importer without making them part of core.
### Steps to Reproduce

Not applicable.
### Themes

Root Theme: Pressbooks Publisher
Book Theme: Luther
### Versions

Pressbooks: 3.5.1
WordPress: 4.5.2
PHP: 5.6

monkecheese:


monkecheese:
This is just to show you what we have currently done to add imscc import functionality for our use case of pressbooks.  I don't anticipate this being merged per our conversation about wanting a more robust system for users being able to add new import types via 3rd party plugins.  Just wanted to get a conversation started.  :)

greatislander:
One thing that will need to be resolved with the module regardless of implementation is the coding standards. Can you follow the instructions [here](https://pressbooks.github.io/pressbooks/coding-standards/#validating-with-php-code-sniffer) to flag any issues?

monkecheese:
Yeah I was looking at that.  Will definitely clean that up.

bracken:
That would also only apply to things actually pulled into PB. If there is a way to add these from a plugin and have the actual import code like `includes/modules/import/imscc/class-pb-imscc.php` in our plugin then it's not so bad.
PB may want Common Cartridge import in core, but I still think the first step is being able to add new import types via filters or whatever.

monkecheese:
So, I've tried what I can think of to fix the error that travis is spitting out, but I don't understand what it wants, so I left it for now in hopes you could shed some light on it @ned.

Other than that, here is what I had to do to hook into these new filter hooks:

https://github.com/lumenlearning/candela-utility/commit/0add00ee891dcfe0dab6c81bfe9e5e9af755f142

templates/admin/import.php
```diff
@@ -8,6 +8,13 @@
 $import_revoke_url = wp_nonce_url( get_admin_url( get_current_blog_id(), '/tools.php?page=pb_import&revoke=yes' ), 'pb-revoke-import' );
 $current_import = get_option( 'pressbooks_current_import' );
 $custom_post_types = apply_filters( 'pb_import_custom_post_types', array() );
+$import_option_types = apply_filters( 'pb_select_import_type', array(
+	'wxr' => 'WXR (WordPress eXtended RSS)',
```
greatislander: The values here should be wrapped in `__()`, e.g. `__( 'WXR (WordPress eXtended RSS)' )`. This will fix the Travis error, because…

templates/admin/import.php
```diff
@@ -153,11 +160,9 @@ function(){
 					</th>
 					<td>
 						<select id="type_of" name="type_of" class="pb-html-target">
-							<option value="wxr"><?php _e( 'WXR (WordPress eXtended RSS)', 'pressbooks' ); ?></option>
-							<option value="epub"><?php _e( 'EPUB (for Nook, iBooks, Kobo etc.)', 'pressbooks' ); ?></option>
-							<option value="odt"><?php _e( 'ODT (word processing file format of OpenDocument)', 'pressbooks' ); ?></option>
-							<option value="docx"><?php _e( 'DOCX (word processing file format of Microsoft)', 'pressbooks' ); ?></option>
-							<option value="html"><?php _e( 'HTML (scrape content from a URL)', 'pressbooks' ); ?></option>
+							<?php foreach ( $import_option_types as $option => $label ) { ?>
+								<option value="<?php echo $option; ?>"><?php echo _e( $label, 'pressbooks' ); ?></option>
```
greatislander: …you can't localize something that isn't a string (`_e( $label )` won't work).

greatislander: A string literal, I mean!

monkecheese: Good to know moving forward, thanks @greatislander.

includes/modules/import/class-pb-import.php
```diff
@@ -305,12 +316,20 @@ static public function formSubmit() {
 			// --------------------------------------------------------------------------------------------------------
 			// Set the 'pressbooks_current_import' option
 
-			$allowed_file_types = array(
+			/**
+			 * Allows users to append import options to the list of allowed file types.
+			 *
+			 * @since 3.9.5.2
```
greatislander: One last change, @monkecheese— this is going to be in 3.9.6.

monkecheese: i figured it was one or the other.  :P  I'll make the change.

PR #583, opened 2016-12-05
Current hacks for latex render options

monkecheese:
These are the current hacks for latex renderers that we use.  Again, this commit is intended to show what we currently have so that we can highlight where changes might need to be made to make this hookable and modular.  All feedback is appreciated (especially around some of the jquery and scripts stuff).


monkecheese:
These are the changes that I came up.  Let me know @greatislander if these seem reasonable and/or if there are any things you'd like me to change.

monkecheese:
And here is our plugin, with functions that hook into these action/filters:

https://github.com/lumenlearning/candela-utility/commit/c32549da753e4cff3c7fbdd2eaeadcef6e7bc401

Issue #616, opened 2017-01-03
Word import produces excessive inline styles


greatislander:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

In the wake of a few recent PRs (#588, #607, #608, #609) we seem to be getting excessive inline styling from Word imports. See [here](https://gist.github.com/greatislander/69e126e7fa22b58bdeda60d13d70f797/revisions) (before and after the PRs).

@crism, hoping you can help with this.

### Steps to Reproduce

1. Import a Word document.

**Expected behavior:** Basic Word styles are imported.

**Actual behavior:** Inline styles appear throughout, sometimes on a per-word level.

### System Information

Pressbooks: 3.9.7-dev
WordPress: 4.7
PHP: 7.1

greatislander:
@crism Here's the doc we used for testing: [lorem-ipsum-footnotes.docx](https://github.com/pressbooks/pressbooks/files/683148/lorem-ipsum-footnotes.docx)

crism:
Yes, I erred on the side of preserving information. We can simplify it: Right now, a CSS style string is created; it is then inspected by a series of templates that generate HTML markup as appropriate, and remove the corresponding styles (_e.g._, `font-style:italic;` becomes `em`). Whatever CSS style is left is then dumped into a `style` attribute on a `span` element. It could be discarded instead.

greatislander:
@crism That would be great. I reverted the PRs that we merged in—would you be able to put together one PR that consolidates your fixes for all the issues you targeted in those four PRs that discards styles other than bold, italic, superscript and subscript?

crism:
Yes, I can do that, @greatislander.

greatislander:
Thanks again for all your work on this, @crism!

Issue #603, opened 2016-12-21
Word import doubles endnotes


crism:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

When importing a Word document with endnotes, each endnote is repeated. Each is formatted correctly, with back-links to the reference, but each is present twice.

### Steps to Reproduce

1. Make a Word document with endnotes.
1. Import it into Pressbooks.

**Expected behavior:**

Each endnote should appear at the end of the chapter or part. Each should appear only once.

**Actual behavior:**

The endnotes are present correctly, but each repeats.

### System Information

Pressbooks: 3.9.6
WordPress: 4.7
PHP: 5.6.28

crism:
This happens in PHP, and it’s code I changed a few months back, and worked correctly then. I’ll dive into it right after addressing #602.

greatislander:
Thanks @crism!

crism:
Hmm… looks like my pull request #608 is actually relevant here—I’ll merge it into the branch for this. The problem is that the PHP presumes that _all_ `<sup>` elements represent a footnote/endnote reference. But there’s still a better way, which is to look for the `<a>` elements that have footnote-shaped references (_i.e._, `sdfootnote*sym`).

crism:
Bah… that commit misses the end notes altogether. Will work on it more tomorrow.

Issue #602, opened 2016-12-21
Word import character-level style problems


crism:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

1. When a Word character style asserts that it’s _not_ bold (`<w:b w:val="0">`), the CSS string constructor asserts it _is_ bold (`font-weight:bold;`).
1. When one style is based on another, redundant assertions lead to doubling in the CSS string (_e.g._, one style is italic, and based on another style, also italic, leads to `font-style:italic;font-style:italic;`). This leads to the resulting generated markup being redundant (`<em class="import-Example"><em>sample</em></em>`).

### Steps to Reproduce

1. Make a Word document with an inline style. Set it to be bold-italic; then unset the bold part.
1. Make a second style based on the first.
1. Import the document into Pressbooks.

**Expected behavior:**

Both styles should be imported as `<em>` elements.

**Actual behavior:**

The first style is imported as `<strong><em>`; the second as `<strong><strong><em><em>`.

### System Information

Pressbooks: 3.9.6
WordPress: 4.7
PHP: 5.6.28

crism:
I’m working on a fix…

Issue #582, opened 2016-12-05
Word import drops styles and language information


crism:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

When Word documents are prepared using best practices—styles for formatting, and no direct bold or italic markup, the Word import process into Pressbooks ignores this information. It also ignores language tagging.

### Steps to Reproduce

1. Import [this Word DOCX file](https://github.com/pressbooks/pressbooks/files/632124/styles_language_clean.docx).

**Expected behavior:** The word “styles” should be italicized. The phrase “en français” should be italicized and marked as French.

**Actual behavior:** The document results in two paragraphs with no inline formatting or tagging whatsoever.

### System Information

Pressbooks: 3.9.5.1
WordPress: 4.6.1
PHP: 5.6.28

thomasdumm:
what is the document language of your word document?

thomasdumm:
maybe related to https://github.com/pressbooks/pressbooks/issues/298

crism:
@thomasdumm, the document is in US English. It’s not #298; that’s a result of the heading styles having different names in different languages. This is a problem with inline styles; the Emphasis style is defined as italic, but the import just ignores it as there’s no direct formatting.

I’m working on a fix in the XSLT…

greatislander:
Thanks, @crism!

crism:
I think this will fix it, but I need to upgrade WordPress before I can test it server-side. [Here’s an updated Word doc](https://github.com/pressbooks/pressbooks/files/643247/styles_language_clean.docx) I used for testing (exercising a few additional features).

crism:
Thanks, @greatislander, @emasters!

PR #629, opened 2017-01-10
Restore inline style and footnote fixes, but drop non-semantic inline style

crism:
This should address #616 and restore #582, #602, and #603.


crism:
IMO, @greatislander, those are discrete issues—footnote content comes in via PHP (and I was going to add an issue about that), and the paragraph styles are the way they’ve always been. Could be fixed, but would be a possible regression.

greatislander:
You're absolutely correct—those issues exist in the current importer. I'll merge this as is then and we can address them separately.

PR #633, opened 2017-01-12
Book cache

monkecheese:
`Book::getBookStructure` gets the `export` metadata in an extra
`get_post_meta()` call.

The current method for this is inefficient. These tests will help
ensure the future optimization does not break this function.

This is based off this discussion here: https://github.com/pressbooks/pressbooks/pull/194

### DIFF
--- a/includes/class-pb-book.php
+++ b/includes/class-pb-book.php
@@ -185,6 +185,15 @@ class Book {
 
 		$q = new \WP_Query();
 
+		/**
+		 * Fetch all pb_export meta values for this book
+		 */
+		global $wpdb;
+		$post_ids_to_export = array();
+		foreach ( $wpdb->get_results( $wpdb->prepare( "SELECT post_id, meta_value FROM {$wpdb->postmeta} WHERE meta_key = %s", 'pb_export' ), ARRAY_A ) as $val ) {
+			$post_ids_to_export[ $val['post_id'] ] = $val['meta_value'];
+		}
+
 		foreach ( $custom_types as $type ) {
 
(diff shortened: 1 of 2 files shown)


includes/class-pb-book.php
```diff
@@ -185,6 +185,16 @@ static function getBookStructure( $id = '' ) {
 
 		$q = new \WP_Query();
 
+		/**
+		 * Fetch all pb_export meta values for this book
+		 */
+		global $wpdb;
+		$post_ids_to_export = array();
+		$sql = $wpdb->prepare( "SELECT post_id, meta_value FROM {$wpdb->postmeta} WHERE meta_key = %s", 'pb_export' );
+		foreach ( $wpdb->get_results( $sql, ARRAY_A ) as $val ) {
```
greatislander: This is silly but I think WPCS is dissatisfied because you are putting the `$sql` variable in `$wpdb->get_results()` rather than your `$wpdb->prepare()` function. So you can remove the previous line and do this:

```
foreach ( $wpdb->get_results( $wpdb->prepare( "SELECT post_id, meta_value FROM {$wpdb->postmeta} WHERE meta_key = %s", 'pb_export' ), ARRAY_A ) as $val ) {
```

monkecheese: OH!  I was trying to figure out what that error was trying to tell me.  I thought it was getting angry at the string interpolation used around `$wpdb->postmeta`.  It appears to be happy now on local.  I'll push up the change.

greatislander:
@monkecheese Going to test locally on Monday but this looks great!

Issue #630, opened 2017-01-10
Import drops language information


crism:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

When importing language-tagged content, the Tidy stage strips out `lang` and `xml:lang` attributes from the resulting HTML.

### Steps to Reproduce

1. Import the styles-and-language document in #582.
2. Examine the resulting markup in the Pressbooks editor. The French phrase has no language tagging associated with it.

**Expected behavior:** 

```html
<em class="import-Foreign" lang="fr-FR" xml:lang="fr-FR">en français</em>
```

**Actual behavior:**

```html
<em class="import-Foreign">en français</em>
```

### System Information

Pressbooks: 3.9.6
WordPress: 4.7
PHP: 5.6.29

crism:
I believe this is due to the configuration in the `htmLawed` module, which is called after every import. There is no security risk in allowing language tagging, so it should be configured not to strip those attributes.

greatislander:
It doesn't seem to be htmLawed— I tried importing your sample doc while bypassing htmLawed altogether in the import module and the `lang` and `xml:lang` attributes were still absent.

crism:
Thanks, @greatislander—I’d only done a superficial search for the source. Let me know if you want help or want to assign it to me.

greatislander:
If you could take a look that would be great, thanks!

crism:
@greatislander, how did you rule out `htmLawed`? It’s called directly in `Docx::tidy()`, though imported in the `Import` class file… I think it is still getting called, and I don’t see anywhere else that the HTML is manipulated to that degree.

greatislander:
I modified the `Docx::tidy()` method to just return the `$html` variable without passing it through `htmLawed`. I missed the identically-named method in the parent class, though.

crism:
[...]

PR #634, opened 2017-01-12
pressbooks/pressbooks#630: allow lang and xml:lang attributes.

crism:
This configuration change seems to fix the problem, in my tests.


greatislander:
Thanks @crism!

PR #635, opened 2017-01-14
adding support for Canadian Indigenous languages

bdolor:
**Languages supported:**
Chipewyan, Inuktitut, Plains Cree, Cree, Moose Cree, Slave, Northern Cree, Naskapi, Swampy Cree, Southern East Cree, Ojibwa

**Unicode Range:**
Basic Latin (U+00000 - U+0007F)
Latin-1 Supplement (U+00080 - U+000FF)
Latin Extended-A (U+00100 - U+0017F)
Spacing Modifier Letters (U+002B0 - U+002FF)
Combining Diacritical Marks (U+00300 - U+0036F)
Unified Canadian Aboriginal Syllabics (U+01400 - U+0167F)
Unified Canadian Aboriginal Syllabics Extended (U+018B0 - U+018FF)

**References:**
https://fontinfo.opensuse.org/fonts/NotoSansCanadianAboriginalRegular.html
http://www.google.com/get/noto/#sans-cans
http://www-01.sil.org/iso639-3/documentation.asp?id=cre
https://github.com/googlei18n/noto-fonts/tree/master/unhinted

Testing does not break EPUB, so fallback font stack is not included.


includes/class-pb-globaltypography.php
```diff
@@ -20,6 +20,7 @@ function getSupportedLanguages() {
 			'grc' => __( 'Ancient Greek', 'pressbooks' ),
 			'ar' => __( 'Arabic', 'pressbooks' ),
 			'he' => __( 'Biblical Hebrew', 'pressbooks' ),
+			'cre' => __( 'Canadian Indigenous ', 'pressbooks' ),
```
greatislander: My convention has been to use the ISO 639-1 identifier for languages, but since this  is an instance where the name of the script and the languages are not identical and where the script supports multiple languages (many of which are not Cree variants), I'd like to request that you switch from `cre` to the ISO 15924 script name identifier, `cans` (see: https://en.wikipedia.org/wiki/ISO_15924).

Also in keeping with this, could you modify the description to Canadian Indigenous Syllabics to be more in line with the  ISO 15924 name? Thanks @bdolor!

bdolor:
👍

Issue #497, opened 2016-10-03
When PDF image resolution is set to 300dpi, LaTeX images are very small


greatislander:
### Prerequisites
- [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
- [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
- [x] Are you experiencing this issue with one of our built-in root or book themes?
- [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?
### Description

When `prince-image-resolution` is set to 300dpi, LaTeX images generated by PB-LaTeX (Automattic LaTeX service) are too small.
### Steps to Reproduce
1. Change PDF image resolution theme option to 300dpi.
2. Export a book with LaTeX content.

**Expected behavior:** LaTeX images are a suitable resolution.

**Actual behavior:** They are very small.
### Themes

Book Theme: ALL
### Versions

Pressbooks: 3.8.1
WordPress: 4.6.1
PHP: 5.6

PR #640, opened 2017-01-20
adding image resolution options

bdolor:
Textbook authors affected by https://github.com/pressbooks/pressbooks/issues/497 can now choose from a greater array of options to decide what works for them. One of our textbook authors is having problems exporting a book with math formulas. Those formulas rendered with `pb-latex` are too big at 72 DPI and too small at 300 DPI. They render with greater clarity at around 100 DPI.


hughmcguire:
Just a note, this is related to a known problem/(expected tech behaviour, but not for users) in Prince , where small user images "shrink" on the page to given "physical" DPI... 

We don't have a final solution for that, but just want to flag this as part of wider problem beyond LaTeX.

greatislander:
Hi @bdolor—as per our discussion, can you update this to just add a filter to the array of resolutions?

PR #667, opened 2017-02-15
Reordering epub CSS rules to enable margins around images.

crism:
Exported ebooks did not have appropriate margins around images due to incorrect ordering of CSS rules. This is fixed in this PR for the Fitzgerald theme; a similar change may be needed in other themes.

### DIFF
--- a/themes-book/fitzgerald/export/epub/style.scss
+++ b/themes-book/fitzgerald/export/epub/style.scss
@@ -310,9 +310,9 @@ img {
         max-width: 100%;
 }
 .wp-caption {
-        height: auto;
-        max-width: 100%;
-        margin: 1em 0;
+  height: auto;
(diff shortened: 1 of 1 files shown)


greatislander:
@crism Can you fix the code formatting violations? Thanks in advance.

crism:
@greatislander, the SCSS files are full of tabs—my PR is consistent with what’s already there. I can convert them to spaces easily enough, but the rest of the file will still get flagged.

greatislander:
Understood. We haven't touched this theme in a couple years, which is why it hasn't been updated to our current coding standards. If it's not too much trouble for you to update the changed portion of your PR regardless, that would be appreciated.

crism:
Thanks, @greatislander.

PR #675, opened 2017-02-28
Themes running content

hughmcguire:
Ned this is the rework of the themes running content & numbering, so that we have one generic file to rule all page structures/running content/number positioning ... and can modify where we want numbers & content, using simple variables. (could be extended later to be user options, but no need now).

I suspect its far more verbose than it needs to be, but on the other hand it appears to work.

NOTE: this includes a fix for Clarke as well ... 
Leaving these in need of similar, I think: 
- Gerhard
- Asimov? 
- anything else?


hughmcguire:
NOTE: this is a breaking change for "new" themes, so need to fix: 
Rothbard, Gerhard & Asimov

hughmcguire:
So, this is ready to merge I think, with following updates made to themes: 
* Clarke: this pull request includes updates to Clarke to support it
* Rothbard: there is a new commit to `dev` to support this
* Asimov: there is a pull-request to fix
* Gerhard: `v2` has been updated to support this

PR #686, opened 2017-03-09
First page numbering

greatislander:
We need to extend the new running content code so that:

- the first page of front matter / parts / chapters / back matter can have the number in a different place than other pages
- the number will only ever appear in one place on the first page of front matter / parts / chapters / back matter (so that when a rule for `@page:left` and `@page chapter:first` applies to the same page it doesn't duplicate efforts)


greatislander:
@hughmcguire If I can get to this before the weekend I will.

Issue #472, opened 2016-09-12
API settings


greatislander:
### Prerequisites
- [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
- [x] Are you running the [latest version of Pressbooks](http://wordpress.org/plugins/pressbooks/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
- [x] Are you experiencing this issue with one of our built-in root or book themes?
- [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?
### Description

We will need these.
### Steps to Reproduce

Not applicable.
### Themes

Not applicable
### Versions

Pressbooks: 3.8-dev
WordPress: 4.6.1
PHP: 5.6

greatislander:
Specifically we need a network-admin level settings page which turns off the root-level `/books/` endpoint(s).

PR #689, opened 2017-03-16
adding wp api rest routes, support for custom post types

bdolor:
This adds WP REST API support for custom post types. 

details here: https://packagist.org/packages/bccampus/rest-routes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
     "gridonic/princexml-php": "^1.2",
     "pimple/pimple": "^3.0",
     "leafo/scssphp": "^0.6.6",
-    "pressbooks/pb-api": "^1.1"
+    "pressbooks/pb-api": "^1.1",
+    "bccampus/rest-routes": "^1.0.3"
   },
   "require-dev": {
     "phpunit/phpunit": "~5.5",


greatislander:
Hi @bdolor, this looks good but I have a few questions:

1. Should we instead add core REST API support on registration for our custom post types (I've already tested this but have held off on implementing—see point 3);
2. Assuming point 1, should we let users install this as a standalone plugin if they need to enable REST API support for other post types?
3. What are your thoughts are on the utility of exposing individual chapters, front/back matter and parts via the core REST API before we have an endpoint that exposes the book structure as a whole via the same REST API? Is that jumping the gun?

bdolor:
1. No, I don't think we should. Instead we would want to consider the plugin registration event as an appropriate time to register the PB custom post types. It could be enabled by default, on the condition that the plugin detects that it's in a PB environment and is PB compatible.
2. Yes, I think we should. Activating a plugin instead of writing this into core follows the principle of modularity. Giving admin users the option to enable/disable makes this functionality usable across a wider variety of use cases.
3. We could frame any questions around timing with a release early, release often paradigm. Features could be shipped early in a minimal state, then improved further in future changes. I'd be open to putting that on the road map.

greatislander:
Closing in favour of registering core post type routes internally.

greatislander:
See also: #472.

PR #734, opened 2017-04-25
Use `network_home_url()` instead of `PATH_CURRENT_SITE`

jeremyfelt:
There are several redirects and an anchor that use `PATH_CURRENT_SITE` to generate the URL. It's possible to configure multisite without this constant, at which point PHP errors are generated in these
areas.

This makes the shift to `network_home_url()`, which uses the domain and path of the current network and provides a filter for anyone wishing to customize.


greatislander:
Thanks @jeremyfelt!

PR #735, opened 2017-04-25
Include current `$post` with `wp_enqueue_media()`

jeremyfelt:
If editing a post (post type), this allows additional settings to be enqueued for media handling. Specifically, this allows the auto embed of content from a URL such as YouTube, which does not work at this time.

If editing a user, term, etc... `$post` will be `null`, which is the default argument in `wp_enqueue_media()`, resulting in no change.

There's a chance that the disable of auto-embed content is intentional. In that case I can find another workaround by filtering the settings. :)

### DIFF
--- a/symbionts/custom-metadata/custom_metadata.php
+++ b/symbionts/custom-metadata/custom_metadata.php
@@ -219,7 +219,10 @@ class custom_metadata_manager {
 	}
 
 	function enqueue_scripts() {
-		wp_enqueue_media();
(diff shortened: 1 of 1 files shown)


greatislander:
@jeremyfelt Our bundled version of https://github.com/Automattic/custom-metadata/ only has a couple modifications (you can diff to see them) and this isn't one of them. I don't think upstream is accepting PRs these days so happy to merge this.

PR #753, opened 2017-05-22
Use template for pb_private() function

stepmuel:
This allows themes to overwrite the standard login prompt by providing their own private.php template.


greatislander:
@stepmuel We won't merge this PR give your new commit in 02ae529. Please submit separate pull requests for the two separate things you are trying to do.

PR #754, opened 2017-05-23
Regex Search and Replace

stepmuel:
* Add regex option to SearchAndReplace
* Escape potential backreferences when not in regex mode

The second point was a preexisting bug where a Replace With value of `a$1b` or `a\1b` would result in `ab`. 

If the regex functionality seems too confusing for existing users, I could imagine having a `PB_REGEX_SEARCH` constant to enable it conditionally, but having regex is a crucial feature in our installation.

### DIFF
--- a/includes/modules/searchandreplace/class-pb-search.php
+++ b/includes/modules/searchandreplace/class-pb-search.php
@@ -22,12 +22,12 @@ class Search {
 	}
 
 	function regex_error( $errno, $errstr, $errfile, $errline ) {
-		$this->regex_error = __( 'Invalid regular expression', 'pressbooks') . ': ' . preg_replace( '/(.*?):/', '', $errstr );
+		$this->regex_error = __( 'Invalid regular expression', 'pressbooks' ) . ': ' . preg_replace( '/(.*?):/', '', $errstr );
 	}
 
 	function search_and_replace( $search, $replace, $limit, $offset, $orderby, $save = false ) {
 		// escape potential backreferences when not in regex mode
-		if ( $this->regex === false ) {
+		if ( ! $this->regex ) {
 			$replace = str_replace( '\\', '\\\\', $replace );
 			$replace = str_replace( '$', '\\$', $replace );
 		}
@@ -54,7 +54,7 @@ class Search {
 				set_error_handler( array( &$this, 'regex_error' ) );
 				$valid = @preg_match( $search, '', $matches );
 				restore_error_handler();
-				if ( $valid === false ) {
+				if ( false === $valid ) {
 					return $this->regex_error;
 				}
 				return $this->find( $search, $limit, $offset, $orderby );
--- a/includes/modules/searchandreplace/class-pb-search.php
+++ b/includes/modules/searchandreplace/class-pb-search.php
@@ -24,7 +24,7 @@ class Search {
 		// evaluate expression without imput and capture potential error message
 		$regex_error = 'invalid';
 		$error_handler = function( $errno, $errstr, $errfile, $errline ) use ( &$regex_error ) {
-			$regex_error = preg_replace( '/(.*?):/', '', $errstr );
+			$regex_error = preg_replace( '/(.*?):/', '', $errstr, 1 );
 		};
 		set_error_handler( $error_handler );
 		// @codingStandardsIgnoreLine
--- a/includes/modules/searchandreplace/class-pb-search.php
+++ b/includes/modules/searchandreplace/class-pb-search.php
@@ -14,8 +14,15 @@ class Search {
 	var $source = null;
 	var $error = null;
 
+	var $regex = false;
+	var $regex_error = null;
+
 	function name() {
 		return '';
+	}
+
+	function regex_error( $errno, $errstr, $errfile, $errline ) {
+		$this->regex_error = __( 'Invalid regular expression', 'pressbooks') . ': ' . preg_replace( '/(.*?):/', '', $errstr );
 	}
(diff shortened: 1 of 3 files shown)


greatislander:
@stepmuel Can you please fix the coding standards errors? You can run `composer standards` within the plugin directory to view.

includes/modules/searchandreplace/class-pb-search.php
```diff
@@ -37,7 +49,18 @@ function search_for_pattern( $search, $limit, $offset, $orderby ) {
 			if ( ! ini_get( 'safe_mode' ) ) {
 				set_time_limit( 0 );
 			}
-			return $this->find( '@' . preg_quote( $search, '@' ) . '@', $limit, $offset, $orderby );
+			if ( $this->regex ) {
+				// First test that the search and replace strings are valid regex
+				set_error_handler( array( &$this, 'regex_error' ) );
+				$valid = @preg_match( $search, '', $matches );
```
greatislander: Can you handle potential errors here rather than suppressing them?

See: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#error-control-operator

templates/admin/search.php
```diff
@@ -53,6 +57,15 @@
 				  <input class="term" type="text" name="replace_pattern" value="<?php echo esc_attr( $replace ) ?>"/><br/>
 				</td>
 			</tr>
+			<tr>
```
greatislander: We would prefer having this off by default, enabled by setting the `PB_ENABLE_REGEX_SEARCHREPLACE` constant to `true`.

includes/modules/searchandreplace/class-pb-search.php
stepmuel: The errors are already being handled via `set_error_handler` (will display the formatted error to the user instead of randomly placing it in the HTML). Removing the `@` doesn't seem to have any effect, but I'd rather keep it for two reasons:

* It makes it clear that potential errors won't generate any error text at that place.
* I got that code extract from https://github.com/johngodley/search-regex/blob/master/models/search.php and I suspect there might be cases where leaving this out would still lead to unwanted behaviour. I don't know for sure why it was put there and am worried I might break something I don't know anything about. (e.g. I had some php installations where errors would always appear on screen twice; maybe its for that case). 

But since it seems to be working without, please let me know if you still want it removed.

connerbw: Interesting.

On Stack Overflow, the [most popular answer for validating a regex](https://stackoverflow.com/a/12941133) uses this same procedure and recommends using `@`

It's probably better to keep it, please add `// @codingStandardsIgnoreLine` where necessary?

stepmuel: `composer standards` doesn't show me any violations. Should I still add the line preemptively? Or is there something wrong with my setup and suppressing errors should trigger a validation error?

greatislander:
@stepmuel We are about to switch to a more strict coding standards ruleset that will enforce this, so would prefer to have this in place when we merge the PR.

includes/modules/searchandreplace/class-pb-search.php
```diff
@@ -14,11 +14,23 @@ class Search {
...
+		$this->regex_error = __( 'Invalid regular expression', 'pressbooks' ) . ': ' . preg_replace( '/(.*?):/', '', $errstr );
+	}
+
 	function search_and_replace( $search, $replace, $limit, $offset, $orderby, $save = false ) {
+		// escape potential backreferences when not in regex mode
+		if ( ! $this->regex ) {
+			$replace = str_replace( '\\', '\\\\', $replace );
+			$replace = str_replace( '$', '\\$', $replace );
+		}
 		$this->replace = $replace;
```
connerbw: Lets the `/e` modifier through which allows users to execute arbitrary code on the server?

@see https://bitquark.co.uk/blog/2013/07/23/the_unexpected_dangers_of_preg_replace

includes/modules/searchandreplace/class-pb-search.php
```diff
@@ -37,7 +49,19 @@ function search_for_pattern( $search, $limit, $offset, $orderby ) {
 			if ( ! ini_get( 'safe_mode' ) ) {
 				set_time_limit( 0 );
 			}
-			return $this->find( '@' . preg_quote( $search, '@' ) . '@', $limit, $offset, $orderby );
+			if ( $this->regex ) {
+				// First test that the search and replace strings are valid regex
+				set_error_handler( array( &$this, 'regex_error' ) );
+				// @codingStandardsIgnoreLine
+				$valid = @preg_match( $search, '', $matches );
```
connerbw: Lets null bytes through which allows users to execute arbitrary code on the server?

@see https://bitquark.co.uk/blog/2013/07/23/the_unexpected_dangers_of_preg_replace

stepmuel: @connerbw thanks for noticing! I'm working on a patch…

stepmuel: This turned out to be more tricky than I imagined… but I did extensive testing and my code did hold up. (I even found undocumented behaviour of `preg_*` while testing all possible characters at various positions.)

connerbw:
Merged into dev (but forgot to squash... doh!) Thanks for this.

@greatislander: How do we communicate the new `PB_ENABLE_REGEX_SEARCHREPLACE` const to users?

greatislander:
Will add to docs.

PR #755, opened 2017-05-23
Use template for pb_private() function

stepmuel:
This allows themes to overwrite the standard login prompt by providing their own private.php template.

### DIFF
--- a/themes-book/pressbooks-book/functions.php
+++ b/themes-book/pressbooks-book/functions.php
@@ -196,30 +196,7 @@ function pb_get_links($echo=true) {
  * Prevent access by unregistered user if the book in question is private.
  */
 function pb_private() {
-	$bloginfourl = get_bloginfo( 'url' ); ?>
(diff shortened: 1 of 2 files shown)


greatislander:
@stepmuel I have no objection to merging this PR. However, you should be aware that we are about to completely rebuild the core book theme, so this will likely be deprecated fairly quickly.

stepmuel:
@greatislander as long as there will be a releases with this included, it will help me. It's the only reason I can't use official releases, and it makes updating my installation complicated and error prone. 

Is there any more information about the book theme rebuild, like approximate schedule, a branch or a list of planned changes? I was about to migrate my theme to SCSS, but am now considering to wait.

greatislander:
We will be working on the rebuild during June and July. More announcements will be forthcoming, but it will be an extensive visual refresh and will also focus on improving accessibility for readers on the web.

PR #776, opened 2017-05-31
Revert 300dpi option , try `auto,normal` (fix #744)

### LINKED ISSUE #744: Remove 300dpi image resolution option and forced 300dpi images from print PDF
### Prerequisites

### Description

Forcing 300dpi doesn't work well. Sometimes the uploaded images are too small and they get downscaled, which is confusing and frustrating. We should remove the 300dpi image resolution option and corresponding SCSS output (as well as the forced version in Print PDF outputs) and seek a better solution for guiding users to upload print-ready images. (See also #497, #640.)

### Steps to Reproduce

Not applicable.

### System Information

Pressbooks: 3.9.9-dev
WordPress: 4.7.4
PHP: 7.1

connerbw:
prince-image-resolution syntax:
+ https://www.princexml.com/doc/properties/prince-image-resolution/

Original 300dpi commits:
+ b15be7e
+ 0b546fb


inc/modules/export/prince/class-printpdf.php
```diff
@@ -85,9 +85,9 @@ protected function themeOptionsOverrides() {
 		$sass = \Pressbooks\Container::get( 'Sass' );
 
 		if ( $sass->isCurrentThemeCompatible( 2 ) ) {
-			$extra = "/* Print Overrides */\n\$prince-image-resolution: 300dpi; \n";
+			$extra = "/* Print Overrides */\n\$prince-image-resolution: auto,normal \n";
```
greatislander: We don't need to output this, do we? The default will be set in [assets/book/styles/variables/_media.scss](https://github.com/pressbooks/pressbooks/blob/05ac44820f1e582dafa89af9ab60521437252e30/assets/book/styles/variables/_media.scss#L43) and this doesn't actually change it.

greatislander: Now that I consider further, we can remove the `themeOptionsOverrides()` method entirely as it should inherit the parent method: https://github.com/pressbooks/pressbooks/blob/05ac44820f1e582dafa89af9ab60521437252e30/inc/modules/export/prince/class-printpdf.php#L80-L119

inc/modules/themeoptions/class-pdfoptions.php
```diff
@@ -1663,7 +1634,7 @@ static function scssOverrides( $scss ) {
 		if ( isset( $options['pdf_image_resolution'] ) ) {
 			$resolution = $options['pdf_image_resolution'];
```
greatislander: We should probably remove this whole block. If someone has saved `pdf_image_resolution` and this stays, they will be stuck because the option can't be changed via the interface anymore.

inc/modules/themeoptions/class-pdfoptions.php
```diff
@@ -1391,7 +1362,7 @@ static function getDefaults() {
 			'pdf_paragraph_separation' => 'indent',
 			'pdf_sectionopenings' => 'openauto',
 			'pdf_toc' => 1,
-			'pdf_image_resolution' => '300dpi',
+			'pdf_image_resolution' => 'auto,normal',
```
greatislander: We can remove this from the defaults, can't we? Since we don't have any way to change it anymore and it's already set in the base SCSS.

PR #802, opened 2017-06-20
adding utility function, test, filter for import template

bdolor:
Makes import template more flexible for developers.

### DIFF
--- a/inc/utility/namespace.php
+++ b/inc/utility/namespace.php
@@ -883,3 +883,25 @@ function rcopy( $src, $dest ) {
 	}
 	return true;
 }
+
+/**
+ * Helper function to transform a one dimensional array to csv
+ *
+ * @since 3.9.10
+ *
+ * @param $any_array
+ *
+ * @return string
+ */
+function array_to_csv( $any_array ) {
+	$result = '';
+
+	if ( is_array( $any_array ) ) {
+		foreach ( $any_array as $item ) {
+			$result .= $item . ', ';
+		}
+		$result = rtrim( $result, ', ' );
+	}
+
+	return $result;
+}
--- a/inc/utility/namespace.php
+++ b/inc/utility/namespace.php
@@ -883,25 +883,3 @@ function rcopy( $src, $dest ) {
 	}
 	return true;
 }
-
-/**
- * Helper function to transform a one dimensional array to csv
- *
- * @since 3.9.10
- *
- * @param $any_array
- *
- * @return string
- */
-function array_to_csv( $any_array ) {
-	$result = '';
-
-	if ( is_array( $any_array ) ) {
-		foreach ( $any_array as $item ) {
-			$result .= $item . ', ';
-		}
-		$result = rtrim( $result, ', ' );
-	}
-
-	return $result;
-}
--- a/templates/admin/import.php
+++ b/templates/admin/import.php
@@ -24,6 +24,8 @@ $import_option_types = apply_filters( 'pb_select_import_type', array(
 	'html' => __( 'HTML (scrape content from a URL)' ),
 ) );
 
+$supported_file_extensions = \Pressbooks\Utility\array_to_csv( array_keys( $import_option_types ) );
+
 ?>
 <div class="wrap">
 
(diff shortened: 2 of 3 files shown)


connerbw:
Cool! 

I suggest renaming  `array_to_csv` to `array_to_cds` (comma delimited string) or some better idea.

When I think of CSV in PHP I think of  `fputcsv`. This function supports different delimiters other than comma, allows "This,is,a,string" to be passed as a single string in the array and escapes, escapes spaces and quotes, etc.

Or, [you could use one of these ideas ](https://stackoverflow.com/questions/4692709/how-to-make-fputcsv-echo-the-data)to just pipe to `fputcsv` itself? You might have to change your unit test.

Example:

```
function array_to_csv( array $any_array ) {
  $result = fopen('php://output', 'w');
  fputcsv($result , $any_array);
  fclose($result );
  return $result;
}
```

PS: Why does the function checks `if ( is_array( $any_array ) ) {` ? It should just horribly fail if the developer can't figure out they should pass an array to this function?

Thank you for your consideration.

bdolor:
using `fputcsv` in combination with `fopen('php://output', 'w') writes to the browser, early...which would look like: 
image

You'll notice a call to `strtoupper` is bypassed and it places itself on the page, outside of the markup, where the variable is `echoed`. 
image

The reason for calling `is_array` is to both to set up/ensure `foreach` has something to iterate over and give it a condition to degrade gracefully. If you'd prefer type hinting arguments, which will give you the errors you're looking for, that's okay. I don't see a general movement towards type hinting in the code base, so I just followed suit. 

I'm not particular about what to call the function. CSV is what the output of the function is, so it made sense to me to call it that. If you prefer CDS, that's fine. CSV might be more familiar to people, iz all.

greatislander:
Not to confuse things further but wouldn't this do the same thing? I'd rather do this than adding another utility function, if it's all the same to you.

```
$string = implode( ', ', array_keys( $array) );
```

connerbw:
> using fputcsv in combination with `fopen('php://output', 'w') writes to the browser, early...

Ouch, well then no to that idea. Sorry. 

My point was more about CSV, and what I expect a function with CSV in the name to do.

bdolor:
so, if given the option of either using a short, testable, reusable function vs inline code, you guys would prefer inline code?

templates/admin/import.php
```diff
@@ -24,6 +24,8 @@
 	'html' => __( 'HTML (scrape content from a URL)' ),
 ) );
 
+$supported_file_extensions = \Pressbooks\Utility\array_to_csv( array_keys( $import_option_types ) );
```
greatislander: Please replace this with:

```
+$supported_file_extensions = implode( ', ', array_keys( $import_option_types ) );
```

inc/utility/namespace.php
```diff
@@ -883,3 +883,25 @@ function rcopy( $src, $dest ) {
 	}
 	return true;
 }
+
+/**
+ * Helper function to transform a one dimensional array to csv
```
greatislander: Not needed, see below.

tests/test-utility.php
```diff
@@ -253,4 +253,14 @@ public function test_rcopy() {
 		$return = \Pressbooks\Utility\rcopy( trailingslashit( $uploads['path'] ) . 'missing', $dest );
 		$this->assertEquals( $return, false );
 	}
+
+	public function test_array_to_csv(){
```
greatislander: Not needed, see above.

connerbw:
> so, if given the option of either using a short, testable, reusable function vs inline code, you guys would prefer inline code?

```
$one_dimensional = ['val1', 'val,2', 'val3'];
$return = array_to_csv( $one_dimensional );
// Assertion?

$one_dimensional = ['val1', 'val2    ', 'val3'];
$return = array_to_csv( $one_dimensional );
// Assertion?

$one_dimensional = ['val1', \StdClass(), 'val3'];
$return = array_to_csv( $one_dimensional );
// Assertion?

$one_dimensional = ['val1', [ 'val2', 'val3' ], 'val4'];
$return = array_to_csv( $one_dimensional );
// Assertion?
```

The issue is that the proposed function is basically `rtrim( implode(', ' $array), ',');` which IMHO doesn't really have much to do with CSV.

Implode is well known and [tested](https://github.com/php/php-src/blob/master/ext/standard/tests/strings/implode.phpt).

Regards,

PR #811, opened 2017-06-21
allow for hook into setCurrentImportOptions if import is HTTP GET …

bdolor:
Using the hook `pb_initialize_import` to add a custom import routine which uses HTTP GET instead of a file upload, is prevented by the existing logic from ever reaching `setCurrentImportOption` because of the existing condition `'html' == $_POST['type_of']`. https://github.com/pressbooks/pressbooks/blob/dev/inc/modules/import/class-import.php#L388

The new condition checks for the existence of a generic `$_POST['import_http']` (updated in the template file as well)  instead of the more specific `$_POST['type_of']` which has been moved further down. Generic checks for valid url input, server response errors, and response code checking will apply to all HTTP GET imports. Similar to uploading a file, the switch statement now looks for `$_POST['type_of']`. HTML scraping logic/checks have been moved into the html-specific switch statement, allowing for generic checks to apply to the custom import type without requiring them, for instance, to return a content-type of `text/html`.

### DIFF
--- a/inc/modules/import/class-import.php
+++ b/inc/modules/import/class-import.php
@@ -430,7 +430,7 @@ abstract class Import {
 					 * Allows users to add custom import routine for custom import type
 					 * via HTTP GET requests
 					 *
-					 * @since 3.9.10
+					 * @since 4.0.0
 					 */
 					$importer = apply_filters( 'pb_initialize_import', null );
 					if ( is_object( $importer ) ) {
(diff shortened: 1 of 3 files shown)


inc/modules/import/class-import.php
```diff
@@ -409,38 +410,37 @@ static public function formSubmit() {
...
+					break;
 
-			$importer = new Html\Xhtml();
-			$ok = $importer->setCurrentImportOption( $body );
+				default:
+					/**
+					 * Allows users to add custom import routine for custom import type
+					 * via HTTP GET requests
+					 *
+					 * @since 3.9.10
```
greatislander: This will be `@since 4.0.0`.

templates/admin/import.php
```diff
@@ -191,7 +191,7 @@ function(){
 					 * @since 3.9.10
```
greatislander: Sorry, should have caught this before—will also be `@since 4.0.0`.

PR #827, opened 2017-06-30
Preview chapter scaffolding 

connerbw:


connerbw:
Thinking about this a bit:

 + Check if the export dependencies are installed?
 + Does mdpf use /formats/xhtml ? If yes, it needs to be updated like in this PR. (also: https://github.com/pressbooks/pressbooks/commit/763c7de8b7d85cd87622013070f9921f6937ccc3)

greatislander:
mpdf doesn't use XHTML. See: https://github.com/BCcampus/pressbooks-mpdf/issues/13

connerbw:
This PR triggered an interesting internal discussion. We agreed it needs more product development. The PR has been modified to remove UX elements. The Book class refactoring remains, so devs can use and build on it.

PR #832, opened 2017-07-05
Metadata improvements (fixes #804)

### LINKED ISSUE #804: Expand book information with setting to show a more limited range of options
### Prerequisites

### Description

We need to offer some more metadata in Book Information. But we also don't want to overload people, so we should add a setting which hides the expanded metadata for users who don't want or need it.

### Steps to Reproduce

Not applicable.

### System Information

Pressbooks: 4.0.0-dev
WordPress: 4.8.0
PHP: 7.1

greatislander:
This PR:

* Improves metadata API endpoint, adding audience, ISBN, and full BISAC subject support;
* Refactors metadata class to implement [JsonSerializable](http://php.net/manual/en/class.jsonserializable.php), moving static functions to a namespace;
* Hides a bunch of book info fields that may not be needed for ordinary users (fixes #804);
* Adds an option to show expanded book information for advanced users;
* Fixes book license display via metadata endpoint;
* Adds support for multiple editors and translators;
* Adds sensible fallbacks for `pb_copyright_date` and `pb_author_file_as`.


inc/modules/export/epub/templates/epub201/opf.php
```diff
@@ -45,6 +47,10 @@
 		echo '<dc:creator opf:role="aut"';
 		if ( ! empty( $meta['pb_author_file_as'] ) ) {
 			echo ' opf:file-as="' . $meta['pb_author_file_as'] . '"';
+		} else {
+			$nameparser = new Parser();
+			$author = $nameparser->parse( $meta['pb_author'] );
```
connerbw: This library [can throw exceptions](https://github.com/davidgorges/HumanNameParser.php/blob/master/src/HumanNameParser/Parser.php#L188). What happens if it's a bad name? Try/Catch?

inc/modules/export/epub/templates/epub3/opf.php
```diff
@@ -69,9 +71,11 @@
 		if ( ! empty( $meta['pb_author_file_as'] ) ) {
 			echo $meta['pb_author_file_as'];
 		} elseif ( ! empty( $meta['pb_author'] ) ) {
-			echo $meta['pb_author'];
+			$nameparser = new Parser();
+			$author = $nameparser->parse( $meta['pb_author'] );
```
connerbw: This library [can throw exceptions](https://github.com/davidgorges/HumanNameParser.php/blob/master/src/HumanNameParser/Parser.php#L188). What happens if it's a bad name? Try/Catch?

inc/metadata/namespace.php
```diff
@@ -0,0 +1,315 @@
...
+			'pb_list_price_web',
+			'pb_audience',
+			'pb_bisac_subject',
+			'pb_bisac_regional_theme',
+			'pb_catalogue_order',
+		];
+		foreach ( $additional_fields as $field ) {
+			if ( isset( $metadata[ $field ] ) && ! empty( $metadata[ $field ] ) ) {
+				update_option( 'pressbooks_show_expanded_metadata', 1 );
+				return true;
```
connerbw: If a user sets `pb_list_price_epub`, then sets `$_GET['pressbooks_show_expanded_metadata'] = 0`, then reloads the Book Information page, we end up in this code block and return true.

Is this intentional?

tests/test-metaboxes.php
```diff
@@ -29,8 +29,6 @@ public function test_update_font_stacks() {
 		$this->assertArrayHasKey( 'part', $wp_meta_boxes );
 		$this->assertArrayHasKey( 'metadata', $c->metadata );
 		$this->assertArrayHasKey( 'general-book-information', $c->metadata['metadata'] );
-		$this->assertArrayHasKey( 'additional-catalogue-information', $c->metadata['metadata'] );
```
connerbw: Instead of removing this test, can we add another test that checks if its hidden or not? (Do `update_option( 'pressbooks_show_expanded_metadata' 1);` and test if `additional-catalogue-information` is there.)

inc/metadata/namespace.php
```diff
@@ -0,0 +1,315 @@
...
+	} else {
+		$text = __( 'Show Additional Book Information', 'pressbooks' );
+		$href = wp_nonce_url( $url . '&pressbooks_show_expanded_metadata=1', 'pb-expanded-metadata' );
+	}
+
+	?>
+	<div id="expanded-metadata-panel" class="postbox">
+		<div class="inside">
+			<p><?php _e( 'The book information you enter here appears on your book’s cover and title pages and in the metadata of your webbook and exported files.', 'pressbooks' ); ?></p>
+			<?php if ( ! $show_expanded_metadata && $show_expanded_metadata !== null ) { ?>
```
connerbw: `$show_expanded_metadata = show_expanded_metadata();` never returns null?

inc/metadata/namespace.php
```diff
@@ -0,0 +1,315 @@
...
+			<p><a class="button" href="<?php echo $href; ?>"><?php echo $text; ?></a></p>
+		</div>
+	</div>
+	<?php
+}
+
+/**
+ * Should we show expanded metadata fields or not?
+ *
+ * @return bool | null
```
connerbw: Function does not return null.

tests/test-metaboxes.php
greatislander: @connerbw All of these requests have been addressed—want to take one last look before I merge?

inc/modules/export/epub/templates/epub201/opf.php
```diff
@@ -45,6 +47,17 @@
 		echo '<dc:creator opf:role="aut"';
 		if ( ! empty( $meta['pb_author_file_as'] ) ) {
 			echo ' opf:file-as="' . $meta['pb_author_file_as'] . '"';
+		} else {
+			$nameparser = new Parser();
+			try	{
+				$author = $nameparser->parse( $meta['pb_author'] );
+				$author_file_as = $author->getLastName() . ', ' . $author->getFirstName();
+			} catch ( Exception $e ) {
```
connerbw: Should be `\Exception`. Optionally, my preference would be to catch `\HumanNameParser\Exception\NameParsingException` instead of searching the string for the error.

inc/modules/export/epub/templates/epub3/opf.php
```diff
@@ -69,9 +71,17 @@
 		if ( ! empty( $meta['pb_author_file_as'] ) ) {
 			echo $meta['pb_author_file_as'];
 		} elseif ( ! empty( $meta['pb_author'] ) ) {
-			echo $meta['pb_author'];
+			$nameparser = new Parser();
+			try	{
+				$author = $nameparser->parse( $meta['pb_author'] );
+				echo $author->getLastName() . ', ' . $author->getFirstName();
+			} catch ( Exception $e ) {
```
connerbw: Should be `\Exception`. Optionally, my preference would be to catch `\HumanNameParser\Exception\NameParsingException` instead of searching the string for the error.

tests/test-metaboxes.php
connerbw: Done.

PR #866, opened 2017-07-26
Fix: open_basedir restriction

connerbw:
open_basedir restriction forbids file_exists() in /usr, changed to filter instead.


greatislander:
FYI @thomasdumm.

PR #901, opened 2017-08-24
Better image cloning (fixes #895)

### LINKED ISSUE #895: Import/clone full-size images
### Prerequisites

### Description

Both the WXR import module and the cloner pull in downsampled images (they look at HTML and fetch whatever is set in an `` `src` attribute.

Ideas for improvement:

- Use the media endpoint (`wp/v2/media`). Fetch all images, then work backwards into posts, updating `src` attributes where possible.
- Use the image `srcset` attribute to determine the largest image; fetch that instead of the downsampled version, then update the `src` to the local image.
- Try to guess the original image URL from the `src` attribute (if the `src` is https://pressbooks.com/wp-content/uploads/sites/1/2017/01/image-500x500.png then test if https://pressbooks.com/wp-content/uploads/sites/1/2017/01/image.png exists); fetch this image, then update the `src` to the local image.
- When trying to find the best version of a resized image, use [Imagehash](https://github.com/jenssegers/imagehash) to check if it's the closest match (thumbnails, for example, are not use different cropping).

### Steps to Reproduce

1. Clone a book or import a WXR file.
2. Compare source and target images.

**Expected behavior:** The "original" sizes of the images are the same.

**Actual behavior:** The cloned/imported images are downsampled.

### System Information

Pressbooks: 4.1
WordPress: 4.8.1
PHP: 7.1

connerbw:
Approach:

 + Scan `/wp/v2/media` for all known images and build a `[resized] => [fullsize]` list.
 + When saving `` tags, if its hosted at the source, and if we find a resized match from our known images list, then download the fullsize one.
 + When we sideload the fullsize image into PB, it will recreate the same files and filenames, so it should just fit.

I did not check if images are the same because that would mean having to download the image first, then checking, then trying again if it didn't work. Maybe that's OK for a local disk, not so much over HTTP. 

Can be improved in the future.

### DIFF
--- a/inc/class-cloner.php
+++ b/inc/class-cloner.php
@@ -285,10 +285,11 @@ class Cloner {
 		foreach ( $this->sourceBookTerms as $k => $v ) {
 			if ( $v['id'] === absint( $term_id ) ) {
 				$term = $this->sourceBookTerms[ $k ];
+				break;
 			}
 		};
 
-		if ( empty( $term['slug'] )|| empty( $term['taxonomy'] ) ) {
+		if ( empty( $term['slug'] ) || empty( $term['taxonomy'] ) ) {
 			// Doing it wrong...
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


connerbw:
I forgot about WXR. I'm going to keep working on this.

PR #923, opened 2017-09-05
Remove old `latex_path` option handling

jeremyfelt:
This area of code is a leftover from the WP LaTeX plugin. In that plugin, the`latex_path` option and the `AUTOMATTIC_LATEX_LATEX_PATH` constant are used only if the `Automattic_Latex_DVIPNG` class is controlling the LaTeX output. Neither of these are used in the classes used to generate LaTex via WPCOM, which Pressbooks includes.

It also appears that in the fork of that plugin, the admin form inputs that could have controlled the `latex_path` option value were removed. This results in the value always being `false`, which in turn results in `which latex` being run on the server every time the `PBLatexAdmin` class is started.

I don't believe there is any harm in removing these. Even if a custom LaTex render was being used, the `latex_path` option would likely be unavailable and managed separately.


connerbw:
>  This results in the value always being false, which in turn results in which latex being run on the server every time the PBLatexAdmin class is started.

Ouch. Thanks.

greatislander:
Thanks @jeremyfelt!

PR #928, opened 2017-09-08
allow other latex renderers to be used for pdf exports

monkecheese:
When attempting a PDF export, if the selected latex renderer was
being loaded from outside of the pressbooks codebase, the export
would fail with an error saying that no such latex render file
exists. This solution uses the `pb_require_latex` filter hook to
reference/load the appropriate latex renderer if it lives outside
of the pressbooks codebase (e.g., another plugin).

### DIFF
--- a/symbionts/pressbooks-latex/pb-latex.php
+++ b/symbionts/pressbooks-latex/pb-latex.php
@@ -130,14 +130,29 @@ class PBLatex {
 	}
 
 	function &latex( $latex, $background = false, $color = false, $size = 0 ) {
-		if ( empty( $this->methods[$this->options['method']] ) ) return false;
+		$method = $this->methods[$this->options['method']];
+
+		if ( empty( $method ) ) return false;
 
 		if ( ! $background )
 				$background = empty( $this->options['bg'] ) ? 'ffffff' : $this->options['bg'];
 		if ( ! $color )
 				$color = empty( $this->options['fg'] ) ? '000000' : $this->options['fg'];
 
-		require_once( dirname( __FILE__ ) . "/automattic-latex-{$this->methods[$this->options['method']]}.php" );
+		if ( 'wpcom' == $method ) {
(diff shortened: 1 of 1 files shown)


symbionts/pressbooks-latex/pb-latex.php
```diff
@@ -137,7 +137,12 @@ function &latex( $latex, $background = false, $color = false, $size = 0 ) {
 		if ( ! $color )
 				$color = empty( $this->options['fg'] ) ? '000000' : $this->options['fg'];
 
-		require_once( dirname( __FILE__ ) . "/automattic-latex-{$this->methods[$this->options['method']]}.php" );
+		if ( file_exists( dirname( __FILE__ ) . "/automattic-latex-{$this->methods[$this->options['method']]}.php" ) ) {
+			require_once( dirname( __FILE__ ) . "/automattic-latex-{$this->methods[$this->options['method']]}.php" );
+		} else {
+			apply_filters( 'pb_require_latex', $this->methods[$this->options['method']] );
```
greatislander: Needs documentation, as per: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

Additionally I'm wondering if it might be better to call this `pb_require_latex_renderer` for clarity. Otherwise looks good! 👍

greatislander: Actually, scratch that. I realize the hook is already documented elsewhere. On an unrelated note it might be a good idea to do the same thing that we do [here](https://github.com/pressbooks/pressbooks/blob/dev/symbionts/pressbooks-latex/pb-latex-admin.php#L197-L199), e.g.

```
if ( 'wpcom' == method) {
  require_once( dirname( __FILE__ ) . "/automattic-latex-wpcom.php" );
} else {
```

monkecheese:
Updated PR w/ your suggested changes.  Happy to fix anything else, so let me know if there's anything else that jumps out at you.  :)

PR #956, opened 2017-09-27
Remove hardcoded emails, fix filter/action comments

connerbw:


inc/utility/namespace.php
```diff
@@ -752,6 +760,15 @@ function email_error_log( $emails, $subject, $message ) {
...
 	);
 
+	/**
+	 * Filter an array of email addresses error logs are sent to
+	 *
+	 * @since 4.3.3
+	 *
+	 * @param array $emails
+	 */
+	$emails = apply_filters( 'pb_email_error_report', $emails );
```
greatislander: Sorry, bad at naming things. Could we call this filter:

`pb_error_log_emails`

?

connerbw: Ok!

PR #961, opened 2017-10-05
Add filter hook to allow users to brand the catalog page

monkecheese:
Instead of making the title 'Pressbooks' on the catalog static, users have the option to change it to reflect their needs.


monkecheese:
Done.

Issue #964, opened 2017-10-06
Custom import option not giving file upload option


monkecheese:
### Prerequisites

_Check the boxes below by changing them from `[ ]` to `[x]`._

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

In the Import area, we have a custom import method we use.  When we select it in the list of import options, we are supplied a field that wants a url.  Before, we had an option to upload a file from our computer.

### Steps to Reproduce

1. Navigate to Admin > Tools > Import
2. Select a custom option from the drop down
3. The next field is not a file upload field as we expect for our custom import stuff to work

**Expected behavior:** We expect a file upload option

**Actual behavior:** We are given an input field that wants `https://file-to-import.com`

### System Information

Pressbooks: 4.3.3
WordPress: 4.8.2
PHP: 7.0

monkecheese:
The problem is in this bit of javascript that show/hides the options based on what import type is selected:

```
// approximately line 140 in import.php

<script type="text/javascript">
	jQuery(function ($) {
		$('#pb-www').hide();

		$(".pb-html-target").change(
			function () {
				var val = $('.pb-html-target').val();

				if (val == 'wxr' || val == 'epub' || val == 'odt' || val == 'docx') {
					$('#pb-file').show();
					$('#pb-www').hide();
					// clear http value at input elem
					$('.widefat').val('');
				} else {
					$('#pb-file').hide();
					$('#pb-www').show();

				}

			});

	});
</script>
```

I can think of a hacky way of the top of my head to solve this, but I'd rather formulate a clean solution.  Any thoughts?

bdolor:
with this filter you can modify the contents of the default table cell - perhaps that will work? 

https://github.com/pressbooks/pressbooks/blob/dev/templates/admin/import.php#L195

monkecheese:
@bdolor, heh...sort of?  It adds the field to the same row in the table.

[...]

PR #967, opened 2017-10-10
Allow users to choose upload field for custom import choices

monkecheese:
The current logic forces the user to user an input filed that
wants an http link to the file. But there are some situations
where users need to have the flexibility to choose what input
type they use for their custom import type.  This allows users
to use the `pb_import_table_cell` hook to render the input field
of choice. This also introduces an element id `pb-other` that
shows/hides the field depending on what import type is selected.

### DIFF
--- a/templates/admin/import.php
+++ b/templates/admin/import.php
@@ -132,6 +132,7 @@ $supported_file_extensions = implode( ', ', array_keys( $import_option_types ) )
 		<script type="text/javascript">
 			jQuery(function ($) {
 				$('#pb-www').hide();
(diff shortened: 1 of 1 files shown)


bdolor:
In situations where there is more than one `#pb-other` what will the behaviour be?

monkecheese:
Yeah, my solution doesn't account for that scenario.  Closing PR per @bdolor's solution found [here](https://github.com/pressbooks/pressbooks/issues/964#issuecomment-335686365).  Thanks!

PR #986, opened 2017-11-16
Add migration routine for Luther

greatislander:


inc/theme/namespace.php
```diff
@@ -130,6 +130,20 @@ function migrate_book_themes() {
 		}
 
 		update_option( 'pressbooks_theme_migration', 1 );
+	} elseif ( get_option( 'pressbooks_theme_migration' ) === 1 ) {
```
connerbw: I'm not sure it should be `elseif` here.

This works for existing books, or books already migrated to state 1, but on a new book or a book that has been in idle mode for many months getting into this `elseif` condition will require two page loads. (Per your unit tests the first page load will set `pressbooks_theme_migration` to 1, the second to 2.)

Maybe a separate `if` condition so that if a migration happens then it happens all at once?

tests/test-theme.php
```diff
@@ -1,15 +1,25 @@
...
 
-	public function test_update_template_root() {
+    public function test_migrate_book_themes() {
+        $this->_book();
+        delete_option( 'pressbooks_theme_migration' );
+        \Pressbooks\Theme\migrate_book_themes();
+        $this->assertEquals( 1, get_option( 'pressbooks_theme_migration' ) );
+        \Pressbooks\Theme\migrate_book_themes();
+        $this->assertEquals( 2, get_option( 'pressbooks_theme_migration' ) );
+    }
```
connerbw: [Triggered on `init`](https://github.com/pressbooks/pressbooks/blob/2b217032b61be032b59ca44865fed6d5fd07f4be/hooks.php#L253) so two page loads required to migrate? 

See previous comment.

greatislander:
@connerbw Good call. After I wrote the test I was thinking of refactoring the function so it takes the value of `get_option( 'pressbooks_theme_migration' )` as input and then has different `if` statements depending on which value is retrieved.

PR #988, opened 2017-11-22
update to OEBPS, Open eBook Publication Structure

bdolor:
cosmetic syntax error, now less likely to induce feelings of momentary dyslexia


greatislander:
Thanks @bdolor!

PR #1011, opened 2017-12-07
New Post Statii, LicenseUpgrade Routine

connerbw:
First pass at Post Statii. They are registered and tested. I have questions about saving them.

An uprade routine to add license terms to existing books for good measure.


connerbw:
Working on contributors in this branch to avoid merge conflicts.

PR #1031, opened 2018-01-06
Contributor Class

connerbw:
Consolidated new Contributor Data Model into a single class.

Instead of `get_contributors()` it's `$contributor->get()`. Tell me if you would prefer it pluralized.

Has "link" method so that it's easier (for me) to set it and forget it when it comes to combining Taxonomy + Meta. Has other reusable stuff.

### DIFF
--- a/inc/class-contributor.php
+++ b/inc/class-contributor.php
@@ -1,0 +1,235 @@
+<?php
+/**
+ * @author  Pressbooks <code@pressbooks.com>
+ * @license GPLv2 (or any later version)
+ */
+
+namespace Pressbooks;
+
+use function Pressbooks\Utility\str_starts_with;
+
+class Contributor {
+
+	/**
+	 * Valid contributor slugs
+	 *
+	 * @var array
+	 */
+	public $valid = [
+		'pb_authors',
+		'pb_editors',
+		'pb_translators',
+		'pb_proofreaders',
+		'pb_reviewers',
+		'pb_illustrators',
+		'pb_contributors',
+	];
+
+	public function __construct() {
+	}
+
+	/**
+	 * Retrieve all author/editor/etc lists for a given Post ID
+	 *
+	 * @param int $post_id
+	 *
+	 * @return array
+	 */
+	public function getAll( $post_id ) {
+		$contributors = [];
+		foreach ( $this->valid as $contributor_type ) {
+			$contributors[ $contributor_type ] = $this->get( $post_id, $contributor_type );
+		}
+		return $contributors;
+	}
+
+	/**
+	 * Retrieve author/editor/etc lists for a given Post ID and Contributor type
+	 *
+	 * @param int Post ID or object.
+	 * @param string $contributor_type
+	 *
+	 * @return string
+	 */
+	public function get( $post_id, $contributor_type ) {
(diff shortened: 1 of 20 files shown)


inc/class-contributor.php
```diff
@@ -0,0 +1,235 @@
...
+							if ( $added !== false ) {
+								$contributors[] = $name;
+							}
+						}
+					}
+				}
+			}
+		}
+
+		// Comma separated, Oxford comma, localized and between the last two items
```
greatislander: This would probably be good to turn into a function in `\Pressbooks\Utility\` (?) so that we can make a formatted, comma separated list with Oxford commas out of any array of items.

inc/class-contributor.php
```diff
@@ -0,0 +1,235 @@
...
+
+		if ( $post_id && is_array( $results ) ) {
+			$this->link( $results['term_id'], $post_id, $contributor_type );
+		}
+
+		return is_array( $results ) ? $results : false;
+	}
+
+	/**
+	 * Associate a Contributor's Term ID to a Post ID (Taxonomy + Meta)
```
greatislander: Technically we are assigning the Term Slug to the Post ID? Maybe good to clarify in the code comment?

inc/class-taxonomy.php
```diff
@@ -649,4 +552,31 @@ protected function upgradeChapterTypes() {
...
+	 * @param string $old_meta_key Meta key.
+	 * @param mixed $meta_value Meta value.
+	 *
+	 * @return array|false An array containing the `term_id` and `term_taxonomy_id`, false otherwise.
+	 */
+	public function upgradeToContributorTaxonomy( $meta_id, $post_id, $old_meta_key, $meta_value ) {
+		$contributor_migration = [
+			'pb_author' => 'pb_authors',
+			'pb_section_author' => 'pb_authors',
+			'pb_editor' => 'pb_editors',
```
greatislander: `pb_editor`, `pb_translator`, and `pb_contributing_authors` all save as arrays. So the `meta_value` will be an array… I believe.

inc/class-taxonomy.php
```diff
@@ -649,4 +552,31 @@ protected function upgradeChapterTypes() {
...
+			'pb_translator' => 'pb_translators',
+			'pb_contributing_authors' => 'pb_contributors',
+		];
+		if ( isset( $contributor_migration[ $old_meta_key ] ) && ! empty( $meta_value ) && is_string( $meta_value ) ) {
```
greatislander: As per above, `$meta_value` should be an array for `pb_editor`, `pb_translator`, and `pb_contributing_authors` so we would need to iterate over it.

inc/class-contributor.php
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * @author  Pressbooks <code@pressbooks.com>
+ * @license GPLv2 (or any later version)
+ */
+
+namespace Pressbooks;
+
+use function Pressbooks\Utility\str_starts_with;
+
+class Contributor {
```
greatislander: Should be pluralized IMHO.

inc/class-taxonomy.php
connerbw: Code now handles both strings and array of strings.

PR #1036, opened 2018-01-08
Add ability to prettify HTML outputs

connerbw:
Debugging feature,


connerbw:
It's intentional. We can't tidy XTHML by default because we don't want to introduce bugs (what do formatting and extra spaces do to existing outputs? Probably nothing but we can't know for sure.). My hope HTMLBook/HTML5 should be tidy from the start.

PR #1120, opened 2018-02-23
Fix when Buckram is updated

connerbw:


inc/class-styles.php
```diff
@@ -543,13 +544,24 @@ public function updateWebBookStyleSheet( $stylesheet = null ) {
...
+			( new ThemeOptions() )->clearCache();
 			$this->updateWebBookStyleSheet();
-			update_option( 'pressbooks_theme_version', $current_version );
+			update_option( 'pressbooks_theme_version', $current_theme_version );
+			return true;
+		}
+
+		// Buckram was updated?
+		$current_buckram_version = $this->getBuckramVersion();
+		$last_buckram_version = get_option( 'pressbooks_buckram_version', $current_buckram_version );
```
greatislander: Set the fallback to `0.1.0` so all books coming from earlier Buckram versions will get their stylesheet recompiled.

connerbw: OK.

greatislander:
TODO: Fix test.

PR #1175, opened 2018-04-05
Prevent slug collisions  (fix #1174)

### LINKED ISSUE #1174: Prevent slug collisions in XHTML and HTMLBook outputs
### Prerequisites

### Description

It is possible for a user to add front matter, chapters, and back matter with identical slugs as WordPress only enforces unique slugs within a post type. This can cause issues in XHTML and consequently PDF as two pieces of content end up with the same ID, which breaks links to the content and causes page numbering issues in the TOC. We should prevent these collisions within the export routine.

### Steps to Reproduce

1. Create front matter and back matter and ensure that they have an identical slug (e.g. `duplicate`).
2. Export to PDF.

**Expected behavior:** TOC page numbering appears as expected, links to front matter and back matter from TOC work.

**Actual behavior:** In the TOC, the back matter title links to the front matter and has the same page number as the front matter. 

### System Information

Pressbooks: 5.2.1
WordPress: 4.9.5
PHP: 7.1

connerbw:
Escape clause:

HTMLawed tries its best [1] but this won't stop a user from generating their own bad content like:

```
<div class="ugc front-matter-ugc">
  <div id="chapter-identical-post-name">Just because</div>
</div>
```

[1] http://www.bioinformatics.org/phplabware/internal_utilities/htmLawed/htmLawed_README.htm#s3.4.2

### DIFF
--- a/inc/modules/export/htmlbook/class-htmlbook.php
+++ b/inc/modules/export/htmlbook/class-htmlbook.php
@@ -645,7 +645,7 @@ class HTMLBook extends Export {
 				$fm->setAttributes(
 					[
 						'class' => "front-matter {$subclass}",
-						'id' => $front_matter['post_name'],
+						'id' => "front-matter-{$front_matter['post_name']}",
 					]
 				);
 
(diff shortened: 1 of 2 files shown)


connerbw:
This PR could be improved by adding section-slugs to `$GLOBALS['hl_Ids']` so that HTMLawed tries to fix the bad user content.

But, I feel that is a higher risk change (new code) and I'm not sure it's worth it.

PR #1216, opened 2018-05-25
Allow usernames to be an email address in catalog rewrite rule 

lukaiser:
Hey guys

Usernames can be email addresses. The current catalog rewrite rule does not allow that.
I added the . and @ to the regex. I hope I haven't missed anything that could complicate matters.

Thanks!

Lukas


connerbw:
Looks OK to me.

PR #1244, opened 2018-06-26
Hide "Export Personal Data" and "Erase Personal Data" from books

connerbw:


greatislander:
@connerbw I think these need to stay, actually.

The scenario: someone comments on my book. Then they decide they want me to remove their personal data from my book. They send me an email. I need access to these pages to complete the request.

connerbw:
WordPress code says that the only people who can see these menus are those with `export_others_personal_data` or `erase_others_personal_data` permissions.

capabilities-1

If we go deeper:

capabilities-2

This, to me, says that on a multisite install, only those with `manage_network` permissions can see that menu.

capabilities-3

So only super admins?

I guess keeping them there is fine since no one will see them. I though the dupe code from a5e1b8fbf (`remove_submenu_page( 'tools.php', 'tools.php' );` twice in a row) was a mistake that was mistake we missed in review and the intention was to remove these tools.

They also bugged me....

PR #1247, opened 2018-06-28
Iframes: Try embed instead.

connerbw:


connerbw:
Related to: https://github.com/BCcampus/pressbooks-openstax-import/issues/67

PR #1258, opened 2018-07-09
Bump minimum WordPress to 4.9.7

connerbw:
Includes security fix so we remove our patch.


greatislander:
@connerbw Looks like something changed in WP 4.9.7 that is causing the Privacy test to fail?

PR #1287, opened 2018-07-19
Attribution

bdolor:
This PR adds a PB standard way to add attributions to images.  

- The input fields are integrated into the media library
- The display of attributions in the content area can be turned on or off (default off) in global theme options
- The html output are formatted according to my best interpretation of [best practices for attribution](https://wiki.creativecommons.org/wiki/Marking_Works_Technical#Attribution) and passes RDFa validation [here](https://www.w3.org/2012/pyRdfa/#distill_by_input)
- They can appear at the bottom of a chapter below the content area and above footnotes: 
- Have been created via shortcode to provide an 'easier' transition path to [Gutenberg](https://wordpress.org/gutenberg/#compatibility). Also for future consideration, the handler `[media_attributions id='33']` generates an individual attribution statement if given the id value of an attachment.  
- exposes metadata that are visible at the `wp-json/wp/v2/media` endpoint.

## Input: ##
image

## Theme Options: ##
image

## Output ##
image

## Individual Output ##
image

## REST API endpoint ##
image


greatislander:
@bdolor If you run:

```
vendor/bin/phpcbf --standard=phpcs.ruleset.xml inc/shortcodes/attributions/class-attachments.php
```

That should fix the sniff error.

greatislander:
And thanks for all your and @alex-418's work on this! Looking forward to reviewing this coming sprint.

bdolor:
@greatislander - it was a pleasure. Please let us know if you need anything else or have any questions.

inc/posttype/namespace.php
```diff
@@ -38,131 +38,131 @@ function list_post_types() {
...
+			'quick_edit'            => false,
+			'capability_type'       => 'post',
+			'has_archive'           => true,
+			'hierarchical'          => false,
+			'menu_icon'             => 'dashicons-book',
+			'supports'              => [ 'title', 'editor', 'author', 'comments', 'page-attributes', 'revisions' ],
+			'show_in_menu'          => false,
+			'show_in_admin_bar'     => true,
+			'show_in_rest'          => true,
+			'rest_base'             => 'chapters',
```
connerbw: [Our coding standards don't enforce this spacing rule](https://github.com/pressbooks/coding-standards/blob/f09f0ba4dd3d20c2acb63c13b28a14bfe86ed229/Pressbooks/ruleset.xml#L27)?

Looking at this part of the diff, and this file in general, it looks like nothing has changed except the spaces for many lines?  I would prefer this kind of alignment be reverted so that the diff is more clear.

For devs, there was a composer vendor update that started enforcing these kinds of spaces, it wasn't the case before. and we made ignoring this rule more explicit.  If this is failing for our current CI tests I'd like to know?

inc/shortcodes/attributions/class-attachments.php
```diff
@@ -0,0 +1,296 @@
...
+				}
+
+				return $allowed_media_types;
+			}
+		);
+
+		// don't show unless user options
+		$options = get_option( 'pressbooks_theme_options_global' );
+
+		if ( 1 === $options['attachment_attributions'] ) {
```
greatislander: I get an `Undefined index` here; maybe check to see if it's set as well as checking the value?

inc/admin/attachments/namespace.php
```diff
@@ -0,0 +1,172 @@
...
+
+		$form_fields['pb_media_attribution_figure'] = [
+			'value' => isset( $figure ) ? $figure : '',
+			'label' => __( 'Figure #', 'pressbooks' ),
+			'input' => 'text',
+		];
+
+		$form_fields['pb_media_attribution_title_url'] = [
+			'value' => isset( $source ) ? $source : '',
+			'label' => __( 'Source', 'pressbooks' ),
```
greatislander: For clarity, I'd suggest labelling this 'Source URL'.

inc/admin/attachments/namespace.php
```diff
@@ -0,0 +1,172 @@
...
+			'input' => 'html',
+			'html'  => "<input type='url' class='text urlfield' placeholder='https://creativecommons.org/' name='attachments[$post->ID][pb_media_attribution_title_url]' value='" . esc_attr( $source ) . "' />",
```
greatislander: I have mixed feelings about the placeholders. When I first opened a media attachment to edit, I thought for a split second that default information had been inserted. I think it might be best to omit placeholders for the URL fields unless the other fields will have them too, and rely on the field labels for clarity (as recommended at [this Mozilla Developer article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Labels_and_placeholders)).

inc/admin/attachments/namespace.php
```diff
@@ -0,0 +1,172 @@
...
+		$adapted_url = get_post_meta( $post->ID, 'pb_media_attribution_adapted_url', true );
+
+		$form_fields['pb_attribution'] = [
+			'value' => '',
+			'label' => __( 'ATTRIBUTIONS', 'pressbooks' ),
+			'input' => 'html',
+			'html'  => '<span></span>',
+		];
+
+		$form_fields['pb_media_attribution_figure'] = [
```
greatislander: My preference would be to remove the figure number field and consider it as a separate issue; dynamic numbering of figures would probably be a better user experience so if this isn't required for the functionality contained in this PR it would be best to omit it (see also: https://github.com/pressbooks/ideas/issues/114).

inc/shortcodes/attributions/class-attachments.php
```diff
@@ -0,0 +1,296 @@
...
+			foreach ( $attributions as $attribution ) {
+
+				// unset empty arrays
+				$attribution = array_filter( $attribution, 'strlen' );
+
+				// only process if non-empty
+				if ( count( $attribution ) > 0 ) {
+					$author_byline  = isset( $attribution['author'] ) ? __( ' by ', 'pressbooks' ) : '';
+					$adapted_byline = isset( $attribution['adapted'] ) ? __( ' adapted by ', 'pressbooks' ) : '';
+					$license_prefix = isset( $attribution['license'] ) ? ' &copy; ' : '';
```
greatislander: Is it correct to use the copyright symbol even with CC licenses? I'm not sure, just curious.

inc/admin/attachments/namespace.php
alex-418: Hmm, yes so the reason behind the placeholder text was to provide an idea of what format the field expected for the URL. If it's not entered in a specific way it doesn't get saved, and there is no error message. Any suggestions on how to improve that, or would you still prefer this placeholder gone?

greatislander: Looking at the filter documentation for `attachment_fields_to_edit`, I found a link to this tutorial which suggests that validation errors can be added and displayed via the `attachment_fields_to_save` hook: https://code.tutsplus.com/articles/creating-custom-fields-for-attachments-in-wordpress--net-13076

I think that would be my preferred approach. It looks like you can also add an array key called `'helps'` to each field with descriptive text which would appear below the field, which might be a good way to provide additional context prior to entry and validation.

alex-418: thanks for that quick feedback, we couldn't get the validation errors to work, but I will try again. Failing that, will use the descriptive text as suggested.

inc/shortcodes/attributions/class-attachments.php
alex-418: ¯\_(ツ)_/¯

inc/admin/attachments/namespace.php
```diff
@@ -0,0 +1,167 @@
...
+ * @param $form_fields
+ * @param $post
+ *
+ * @return mixed
+ */
+function add_metadata_attachment( $form_fields, $post ) {
+
+	if ( substr( $post->post_mime_type, 0, 5 ) === 'image' ) {
+
+		$author      = get_post_meta( $post->ID, 'pb_media_attribution_author', true );
```
greatislander: @alex-418 I think in your most recent changes the equals signs got aligned again; this isn't part of the Human Made coding standards we like so if you could revert this (so one space on either side of the `=` in variable assignments) that would be great. Thanks in advance!

inc/shortcodes/attributions/class-attachments.php
greatislander: Looks like it is correct: https://creativecommons.org/2011/04/15/plaintext-versions-of-creative-commons-licenses-and-cc0/

greatislander:
Thanks all! 🎉

PR #1299, opened 2018-07-30
Add missing IDs to attribution URL fields

greatislander:
This PR adds IDs to attribution fields with custom HTML (URLs) to ensure that clicking the label focuses the associated field.


greatislander:
FYI @alex-418 and @bdolor. We caught this in further review of the merged PR.

PR #1309, opened 2018-08-07
Improve register_meta() usage with object subtypes

greatislander:
See: https://make.wordpress.org/core/2018/07/27/registering-metadata-in-4-9-8/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ php:
 - 7.2
 - nightly
 env:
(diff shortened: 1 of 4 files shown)


greatislander:
In 4.9.8 the global `$wp_meta_keys` array has an extra level of nesting, so our test needed to be updated to look for this:

`$wp_meta_keys['post']['chapter']['pb_short_title']`

PR #1310, opened 2018-08-07
 Add fixInternalLinks() method to cloner (fixes #1279)

### LINKED ISSUE #1279: Internal links are not properly updated when cloning
### Prerequisites

### Description

Generally, internal links are not guaranteed to work across cloning operations at present. We should ensure that they work in all of the following conditions:

- [x] Links to content that begin with the source book URL should be updated to the target book URL
- [x] Relative and absolute internal links in subdirectory source books should be updated so that they work in both subdomain and subdirectory target books
- [x] Relative and absolute internal links in subdomain source books should be updated so that they work in both subdomain and subdirectory target books

### Steps to Reproduce

Not applicable.

### System Information

Pressbooks: 5.4.1

greatislander:
This PR adds a new `fixInternalLinks()` method to the Cloner class, which does exactly what you'd expect.

### DIFF
--- a/inc/class-cloner.php
+++ b/inc/class-cloner.php
@@ -17,6 +17,9 @@ use function Pressbooks\Utility\getset;
 use function Pressbooks\Utility\oxford_comma_explode;
 use function Pressbooks\Utility\str_ends_with;
 use function Pressbooks\Utility\str_lreplace;
+use function Pressbooks\Utility\str_remove_prefix;
+use function Pressbooks\Utility\str_starts_with;
+
 use Masterminds\HTML5;
 use Pressbooks\Admin\Network\SharingAndPrivacyOptions;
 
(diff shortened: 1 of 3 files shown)


inc/class-cloner.php
```diff
@@ -1223,6 +1226,31 @@ protected function sameAsSource( $url ) {
...
+			// Add book path if source book is in a subdomain install and target book is in a subdirectory install.
+			$content = preg_replace( '/href\="\/(front\-matter|chapter|back\-matter|part)/', 'href="/' . $subdirectory . '/$1', $content );
+		}
+
+		// Fix absolute URLs
+		$content = str_replace( 'href="' . untrailingslashit( $this->sourceBookUrl ), 'href="' . untrailingslashit( $this->targetBookUrl ), $content );
+
+		return $content;
+	}
+
```
connerbw: Per our discussion, I feel this would be better using the HTML5 parser looking for a href tags.

For example, If my books is on pressbooks.test/book1  how well does this handle me linking having both and internal link and a link to http://neuroscience.openetext.utoronto.ca/chapter/anatomy-physiology-the-cellular-levels-of-organization/ in the same chapter?

I feel this regex will touch edge cases it shouldn't.

PR #1320, opened 2018-08-10
Glossary

bdolor:
This PR creates default glossary functionality with the following:

- Integration with editing dashboard
- API endpoints
- Taxonomy and custom post type
- shortcode for Individual glossary terms with tooltip 
- click to see definition
- shortcode for Glossary list 
- compliance with [HTMLBook specification] for glossary list (https://oreillymedia.github.io/HTMLBook/#glossary)
- support for import and export(xml) 

**Admin dashboard**:
image

**API endpoints**
image

**Taxonomy**
image

**Shortcode for individual terms**
image

**Click to see definition**
image

**Shortcode for glossary list** 
image

**Support for import** 
image


connerbw:
Myself and @greatislander will both be on vacation for the next week so this may linger here until the week of the 20th. Thanks for your understanding!

inc/shortcodes/glossary/class-glossary.php
```diff
@@ -0,0 +1,268 @@
...
+	/**
+	 * Function to init our class, set filters & hooks, set a singleton instance
+	 *
+	 * @return Glossary
+	 */
+	static public function init() {
+		if ( is_null( self::$instance ) ) {
+			self::$instance = new self();
+			self::hooks( self::$instance );
+			self::setGlossaryTerms();
```
connerbw: I don't think `setGlossaryTerms` should in init(),

If a book has 10,000 glossary items this implies that 10,000 posts will be loaded on every page refresh. Even when not needed. This is going to cause performance issues.

Can you explore changing:

 + `self::setGlossaryTerms()` to `self::getGlossaryTerms()`
 + `$foo = self::$glossary_terms` to $foo =self::getGlossaryTerms()`
 + Caching when invoking getGlossaryTerms ( [`// cheap cache`  is probably good enough here?](https://github.com/pressbooks/pressbooks/blob/15f946b1a976bf0082d50bbcc33047f2a9be0679/compatibility.php#L21) I would like to see what happens with a glossary of 10,000 items but I'm sure we'll see it in Production and New Relic sooner than later )

I also don't think these needs to be static. All calls to  self::$glossary_terms that I see can be done in `$this` scope? If it doesn't work in `'glossary_terms'     => wp_json_encode( self::$glossary_terms )` because of the closure then `Glossary::init()->getGlossaryTerms()` would work fine.

SteelWagstaff:
Wow--really impressive work Brad and Alex. Looking forward to testing and using this with our authors soon!

inc/shortcodes/glossary/class-glossary.php
```diff
@@ -0,0 +1,287 @@
...
+		// Cheap cache
+		static $terms = null;
+		if ( null !== $terms ) {
+			return $terms;
+		} else {
+			$terms = $this->glossary_terms;
+		}
+
+		return $terms;
+	}
```
connerbw: See comments here:

https://github.com/pressbooks/pressbooks/pull/1321#pullrequestreview-148512332

Thoughts?

connerbw: CC: @alex-418

jcguan1:
Hi Brad, 
I noticed that the Glossary Terms and Glossary Types are located under the Organize menu, which is starting to be quite big. 

I suggest putting Glossary terms as a main menu item to make it more easily reachable.

bdolor:
🍦 🎉 🍰 👍

PR #1321, opened 2018-08-13
Attribution

bdolor:
addresses similar issue surfaced in https://github.com/pressbooks/pressbooks/pull/1320#discussion_r209426955

### DIFF
--- a/inc/shortcodes/attributions/class-attachments.php
+++ b/inc/shortcodes/attributions/class-attachments.php
@@ -19,7 +19,7 @@ class Attachments {
 	/**
 	 * @var array
 	 */
-	static $book_media = [];
+	private $book_media = [];
 
 	/**
 	 * Function to init our class, set filters & hooks, set a singleton instance
@@ -32,7 +32,6 @@ class Attachments {
 		if ( is_null( self::$instance ) ) {
 			self::$instance = new self();
 			self::hooks( self::$instance );
-			self::setBookMedia();
 		}
 
 		return self::$instance;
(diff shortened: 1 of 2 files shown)


inc/shortcodes/attributions/class-attachments.php
```diff
@@ -83,14 +82,33 @@ function ( $excluded_shortcodes ) {
...
+		// Cheap cache
+		static $media = null;
+		if ( null !== $media ) {
+			return $media;
+		} else {
+			$media = $this->book_media;
+		}
+
+		return $media;
+	}
```
connerbw: Can you clarify this approach?  I feel it would be confusing in this context:

```php
$one = new Attachments();
$one->setBookMedia();
$results = $one->getBookMedia();

// 1000 lines of code later

$two = new Attachments();
$results = $two->getBookMedia();

// $results will contain whatever $one set/get
```

The kind of caching I was thinking of was more like:

```php
$a = new Attachments();
$r1 = $a->getBookMedia(); // This queries the database
$r2 = $a->getBookMedia(); // Cached in memory, no database,
$r3 = $a->getBookMedia(); // Cached in memory, no database,
// ...
```

However, since we now have both  `setBookMedia()` and `getBookMedia()`, and set does all the work, then I feel the method could simply be:

```
function getBookMedia() {
   return $this->book_media;
}
```

The developer would be repsonsible for (re)setting as needed, using the same instance of the class as needed, etc.

Thoughts?

connerbw: CC: @alex-418

greatislander:
Hi @bdolor! Now that you are back from your well-deserved vacation, do you have some bandwidth to address @connerbw's questions above? I'd like to get an updated version of this PR merged for our forthcoming release of Pressbooks 5.5.0.

bdolor:
Hello @greatislander - I will endeavour to address this. When is the expected release date for 5.5.0?

greatislander:
We are aiming for the 12th of September; if that's too short a window we can implement this PR in a patch release after the main release.

bdolor:
I can commit to that deadline.

greatislander:
Many thanks!

greatislander:
Merging, will tweak after.

Issue #1218, opened 2018-05-30
SVG's in ePub 2.0


lukaiser:
Hi guys,

I have the following case:
In some books, users are using SVG's, well knowing that not all ePub-Reader support them.
If you export an ePub, the img tag gets included, but the image file is not added to the svg bundle.
When viewing the ePub in a reader, that supports SVG, they show up but only if the reader has a internet connection.

That is not really understandable for a user.
I see two options:

1. Add the SVG's to the bundle.
2. Remove the image tags from the ePub html and alert the user.

I would go with option one as more and more readers do support svg's.

Whats your take on that?
I would implement it if you are okay with it.

connerbw:
Epub2 files are used to build mobi files. Many Pressbooks users care more about the mobi files than they do the epub2 files. If you implement it then the resulting mobi files will need to work.

lukaiser:
I am still working on this. It is just moving slower then I thought.

PR #1322, opened 2018-08-13
Image filters

lukaiser:
Hey guys

Instead of adding the SVG support right away, as I was asking in #1218, I added 4 filters instead:

- pb_is_valid_image_extension
- pb_is_valid_image_type
- pb_epub201_fetchAndSaveUniqueImage_filename
- pb_epub201_fetchAndSaveUniqueImage_compress

They allow a plugin developer to add other images to the epub export or use other image generators then wordpress.com. 

An example for a SVG plugin:
https://github.com/eSkript/pressbooks-epub-with-svg

I hope that is a solution that makes everybody happy.

Lukas


inc/modules/export/epub/class-epub201.php
```diff
@@ -2034,12 +2050,22 @@ protected function fetchAndSaveUniqueImage( $url, $fullpath ) {
 		}
 
 		if ( $this->compressImages ) {
-			$format = explode( '.', $filename );
-			$format = strtolower( end( $format ) ); // Extension
-			try {
-				\Pressbooks\Image\resize_down( $format, $tmp_file );
-			} catch ( \Exception $e ) {
-				return '';
+			/**
```
connerbw: Can you move this filter into the contsructor? Set `$this->compressImages = false` next to this code:

https://github.com/pressbooks/pressbooks/blob/dev/inc/modules/export/epub/class-epub201.php#L403

So that it will be easier to spot what's going on if ever there's an issue.

lukaiser: Hmmm, I see your point, but then it isn't the same filter anymore. How I implemented the filter, you can turn of compression on a per image base and that is needed here, as I don't want to turn of compression for all the images.

greatislander:
Thanks @lukaiser! After some further discussion, we're fine with this as is.

PR #1326, opened 2018-08-22
Fix 500 Error

connerbw:
Uncaught exception 'Error' with message 'Call to a member function getParsedFiles() on null'

See: https://github.com/pressbooks/pressbooks/commit/967691052a14d5b90be17e82d05734acc2c3f636


connerbw:
This will fix the crash but doesn't really explain why so many exceptions are being thrown?

PR #1361, opened 2018-09-19
Transition from Dillard + Dillard Plain 1.x to Dillard 2.0

greatislander:
This should handle both locked and unlocked themes.

### DIFF
--- a/inc/theme/namespace.php
+++ b/inc/theme/namespace.php
@@ -167,6 +167,27 @@ function migrate_book_themes() {
 		$pressbooks_theme_migration = 3;
 		update_option( 'pressbooks_theme_migration', $pressbooks_theme_migration );
 	}
+
+	// Substitute Dillard Classic for Dillard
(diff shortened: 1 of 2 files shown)


inc/theme/namespace.php
```diff
@@ -167,6 +167,33 @@ function migrate_book_themes() {
...
+			$options = get_option( 'pressbooks_theme_options_global' );
+			$options['enable_title_decoration'] = 1;
+			update_option( 'pressbooks_theme_options_global', $options );
+		}
+
+		if ( $theme === 'pressbooks-dillardplain' ) {
+			// Migrate to Dillard 2.0
+			switch_theme( 'pressbooks-dillard' );
+			$lock = Lock::init();
+			if ( $lock->isLocked() ) {
```
connerbw: Can I get a code comment here? I'm not sure what `Migrate to Dillard 2.0` followed by some lock picking is doing.

PR #1363, opened 2018-09-20
'edition' field for 'general information' metabox

danzhik:
Edition of books is important information for any book, especially if we're talking about web-books, which content is easely and contineously changing. This field will provide consistenty between web-books and exported materials, eg in any moment of time one can see if printed edition or exported file of book is up-to-date/outdated for several purpose, for example adapting modified open books with latest changes of original source. For sure this commit doesn't provide functionality for features described above, but at least it will give people opportunity to save this information for further use.

### DIFF
--- a/inc/admin/metaboxes/namespace.php
+++ b/inc/admin/metaboxes/namespace.php
@@ -299,7 +299,7 @@ function add_meta_boxes() {
 			'description' => __( 'This is added to the metadata in your ebook.', 'pressbooks' ),
 		]
 	);
-	
+
 	x_add_metadata_field(
 		'pb_edition', 'metadata', [
 			'group' => 'general-book-information',
--- a/inc/admin/metaboxes/namespace.php
+++ b/inc/admin/metaboxes/namespace.php
@@ -299,6 +299,13 @@ function add_meta_boxes() {
 			'description' => __( 'This is added to the metadata in your ebook.', 'pressbooks' ),
 		]
 	);
+	
+	x_add_metadata_field(
+		'pb_edition', 'metadata', [
+			'group' => 'general-book-information',
+			'label' => __( 'Edition', 'pressbooks' ),
+		]
+	);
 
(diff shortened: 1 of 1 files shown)


greatislander:
Hi @danzhik — can you fix the coding standards error and then we will evaluate this?

greatislander:
A couple initial notes:

- [ ] This data would need to be added to the Book Info output in pressbooks/pressbooks-book
- [ ] This data would need to be added to Ebook and PDF exports (location and format TBD)
- [ ] This data would need to be added to the metadata endpoint of the Pressbooks REST API
- [ ] How would cloning handle this data? If someone clones a book and adapts it but neglects to update the edition, what happens?

connerbw:
Neat idea but this code is incomplete.

Adding the field doesn't add them to book outputs.

We try to make our metadata  https://schema.org/Book compatible so there is work to do in the API to make it look like https://schema.org/bookEdition

Etc.

If you need this ASAP then your code snippet would be better as a separate plugin on your own system. Ie.

```
add_action(
	'custom_metadata_manager_init_metadata', function () {
		x_add_metadata_field(
			'your_plugin_prefix_edition', 'metadata', [
				'group' => 'general-book-information',
				'label' => __( 'Edition', 'your-plugin-name' ),
			]
		);
	}
);
```
[For me, the code needs to do much more.](https://github.com/pressbooks/pressbooks/pull/1363#issuecomment-423161595) For me, this is not enough to merge into Pressbooks. It needs more thought. Get feedback in https://github.com/pressbooks/ideas or https://discourse.pressbooks.org/ first.

Regards,

greatislander:
@danzhik Thinking about this more, I'd rather discuss the implementation in a ticket, not in PR; this would need a ton of work to consider merging and I think it's best to figure out a comprehensive approach elsewhere rather than leaving an open PR for however long it takes to decide on implementation. Closing in lieue of https://github.com/pressbooks/ideas/issues/133. Let's continue the discussion there.

PR #1420, opened 2018-10-24
Glossary term widget with tabs (fix #1375)

### LINKED ISSUE #1375: TinyMCE button for highlighting a glossary term should use REST API to allow dynamic term loading and creation
### Feature Description

Currently, if a user highlights a word or phrase and clicks the button to add a glossary term, if the term is not already created it cannot be assigned to the word or phrase. Furthermore, if the term is created in a separate window, the editor must be refreshed to access the newly-created term. This experience could be improved by allowing the terms to be created dynamically from the modal using the REST API, and by loading/updating terms via the REST API when the modal is launched.  

### Feature Use Case

This would create a more frictionless user experience for managing glossary terms.

### Other Notes

N/A.

### System Information

Pressbooks: 5.6.0-dev

connerbw:


connerbw:
Depends on: https://github.com/pressbooks/pressbooks-book/pull/291

PR #1422, opened 2018-10-25
Add the_export_content filter hook and sanitize_webbook_content function, remove table border property

greatislander:
The `border` attribute for tables is deprecated, and conflicts with CSS styles for table borders. This PR:

- [x] Adds a new sanitization function, `sanitize_webbook_content`, which removes the deprecated table `border` attribute using HtmLawed by filtering `the_content`
- [x] Removes the deprecated table `border` attribute during EPUB, HTMLBook and XHTML export routines using HtmLawed
- [x] Adds a new filter hook, `the_export_content`, which unhooks `sanitize_webbook_content` before applying `the_content` filters to prevent processing post content twice during exports 
- [x] Sets the default table `border` to `0` so that TinyMCE will display tables properly in editing mode
- [x] Hides the table `border` field from the TinyMCE table properties modal


greatislander:
@connerbw In my testing this isn't a problem, because I'm only removing that filter within a function that processes shortcode content. I have a test chapter with this content (add via the text editor):

https://gist.github.com/greatislander/1afdbcc0bda88085cea1f896c49584a9

The table has the `border` attribute removed (as expected). So removing the filter in the shortcode handler doesn't appear to affect the `apply_filters( 'the_content' )` call in the post display.

When I left that filter in place, it broke my tests for those shortcodes, which is why I removed the filter. If this doesn't make sense to you I'm happy to look into other options in the AM.

PR #1426, opened 2018-10-29
Simplified glossary UX (fix #1424)

### LINKED ISSUE #1424: Simplify and improve editor experience for glossary terms
### Feature Description

A few changes are needed for the initial release of the glossary tool:

- [x] The glossary terms editor should not support rich text, shortcodes, etc. To communicate this, the visual editor should be disabled and a line of text should be added [above the editor](https://developer.wordpress.org/reference/hooks/edit_form_after_title/) to clarify this, reading: "HTML and shortcodes are not supported in glossary terms."
- [x] The glossary terms editor should have a simplified post status metabox and a visibility toggle (which simply switches post status between `publish` and `private`). It's important to verify that tooltips will still appear for `private` glossary terms, and that `private` glossary terms will not appear in the shortcode-generated list.
- [x] The 'View Glossary Terms' link should be removed from the glossary terms table admin bar, and the 'View Glossary Term' link should be removed from the glossary terms editor.
- [x] The glossary terms table should have the Quick Edit and View action links [removed](https://frique.me/blog/how-to-remove-wordpress-quick-edit/).
- [x] The glossary terms table should have the date filter removed.

### Feature Use Case

Not applicable.

### Mockups

Status & Visibility

Glossary Terms

View Link

### Other Notes

Not applicable.

### System Information

Pressbooks: 5.6.0-dev

connerbw:


greatislander:
Is it possible to change the status label here? If so, we should probably change it to something like "Unlisted".

screen shot 2018-10-29 at 5 32 27 pm

connerbw:
> particularly because I can add <script> tags

Admins can do anything. I tried as an Editor and <script> was removed.

Incoming commit removes all tags though, harder than I thought it would be.

PR #1429, opened 2018-10-30
Add DOI support at book and section level, exposed via metadata API

greatislander:
See implementation notes [here](https://discourse.pressbooks.org/t/support-for-digital-object-identifier-doi-in-book-info-metadata/684/4). Also includes a new filter hook, `pb_doi_resolver`, which allows the default DOI resolver service URL (https://dx.doi.org) to be changed.


greatislander:
@connerbw Correct, the DOI in tests is not a real DOI; for the first release of this feature we are going to assume people who know what a DOI is know what to put in. In future we may add some kind of admin-side validation (check to see if it's a valid DOI, warn and don't save if not.

PR #1442, opened 2018-11-05
Display glossary in back matter (fixes #1433, fixes #1434)

### LINKED ISSUE #1433: Remove Glossary List shortcode button from TinyMCE
For now, we won't be surfacing this functionality. See follow-up issue #1434.

### LINKED ISSUE #1434: Automatically display shortcode list in Glossary back matter if content is empty
### Feature Description

If back matter with the Glossary back matter type exists AND the content is empty (whitespace should be ignored if that's all that's there), the rendered output of `[pb_glossary]` should be displayed instead of the content. A notice should be displayed above the editor which reads:

`To display a list of glossary terms, leave this back matter's content blank.`

This notice should be inserted via JavaScript when the Glossary back matter type is selected and removed when it is deselected. Loading existing back matter for editing with the Glossary back matter type should display the message  via the `edit_form_after_title` action hook.

### Feature Use Case

Until we refine the glossary list shortcode further, this will be an easy method for users to output their glossary content.

### Other Notes

Not applicable.

### System Information

Pressbooks: 5.6.0-dev

connerbw:
Note; Put JS scripts (back) into alphabetical order and webpack made a big diff. Sorry.


inc/posttype/namespace.php
```diff
@@ -233,6 +233,16 @@ function after_title( $post ) {
...
 		echo '</p>';
 	}
+	if ( $post->post_type === 'back-matter' ) {
+		$taxonomy = \Pressbooks\Taxonomy::init();
+		if ( $taxonomy->getBackMatterType( $post->ID ) === 'glossary' ) {
+			echo '<div id="pb-post-type-notice" class="notice notice-info"><p>';
+			_e( "To display a list of glossary terms, leave this back matter's content blank.", 'pressbooks' );
+			echo '</p></div>';
+		} else {
+			echo '<div id="pb-post-type-notice" class="notice notice-info" style="display:none;"></div>'; // Placeholder
```
greatislander:
```suggestion
			echo '<div id="pb-post-type-notice" class="notice notice-info" role="alert" style="display:none;"></div>'; // Placeholder
```

This should be an [ARIA live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) so that it will be announced to screen reader users when they switch the back matter type to Glossary.

inc/posttype/namespace.php
```diff
@@ -233,6 +233,16 @@ function after_title( $post ) {
 		_e( 'HTML and shortcodes are not supported in glossary terms.', 'pressbooks' );
 		echo '</p>';
 	}
+	if ( $post->post_type === 'back-matter' ) {
+		$taxonomy = \Pressbooks\Taxonomy::init();
+		if ( $taxonomy->getBackMatterType( $post->ID ) === 'glossary' ) {
+			echo '<div id="pb-post-type-notice" class="notice notice-info"><p>';
```
greatislander:
```suggestion
			echo '<div id="pb-post-type-notice" class="notice notice-info" role="alert"><p>';
```

This should be an [ARIA live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) so that it will be announced to screen reader users when they switch the back matter type to Glossary.

PR #1453, opened 2018-11-07
 Add getAllSubsections method to \Book (fix #1438)

### LINKED ISSUE #1438: Move subsection caching code from pressbooks/pressbooks-book#292 into core

greatislander:
NB: builds will fail until #1452 is merged. Corresponding PR: https://github.com/pressbooks/pressbooks-book/pull/358.


inc/class-book.php
```diff
@@ -490,6 +494,9 @@ static function deleteBookObjectCache() {
 		static::$preview = [];
 		static::$__order = [];
 
+		// Subsections
+		delete_transient( \Pressbooks\Book::SUBSECTIONS_TRANSIENT );
```
connerbw: I prefer `delete_transient( static::SUBSECTIONS_TRANSIENT );` because it's referencing itself here.

inc/class-book.php
```diff
@@ -544,6 +551,52 @@ static function getSubsections( $id ) {
...
 	}
 
+	/**
+	 * Returns an array of front matter, chapters, and back matter which contain subsections.
+	 *
+	 * @param array $book_structure The book structure from getBookStructure()
+	 * @return array The subsections, grouped by parent post type
+	 */
+	static function getAllSubsections( $book_structure ) {
+		if ( Export::isParsingSubsections() ) {
```
connerbw: This is the wrong time to notice this since this dates back to 2015 and the change was made to match coding standards (da87d42) but with several new changes related to cache stampedes, this feels like a weird function name?

inc/class-book.php
greatislander:
```suggestion
		delete_transient( static::SUBSECTIONS_TRANSIENT );
```

inc/class-book.php
greatislander: Agreed…

PR #1460, opened 2018-11-13
Fix: Numberless chapters in the beginning of the main body cause chapter number error in subsequent chapters

connerbw:


connerbw:
image

The reason for the numberless bug was that we were checking if a $subclass === 'numberless' but it could include other classes ('numberless introduction' in the screenshot, for example)

PR #1467, opened 2018-11-16
lower version numbers for rhel compatible cover generator dependencies

bdolor:
As discussed in https://discourse.pressbooks.org/t/cover-generator-commands-versions/787 this brings server dependency versions to a level consistent with what is available to RHEL. 

Testing confirms no loss of functionality on 16.04.5 LTS (Xenial Xerus) and enables functionality on RHEL 6.10 (Santiago)


greatislander:
Many thanks, @bdolor!

PR #1474, opened 2018-11-22
png to jpeg as failsafe routine

bdolor:
This PR does two things. It allows the Exception to be an exception under all exceptional circumstances...rather, it removes the potential for a `PHP Fatal` if `$pid` is not able to return error codes (if the file is empty) 

image

The certain circumstances alluded to above in which the file will return empty is captured here: https://stackoverflow.com/questions/18719256/why-does-pdftoppm-poppler-utils-have-no-jpeg-option 

Under the `pdftoppm` has no `-jpeg` argument, the error code from the first command will be something other than `0` and no file will be created. 

screen shot 2018-11-21 at 4 39 45 pm

Under those two conditions the proposed PR will convert to png then use `imagemagick` to convert png to jpg.


greatislander:
Thanks @bdolor -- this won't make it into Wednesday's release but we will aim to include it in the next patch (5.6.1).

greatislander:
@bdolor We're hoping to get 5.6.3 out next week so I'm going to merge this as is and we can follow up later with potential improvements using https://github.com/intervention/image.

PR #1523, opened 2018-12-18
Capability type

connerbw:
page: part, metadata
post: chapter, front-matter, back-matter, glossary


connerbw:
Not sure what is going with the diff. Doesn't seem to be comparing with current version of dev, rather previous version of dev.

greatislander:
@connerbw Looks like you branched from your last PR? I'm also confused.

PR #1524, opened 2018-12-19
regex and case insensitive for epubcheck version

bdolor:
EpubCheck has changed their brand name from `EpubCheck` to `EPUBCheck`. 
ref: https://github.com/w3c/epubcheck/commit/dfd7fd272068cbca4c0c3929742e514c7de8affb

Case sensitive checking for dependencies fails with an update to EPUB v4.1.0. This PR moves to case insensitive `stripos` and `preg_match` for forward and backward compatibility.

https://github.com/w3c/epubcheck/releases/tag/v4.1.0

### DIFF
--- a/inc/utility/namespace.php
+++ b/inc/utility/namespace.php
@@ -298,9 +298,9 @@ function check_epubcheck_install() {
 	exec( PB_EPUBCHECK_COMMAND . ' -h 2>&1', $output, $return_val );
 
 	$output = $output[0];
-	if ( false !== strpos( $output, 'EpubCheck' ) ) { // Command found.
-		$output = explode( 'EpubCheck v', $output );
-		$version = $output[1];
+	if ( false !== stripos( $output, 'EPUBCheck' ) ) { // Command found.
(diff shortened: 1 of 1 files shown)


greatislander:
@bdolor In further testing, I think when EPUBCheck is absent this may not work.

If I run:

```
/usr/bin/java -jar /opt/epubcheck-is-not-here/epubcheck.jar
```

The output (Ubuntu Bionic) will be:

```
Error: Unable to access jarfile /opt/epubcheck-is-not-here/epubcheck.jar
```

The problem then is that `stripos( 'Error: Unable to access jarfile /opt/epubcheck-is-not-here/epubcheck.jar', 'EPUBCheck' )` will return `true`, and `$matches[1]` will be undefined because a matching version string will not be found. The function will still return false as expected but we get undefined offset errors as seen here: https://travis-ci.org/pressbooks/pressbooks-lti-provider/jobs/474428782

bdolor:
nice catch @greatislander

PR #1542, opened 2019-01-07
Handle missing legacy images (fix pressbooks/pressbooks-donham#3)

greatislander:
Recent changes to `normalizeCssUrls()` method appears to have broken images referenced from CSS in legacy themes. This is very uncommon but it affects the current 1.x version of Donham. This commit fixes the issue.

### DIFF
--- a/inc/modules/export/epub/class-epub201.php
+++ b/inc/modules/export/epub/class-epub201.php
@@ -704,9 +704,9 @@ class Epub201 extends Export {
 	/**
 	 * Search for all possible permutations of CSS url syntax -- url("*"), url('*'), and url(*) -- and update URLs as needed.
 	 *
-	 * @param string $css
-	 * @param string $scss_dir
-	 * @param string $path_to_epub_assets
+	 * @param string $css The EPUB's (S)CSS content.
+	 * @param string $scss_dir The directory which contains the theme's SCSS files. No trailing slash.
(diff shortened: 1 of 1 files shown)


greatislander:
Branched from dev so this will need to be cherry-picked into hotfix.

inc/modules/export/epub/class-epub201.php
```diff
@@ -734,6 +734,13 @@ protected function normalizeCssUrls( $css, $scss_dir, $path_to_epub_assets ) {
 					if ( $my_image ) {
 						copy( $my_image, "$path_to_epub_assets/$filename" );
 						return "url(assets/$filename)";
+					} else {
+						// This is not a Buckram theme, maybe.
+						$my_legacy_image = realpath( $scss_dir . '/' . $url );
```
connerbw: To be consistent with the code a few lines down, use` "$scss_dir/$url"` instead.

Also weird is sometimes we do `$my_image = realpath( "$scss_dir/$url" );` other times we do `$my_image = realpath( $buckram_dir . $url );`  (One has a slash, the other doesn't. I just want to flag it as something that bugs me. I am not saying refactor.

greatislander: It annoyed me too. I've made the method more consistent.

PR #1553, opened 2019-01-15
Rebuild Export Page (fixes #1525)

### LINKED ISSUE #1525: Rebuild export page with bulk actions, pinned exports, improved accessibility
### Mockup

Mockup of new export page.

### Implementation Notes

+ This can be developed separately from the WIP export progress bar branch; I will merge the two later on.
+ The table must should be sortable by filename, date, file type, and pinned status.
+ Selecting multiple files should allow bulk deletion.
+ Instead of keeping the last five of any type of export, we now keep the last three of each specific export format.
+ The table view should display up to 50 rows by default. The number of visible rows should be manageable with the standard WordPress interface (Screen Options).
+ Checking the "Pin" box prevents the related export file from being deleted. Up to five files can be pinned at once.
+ If the user has pinned three files of a given export type and then tries to export that format, the export job should be stopped and an error should be displayed instructing them to deselect one of the pinned files before attempting to export. 
+ If the user has pinned three files of a given export type and they then try to pin an additional file of that type, an error should be displayed instructing them to deselect one of the pinned files before attempting to pin another.

connerbw:


inc/modules/export/class-table.php
```diff
@@ -0,0 +1,451 @@
...
+	/**
+	 * @return array
+	 */
+	public function get_columns() {
+		return [
+			'cb' => '<input type="checkbox" />',
+			'file' => __( 'File', 'pressbooks' ),
+			'format' => __( 'Format', 'pressbooks' ),
+			'size' => __( 'Size', 'pressbooks' ),
+			'pin' => '📌', // <span class="dashicons dashicons-admin-post"></span> or 📌
```
greatislander:
```suggestion
			'pin' => __( 'Pin', 'pressbooks' ),
```

inc/modules/export/class-table.php
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	/**
+	 * @return array
+	 */
+	public function get_sortable_columns() {
+		$sortable_columns = [
+			'file' => [ 'file', true ],
+			'format' => [ 'format', true ],
+			'pin' => [ 'pin', true ],
```
greatislander: Should also be sortable by size.

inc/modules/export/class-table.php
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	/**
+	 * @param string $file
+	 *
+	 * @return string
+	 */
+	protected function getFormat( $file ) {
+		$file_class = $this->getCssClass( $file );
+		return ucfirst( $file_class );
```
greatislander: We should use something like this instead (moving the functionality into core would mean we could also remove it from McLuhan): https://github.com/pressbooks/pressbooks-book/blob/dev/inc/helpers/namespace.php#L76-L107

inc/modules/export/namespace.php
```diff
@@ -0,0 +1,201 @@
...
+	 *    $formats['exotic']['docx'] = __( 'Word (Beta)', 'pressbooks' );
+	 *    return $formats;
+	 * } );
+	 *
+	 */
+	$formats = apply_filters( 'pb_export_formats', [
+		'standard' => [
+			'print_pdf' => __( 'PDF (for print)', 'pressbooks' ),
+			'pdf' => __( 'PDF (for digital distribution)', 'pressbooks' ),
+			'epub' => __( 'EPUB (for Nook, iBooks, Kobo etc.)', 'pressbooks' ),
```
greatislander:
```suggestion
			'epub' => __( 'EPUB (for Nook, Apple Books, Kobo etc.)', 'pressbooks' ),
```

inc/modules/export/class-table.php
```diff
@@ -0,0 +1,451 @@
...
+
+	/**
+	 * @param array $item A singular item (one full row's worth of data)
+	 *
+	 * @return string Text to be placed inside the column <td>
+	 */
+	public function column_pin( $item ) {
+		$format = $this->getTinyHash( $item['format'] );
+		$html = "<input type='checkbox' name='pin[{$item['ID']}]' value='{$format}' " . checked( $item['pin'], true, false ) . '/>';
+		return $html;
```
greatislander: Each of these needs a text label for accessibility. This means they need an ID attribute and a `<label for="<ID>">Pin this file</label>`. To visually hide the label in mobile view (when the column header is shown next to the checkbox), you can use this SCSS mixin:

https://github.com/pressbooks/pressbooks/blob/dev/assets/src/styles/organize.scss#L89-L96

inc/modules/export/class-table.php
```diff
@@ -0,0 +1,451 @@
...
+		// Define Columns
+		$columns = $this->get_columns();
+		$hidden = [];
+		$sortable = $this->get_sortable_columns();
+		$this->_column_headers = [ $columns, $hidden, $sortable ];
+
+		// Data
+		$data = $this->getLatestExports();
+
+		// Pagination
```
greatislander: Not sure this will be an issue but best to check out: https://make.wordpress.org/core/2019/01/14/new-styling-for-admin-table-pagination-links-in-wordpress-5-1/

assets/src/scripts/export.js
```diff
@@ -50,22 +85,88 @@ jQuery( function ( $ ) {
...
+				// Unchecked
+				delete _pb_export_pins_inventory[ name ];
+				cb.prop( 'disabled', false );
+				tr.find( 'td.column-file span.delete' ).show();
+			}
+			$.ajax( {
+				url: ajaxurl,
+				type: 'POST',
+				data: {
+					action: 'pb_update_pins',
```
greatislander: We should display a success notice (which fades out) for successful pinning. It should also function as an aria-live region to announce the successful event. Something like:

```php
<div class="notice notice-success fade">
<?php
  printf(
    __( 'The file %s has been pinned successfully.', 'pressbooks' ),
    $filename
  );
?>
</div>
```

connerbw: Code is in place but I am having a hard time getting `notice notice-success fade` to not feel awkward. It makes the interface jump around while pinning. Maybe this could be screen-reader only? Please tweak. Currently has no classes...

greatislander: I am changing it to `.screen-reader-only`.

Issue #590, opened 2016-12-13
Language packs


greatislander:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

It would be nice if we could update il8n files without having to release a new version of the plugin. This might let us do that: https://github.com/afragen/github-updater-language-pack-maker

### Steps to Reproduce

Not applicable.

### System Information

Pressbooks: 3.9.6
WordPress: 4.7
PHP: 7.1

greatislander:
This will make @zwhpressbooks’ life much easier! 🇳🇿

greatislander:
Never mind.

afragen:
Just an FYI. I created a composer library to make it easier to add to other projects.

https://github.com/afragen/translations-updater

It requires the translations be kept in a separate public repository.

If this is still on your radar let me know, I’m happy to help implement it with you.

greatislander:
@afragen That's very exciting news. We may follow up with you about this once we enter the 5.8.0 release cycle for Pressbooks. I appreciate all your contributions to the community in building these tools!

connerbw:
Link to docs: https://thefragens.com/translations-updater/

PR #1575, opened 2019-01-30
Add support for decoupled language pack updates

afragen:
Fixes https://github.com/pressbooks/pressbooks/issues/590

Still need to create a public repository to contain translations. Also need to update the $config array to use that URL.

What I do for the translations repository is use it for people to make PRs only to MO/PO files. After committing the PRs I pull the updates locally where I run the Language Pack Maker. This creates updated language pack zip files and the `language-pack.json` that is read by the `translations-updater` library.

It may take a little hand holding from me the first time, but after that it's pretty simple. Then you can remove the MO/PO files from the main plugin.

### DIFF
--- a/inc/class-updates.php
+++ b/inc/class-updates.php
@@ -44,6 +44,7 @@ class Updates {
 			add_action( 'core_upgrade_preamble', [ $obj, 'coreUpgradePreamble' ] );
 		}
 		add_filter( 'extra_plugin_headers', [ $obj, 'extraPluginHeaders' ] );
+		add_action( 'admin_init', [ $obj, 'translationsUpdater' ] );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


greatislander:
@afragen One implementation question about this: can decoupled language packs be installed using `wp language plugin install pressbooks fr_FR`? https://developer.wordpress.org/cli/commands/language/plugin/

afragen:
@greatislander since the languages would no longer reside with the main plugin, I really don't know. If the command can point to an outside source I believe it should be possible.

If the site's locale is set appropriately then the appropriate language packs will be installed and stay updated.

You also have some other non-standard translation files that I don't really know how they function.

afragen:
Where does the wp-cli command obtain the language files from?

inc/class-updates.php
```diff
@@ -210,4 +211,22 @@ public function getPluginsWithPressbooksInDescription() {
...
+	 *
+	 * @return void
+	 */
+	public function translationsUpdater() {
+		$config = [
+			'git'       => 'github',
+			'type'      => 'plugin',
+			'slug'      => 'pressbooks',
+			'version'   => PB_PLUGIN_VERSION,
+			'languages' => 'https://my-path-to/language-packs',
```
connerbw: Proof-of-concept is nice, but I'm concerned that we're merging something not finished into dev. This URL clearly doesn't exist.

PS: How can we test this?

connerbw: I'll rephrase.

The code in dev should be working code. This can't work with this URL,  so until it does i'd prefer it stays in a branch.

I'd like a unit test that, at the very least, covers this code so that it's triggered and we can at least see if it crashes or not. A better unit test would prove that this is working somehow.

Thanks!

afragen: The URL needs to be added. I stated that above. A public repository that contains the translations will need to be created. This is the URL that goes here.

afragen:
https://thefragens.com/translations-updater/

connerbw:
Thanks for this. Closing for now. Will revisit and implement when ready.

PR #1576, opened 2019-01-30
H5P Cloning (fixes #1529)

### LINKED ISSUE #1529: Support cloning of H5P activities
### Feature Description

Cloning operations should include the cloning of [H5P](https://h5p.org/) activities.

### Feature Use Case

H5P activities are often an integral part of open textbooks created on Pressbooks and as such it should be included when cloning these books.

### Implementation Notes

- Depends on the production release of changes currently in the H5P WordPress plugin's master branch which address https://github.com/h5p/h5p-wordpress-plugin/issues/63
- H5P cloning will only work from networks running a supported version of the H5P plugin (which exposes REST API endpoints for H5P activities); if content is found but the H5P endpoints aren't available, we need to notify the user
- The H5P plugin needs to be active in the target book in order to create the H5P activities; if it is not installed we need to notify the user
- Initial activation of the H5P plugin involves showing a privacy/welcome notice et cetera, so this is an expectation we should set with users when they navigate to a freshly-cloned book

Maybe something like this?

1. Look for H5P endpoints on source book when setting up clone data, set `$this->sourceHasH5pApi` accordingly;
1. Check for substring `[h5p id=` to the cloner's processing of post content; the first time it detects that, set `$this->sourceHasH5p` to true;
1. If `$this->sourceHasH5p` === true but `$this->sourceHasH5pApi` === false, add a notice to the user indicating that the H5P could not be cloned and why.
1. If `$this->sourceHasH5p` === true and `$this->sourceHasH5pApi` === true, activate the plugin so content can be cloned properly.
1. If the plugin is not locally available, add a notice to the user indicating that the H5P could not be cloned and why.

### System Information

H5P: master
Pressbooks: 5.7.0-dev

connerbw:


connerbw:
In `\Pressbooks\Cloner\Cloner::clonePostProcess` i'm looking for a new string that explains: `$this->sourceHasH5p === true && ( $this->sourceHasH5pApi === false || $this->targetHasH5pApi === false )`:

`$_SESSION['pb_notices'][] = __( 'This book contains H5P content that could not be cloned. Please review the cloned version of your text carefully, as missing H5P content will be indicated. You may want to remove or replace these sections.', 'pressbooks' );`

Using `$_SESSION['pb_notices']` instead of the newer `\Pressbooks\add_notice` because the change to SSE is in a weird, half-way, backwards compatible state. Once this is merged we should go through the whole app and see how many of these we can change.

connerbw:
I've split the downloading code into a separate class.

Right now, cloning (and WXR import) works a little something like:

 + Setup
 + Pre Process
   + Example: Check if there's H5P content and if yes, activate the plugin on a new book
 + Clone stuff
   + Move HTML/Text content from source into destination
   + When a media element is encountered in the HTML/Text, download it.
   + If an ID is found, and it has changed from source into destination, flag it.
   + When shortcode is encountered, and the shortcode has an ID, flag it.
 + Post Process:
   + Fix shortcode IDs
 + Done
 
 
The idea of separating the downloading into a separate class is, hopefully, to move the "When a media element is encountered in the HTML/Text, download it." step outside of the "Clone stuff" step. Something like:

 + Clone stuff [...]
 + Download stuff
 + Post Process

This would improve the SSE progress bar. Right now, if a section has a lot of media in it, the progress bar moves very slowly from section to section. There's no sensible way to improve that progress bar feedback until the download step is moved out of those steps.

greatislander:
Suggested revision to copy:

> The source book contained H5P content that could not be cloned. Please review the cloned version of your book carefully, as missing H5P content will be indicated. You may want to remove or replace these elements.

inc/cloner/class-cloner.php
```diff
@@ -636,14 +694,19 @@ public function cloneGlossary( $id ) {
 	 */
 	public function clonePostProcess() {
 		$this->fixInternalShortcodes();
+		// H5P
+		if ( $this->sourceHasH5p === true && ( $this->sourceHasH5pApi === false || $this->targetHasH5pApi === false ) ) {
+			// Add a notice to the user indicating that the H5P could not be cloned
+			$_SESSION['pb_notices'][] = __( 'This book contains H5P content that could not be cloned. Please review the cloned version of your text carefully, as missing H5P content will be indicated. You may want to remove or replace these sections.', 'pressbooks' );
```
greatislander:
```suggestion
			$_SESSION['pb_notices'][] = __( 'The source book contained H5P content that could not be cloned. Please review the cloned version of your book carefully, as missing H5P content will be indicated. You may want to remove or replace these elements.', 'pressbooks' );
```

inc/cloner/class-cloner.php
```diff
@@ -680,119 +743,37 @@ public function buildListOfKnownMedia( $url ) {
...
-						Shortcodes\Attributions\Attachments::SHORTCODE,
-						'id',
-						$transition->oldId,
-						$transition->newId
-					);
-				}
-			}
-			wp_update_post( $post );
-		}
+	public function createTransition( $type, $old_id, $new_id ) {
```
greatislander: Nice!

inc/interactive/class-h5p.php
```diff
@@ -121,36 +167,43 @@ public function replaceShortcode( $atts ) {
...
 	 */
-	public function replaceCloneable( $content ) {
-		$this->setCloneableWarning();
+	public function replaceUncloneable( $content ) {
 		$pattern = get_shortcode_regex( [ self::SHORTCODE ] );
 		$content = preg_replace_callback(
 			"/$pattern/",
 			function ( $m ) {
-				return __( 'The original version of this chapter contained H5P content. This content is not supported in cloned books. You may want to remove or replace this section.', 'pressbooks' );
+				return __( 'The original version of this chapter contained H5P content. You may want to remove or replace this section.', 'pressbooks' );
```
greatislander:
```suggestion
				return __( 'The original version of this chapter contained H5P content. You may want to remove or replace this element.', 'pressbooks' );
```

PR #1600, opened 2019-02-14
Handle weird usernames in Network Managers table

greatislander:
Using `get_user_by( 'login', $username )` works with users with unexpected characters in their usernames (e.g. `@`, `.`) that may have been added by third-party plugins. I don't actually know why I used `get_user_by( 'slug', $username )` in the first place? (FYI @SteelWagstaff.)


SteelWagstaff:
<3 Thanks Ned!

PR #1607, opened 2019-02-21
Rename Austen theme (pressbooks-austentwo → pressbooks-austen) (fix #1601)

### LINKED ISSUE #1601: Handle Austen upgrade

greatislander:
We will be renaming Austen with the v3.0 upgrade (`"pressbooks/pressbooks-austentwo":"^3.0.0"` gives me a headache). This handles the upgrade, which will only take place on Pressbooks/PressbooksEDU infrastructure.

### DIFF
--- a/inc/theme/namespace.php
+++ b/inc/theme/namespace.php
@@ -194,6 +194,26 @@ function migrate_book_themes() {
 		$pressbooks_theme_migration = 4;
 		update_option( 'pressbooks_theme_migration', $pressbooks_theme_migration );
 	}
+
+	// Rename Austen Two to Austen
+	if ( $pressbooks_theme_migration === 4 ) {
+		$theme = wp_get_theme()->get_stylesheet();
+		if ( $theme === 'pressbooks-austentwo' ) {
+			// Switch theme to Austen 3.0
+			switch_theme( 'pressbooks-austen' );
+			$lock = Lock::init();
+			if ( $lock->isLocked() ) {
+				$data = $lock->getLockData();
+				$data['stylesheet'] = 'pressbooks-austen';
+				$json = wp_json_encode( $data );
+				$lockfile = $lock->getLockDir() . '/lock.json';
+				\Pressbooks\Utility\put_contents( $lockfile, $json );
+			}
+		}
+
+		$pressbooks_theme_migration = 5;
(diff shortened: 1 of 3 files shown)


inc/theme/namespace.php
```diff
@@ -194,6 +194,26 @@ function migrate_book_themes() {
...
+			// Switch theme to Austen 3.0
+			switch_theme( 'pressbooks-austen' );
+			$lock = Lock::init();
+			if ( $lock->isLocked() ) {
+				$data = $lock->getLockData();
+				$data['stylesheet'] = 'pressbooks-austen';
+				$json = wp_json_encode( $data );
+				$lockfile = $lock->getLockDir() . '/lock.json';
+				\Pressbooks\Utility\put_contents( $lockfile, $json );
+			}
```
connerbw: The `migrate_book_themes` function is starting to get a bit unwieldy. The 8 lines above are, more or less, copy/pasted 3 times now.

On the one hand I think this should be refactored. On the other hand, maybe it's fine because it will only be run once.

What do you think?

greatislander: Probably worth pulling that chunk into its own utility function 👍

greatislander:
Thanks @connerbw, will try to fix later :)

connerbw:
> You appear to have already installed WordPress. To reinstall please clear your old database tables first.

Something is broken and I don't know what. I reset everything I could and if I run the test suite twice I keep getting this error. The lite stack trace is:

+ _book;
+  $this->factory()->blog->create()
+ wpmu_create_blog

Feels like WordPress is releasing something today and I'm caught in the middle?

connerbw:
Maybe "Test Suite: New Test Config File Constant" ?
https://make.wordpress.org/core/2019/01/23/miscellaneous-developer-focused-changes-in-5-1/

greatislander:
@connerbw I set a hard version for Travis to run WordPress 5.0.3 (and reinstalled my local test environment using that version). Still getting the error? I'm confused. Maybe related? https://core.trac.wordpress.org/ticket/40364#comment:63

greatislander:
Ancient Trac ticket: https://core.trac.wordpress.org/ticket/27001

greatislander:
Got it working!

greatislander:
Fixes #1601.

PR #1614, opened 2019-02-26
Fix: Notice: Undefined index: pb_cover_image

connerbw:
See: https://pressbooks.community/t/undefined-index-pb-cover-image-error-shows-in-some-catalog-dashboards/890


rootl:
Thank you! I'd put in a thumbs up emoji if I could!

PR #1666, opened 2019-04-16
Fix table headers to display accessibility info (fix #1633)

### LINKED ISSUE #1633: Table headers on the organize page are not applying role information in semantically meaningful ways
### Description

Table headers on the Organize page currently have role=”button”: https://github.com/pressbooks/pressbooks/blob/66fda01dc395d98bddeecf9705fb520837530d09/templates/admin/organize.php#L129-L131 

These header cells are read as blank or skipped by the screen reader (depending on the browser), likely because the button role overrides the semantic information of the table header. Because <th> elements are not keyboard accessible by default, these buttons also do not currently receive keyboard focus.

We should consider adding the button role to an element, such as a span, nested within the table header, rather than the <th> element itself. We should also add a tabindex attribute to this internal element. For example: 
`<th><span role="button" tabindex="0">Show in Web</span></th>`

In addition, neither the purpose (check/uncheck all pages in column) nor the status (checked/unchecked) of these buttons is clear. We should add additional information through the use of an ARIA label to clarify what pressing the button does and the current state of the button (for screen reader users). 

For example, we should consider adding something like `aria-label="check/uncheck Show in Web for all Front Matter pages"` as well as an `aria-pressed` attribute. For example, if some buttons within the table are checked while others are not, the attribute should be `aria-pressed="undefined"`; if all boxes have been checked, it should be set to `aria-pressed="true"`, and if unchecked, `aria-pressed="false"`. Ideally the label attribute should change dynamically depending on the state of the buttons within the table.

Alternatively, we may with to place an actual checkbox in the header itself to allow users to explicitly select/deselect all items in that section column.

dannylonglegs:
Fix #1633


dannylonglegs:
This PR fixes how table headers display accessibility info and how they are highlighted by keyboard selection. It does not address adding the checkmark to the table header buttons to better display their function. This should be addressed in another issue in which the dev team can spec it first.

PR #1671, opened 2019-04-17
Convert a11y.js file to vanilla js, add attributes for accessibility

### LINKED ISSUE #1636: Sortable tables lack ARIA-sort properties
### Description

When tables allow users to sort the content in a column, the column header should receive an ARIA `columnheader` role and `col` scope property. This will let users of assistive technology know that they are able to sort the information in the table using the table headers. For an example of an accessible sortable  table, see: https://dequeuniversity.com/library/aria/tables/sf-sortable-grid

Two examples of sortable tables that lack these attributes are the table on the '[My Catalog](https://github.com/pressbooks/pressbooks/blob/a93df1e7297b8ad0a4d9339b8d735c2a1ebdc06b/inc/admin/class-catalog-list-table.php)' page and the table on the 'Contributors' page.

### LINKED ISSUE #1641: Tables used for presentation in theme options aren't marked as such
### Description

On the theme options menus and on our cloning routine page, we present several sets of options within tables. These tables are used for presentation, but this information is not clearly indicated as such for users of assistive technology. The simplest fix would be to add `role="none"` to the various `<table>` elements to allow AT to disregard the tabular structure. In addition, we might also wish to replace tables altogether, dropping row header `<th>` elements with `<label>` elements.

dannylonglegs:
fix #1636 , fix #1641 

These changes make sure that all tables with the class of form-table receive a `role=none` attribute and that table headers on `wp-list-table`s receive a `role=columnheader`


assets/src/scripts/a11y.js
```diff
@@ -1,5 +1,12 @@
-jQuery( function ( $ ) {
+document.addEventListener( 'DOMContentLoaded', function () {
 
-	// Add `role: columnheader` to Sortable Table headers
-	$( 'table.wp-list-table th' ).attr( 'role', 'columnheader' );
+	function addAttribute( x, y, z ){
```
connerbw: Change variable names `x, y, z` to something meaningful.

Suggestion: `selectors, att, val`

PR #1710, opened 2019-05-18
Croatian Language Support

dcvikl:
.tx/config : added hr:hr_HR mapping for Croatian.

Completely translated languages/core-hr-hr.php to Croatian Language.


connerbw:
Will fix myself.

PR #1713, opened 2019-05-22
Add aria-labels to color picker features for accessibility 

### LINKED ISSUE #1659: Color picker not fully accessible
### Description

The content of the color picker used in various locations in the Pressbooks interface (inherited from WordPress) is not communicated to assistive technologies. The selector within the gradient is communicated as “link”, and colors (black, white, red, etc.) are announced as group. Can be fixed by adding an ARIA label to each color so that it says something other than 'group' (i.e. the color's name).

dannylonglegs:
fix #1659


assets/src/scripts/a11y.js
```diff
@@ -26,4 +26,48 @@ document.addEventListener( 'DOMContentLoaded', function () {
 	addAttribute( 'div.updated', 'role', 'status' );
 	addAttribute( 'div.notice', 'role', 'status' );
 	addAttribute( 'div.error', 'role', 'alert' );
+
+	// Add aria-label attribute to color picker slider
+	// https://github.com/Automattic/Iris/issues/69
+	addAttribute( 'div.iris-slider.iris-strip', 'aria-label', 'Gradient selector' );
+	addAttribute( '.ui-slider-handle', 'aria-label', 'Gradient selector' );
```
connerbw: English strings should be made i18n compatible.

Let's try: https://pascalbirchler.com/internationalization-in-wordpress-5-0/

PR #1720, opened 2019-06-04
Add contact link form and checks for contact link display

dannylonglegs:
https://github.com/pressbooks/ideas/issues/180

### DIFF
--- a/inc/admin/laf/namespace.php
+++ b/inc/admin/laf/namespace.php
@@ -23,6 +23,21 @@ use Pressbooks\Metadata;
  * Add a custom message in admin footer
  */
 function add_footer_link() {
+/**
+ * Add checks to determine what contact link returns
+ */
+$contactForm = get_blog_option( get_main_site_id() , 'pb_network_contact_form');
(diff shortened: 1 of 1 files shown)


inc/admin/laf/namespace.php
```diff
@@ -23,6 +23,21 @@
...
+	$pb_network_contact_form = get_blog_option( get_main_site_id(), 'pb_network_contact_form' );
+	$pb_network_contact_link = get_blog_option( get_main_site_id(), 'pb_network_contact_link' );
+
+	if ( $pb_network_contact_form ) {
+		$contact_link = network_home_url( '/#contact' );
+	} else {
+		if ( ! empty( $pb_network_contact_link ) ) {
+			$contact_link = $pb_network_contact_link;
+		} else {
+			$contact_link = 'mailto:' . get_option( 'admin_email' );
```
connerbw: Can you double check this?

This code doesn't always run in the same book as Aldine.

`get_option` vs `get_site_option` vs `get_blog_option`

PR #1724, opened 2019-06-05
Replace PB-LaTeX with MathJax

connerbw:
See:

+ https://github.com/pressbooks/ideas/issues/189
+ https://github.com/pressbooks/ideas/issues/190

Depends on new microservice: 

+ https://github.com/pressbooks/pb-mathjax


connerbw:
+ This is a replacement for PB LaTeX and will be better in just about every way. Users who don't like/want it can still use WP QuickLaTeX as replacement, but WP QuickLaTeX will only accept LaTeX as input, only output as SVGs/PNGs. You can't use both our new solution and WP Quick LaTeX simultaneously -- it's one or the other. 
  + Along the same lines, this feature has a new `pb_mathjax_disabled` filter. Developers who wants to replace this plugin with their own, can. No more support for, or concept of, rendering engines. Come up with a better solution or send pull requests to solution. No more in-between-ing.
+ Supports AsciiMath (with [asciimath] as shortcode) and LaTeX as inputs, with MathML coming later.
+ Uses CommonHTML output in webbooks, SVGs in PDFs and PNGs in mobi/epubs,

This PR will remain open for a few days as we do more testing, and work out a few more details.

Instructions for open source users to install and run the pb-mathjax microservice, a new dependency written in node.js, coming soon.

greatislander:
Nice work, @connerbw. This looks awesome.

PR #1726, opened 2019-06-07
Use PDF/UA-1 for digital distribution PDFs (fix #1695)

### LINKED ISSUE #1695: Export Tagged PDFs for better accessibility
### Feature Description

PDFs would be more useful for screenreaders if they came tagged with the markup from the HTML. Currently exported digital distribution PDFs aren't tagged, but Prince has the options available. 

I was able to use prince to be able to make a tagged PDF from the command line, so I believe that it has the capability, it just needs an extra command.

https://www.princexml.com/doc-prince/#pdf-profiles

From the prince documentation, adding an extra command will make the PDFs tagged. "To enable tagged PDF without using one of the profiles that already imply it, the command-line option --tagged-pdf can be used."

### Feature Use Case

Tagged PDFs are indexed for assistive technology, so that users can easily move throughout the structure of the document.

### Other Notes

When I created a document on my computer with Prince using a Pressbooks source, the only errors thrown doing a PDF Full Check in Adobe Acrobat Pro were from Chapter names and Page numbers at the bottom of the page. I don't know if it is possible to add a tag to that element as the PDF is being created. I've attached the document here.

[Disease-Prevention-and-Healthy-Lifestyles-tagged.pdf](https://github.com/pressbooks/pressbooks/files/3124178/Disease-Prevention-and-Healthy-Lifestyles-tagged.pdf)

connerbw:
Requires Princexml 12 or Docraptor Pipeline 7

See:
 
 + https://www.princexml.com/doc-prince/#pdf-profiles
 + https://docraptor.com/blog/pipeline-7-with-flexbox-support-and-prince-12/

Digital PDFs will be (more) accessible.

Note: `--taged-pdf` not supported by Docraptor so "use the same PDF style we have been using for print, but add the --tagged-pdf option" is not something we will be doing at this time. Print PDFs will continue to be focused on printing the PDF, not reading the PDF on a computer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 dist: xenial
+services:
+  - mysql
 language: php
 sudo: required
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: php
 sudo: required
 notifications:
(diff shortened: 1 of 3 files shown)


connerbw:
This PR addresses #1695 but will remain open for a few days. We need to thoroughly test this change before deciding to merge it.

PR #1733, opened 2019-06-19
Fix admin menu for networks that don't allow new books

connerbw:


SteelWagstaff:
Works as intended on textopress, which uses a subdomain installation method. Network managers and super admins can create books from their network manager menu, and can clone books. Regular users on the network no longer see the create a book or clone book navigation menus and will see a Book Permissions interface menu when they do not have permissions in any books on the network. I have not tested on subdirectory installation (like university.pressbooks.pub), but can do so if desired.

PR #1747, opened 2019-07-25
Support install using github-updater.

BillDStrong:
This is a very simple change, just adding meta-data to force github-updater to use releases when installing pressbooks. 

github-updater is smart enough to be able to install without this, but pressbooks doesn't work due to missing dependencies. These dependencies are already in the release files, which will facilitate this workflow.

### DIFF
--- a/pressbooks.php
+++ b/pressbooks.php
@@ -2,6 +2,8 @@
 /*
 Plugin Name: Pressbooks
 Plugin URI: https://pressbooks.org
+GitHub Plugin URI: pressbooks/pressbooks
+Release Asset: true
 Description: Simple Book Production
 Version: 5.9.1-dev
 Author: Pressbooks (Book Oven Inc.)


BillDStrong:
If this is accepted, the same can be done for the aldi theme, as it has a similar issue.

connerbw:
Can you explain this PR? We use:

https://github.com/YahnisElsts/plugin-update-checker

Pressbooks automatically updates from GitHub already. 

Is this for: https://github.com/afragen/github-updater ?

BillDStrong:
@connerbw 
Correct, this is for [https://github.com/afragen/github-updater](https://github.com/afragen/github-updater).

I use this to install plugins. I noticed when installing pressbooks, it required dependencies that are only added in the pressbook releases. This just forces github-updater to use the release, instead of the master branch. With this patch, github-updater will only update on new releases, as github-updater will only choose to use either the main repository, in which case you can switch between branches on your site, or releases which force you to use only the releases.

What you currently use works for after you install the plugin, where-as github-updater allows to just copy and past the repo link to install, and update from that. 

The downside of this approach, is github-updater is really useful for devs working on the project, since it allows switching branches in WP. You won't get that functionality, due to forcing releases. Allowing for that use case would require more work, and maintenance for the project, which may be out of scope. (Either adding dependencies to the repo, or a script that auto downloads said dependencies.)

Consider this a defensive measure to simply make github-updater work without having to do much.
 For the new install, it makes things easier for those used to the WordPress.org installer, don't want to download and upload zip files, and are unfamiliar with git.

Of course, you can choose not to use this, but it was very simple to add and maintain, so :).

connerbw:
Neat. I don't know https://github.com/afragen/github-updater

This seems fine to me.  Maybe a use case to drop "YahnisElsts/plugin-update-checker" support and just tell people to install github-updater? Will look harder, later.

Issue #1508, opened 2018-12-11
Unless chapter is licensed differently, it should display the book license statement


greatislander:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

As [reported](https://discourse.pressbooks.org/t/rationale-for-chapter-specific-attribution-statements-in-webbook/804/) by @josiegray, chapter-level license statements should indicate the book's license (e.g. "Chapter Title from _Book Title_ by Author Name is licensed…") unless the chapter has a license which differs from that of the book. 

### Steps to Reproduce

Not applicable.

### System Information

Pressbooks: 5.6.x
McLuhan: 2.6.x

greatislander:
### Implementation Notes

The license statement for chapters with no specified license should indicate the license of (and link to) the parent book:

> _Book Title_ by Author Name is licensed… etc.

The license statement for chapters with an individually specified license should indicate the license of (and link to) the chapter itself:

> _Chapter Title_ by Author Name is licensed… etc.

PR #1758, opened 2019-08-24
21 Aug 2017 result new commit

lingma:
This is a new patch with all the failed test corrected.


inc/class-licensing.php
```diff
@@ -220,8 +220,8 @@ public function doLicense( $metadata, $post_id = 0, $title = '' ) {
 			return '';
 		}
 
-		// Unless section is licensed differently, it should display the book license statement
-		if ( $section_license === $book_license || empty( $section_license ) ) {
+		// Unless section is empty, it should display the book license statement.
+		if ( empty( $section_license ) ) {
```
connerbw: If you do git blame (git-annotate) you'll see that this line is linked to this issue:

> Unless chapter is licensed differently, it should display the book license statement 
https://github.com/pressbooks/pressbooks/issues/1508

We want to preserve that fix. 

They key part of the new bug is:

> Chapter license statement reflects only attribution for that specific chapter ***when a chapter author has been defined.***
https://github.com/pressbooks/pressbooks-book/issues/588

Probably (untested):

```php
if (
	( $section_license === $book_license && empty( $section_author ) )
	|| empty( $section_license )
) {
```

PR #1759, opened 2019-08-24
21 Aug 2019 new request

lingma:
Changed the logic, firstly I ignored part of the comment by Dac then read again before the pull request, so that I could modify a potential bug.


lingma:
This patch is OK, though the test coverage is not good. The program will run fine, though the test case is not enough. Doing more test will need to consider more modification. I just let it be for now.

PR #1760, opened 2019-08-24
21 Aug 2019 part 1

lingma:
Separate the problem, first part commit.


lingma:
TODO: The license part test cases need more work, while the code itself works via manual verification. I have not got time to check in details while I have not found simple way in the current test.

PR #1829, opened 2019-11-08
Escape my catalog output

connerbw:
When this feature was developed 7 years ago (?!) it was implied that HTML tags were allowed. This PR sanitizes saving the HTML into the database. For things already in the database, it escapes.

It also fixes the "sidebar color" feature from 2016: https://github.com/pressbooks/pressbooks/commit/3756c22fddecc3f8301625d09bde70719ba07472

Time flies.

### DIFF
--- a/inc/class-catalog.php
+++ b/inc/class-catalog.php
@@ -754,7 +754,12 @@ class Catalog {
 			} elseif ( '%f' === $this->profileMetaKeys[ $key ] ) {
 				$val = (float) $val;
 			} else {
-				$val = (string) $val;
+				$val = HtmLawed::filter(
+					$val, [
+						'safe' => 1,
+						'deny_attribute' => 'style',
+					]
+				);
 			}
 
 			update_user_meta( $this->userId, $key, $val );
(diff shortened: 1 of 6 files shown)


ho-man-chan:
At a glance, the string escape seems to work as expected. I would test with interesting strings such as `<<a>script>alert('ciao');<</a>/script>` and other weird combination to ensure no crafty inputs will pass. This really depends on how PHP and WordPress builts their escapes and strip_tag functions. I will try to play around more.

ho-man-chan:
Further tested with `<<a>script>alert('ciao');<</a>/script>` and various string provided in the following link:

ho-man-chan:
Catalog Sidebar background on non-empty colour option working.

inc/class-catalog.php
```diff
@@ -754,7 +754,12 @@ function saveProfile( array $item ) {
 			} elseif ( '%f' === $this->profileMetaKeys[ $key ] ) {
 				$val = (float) $val;
 			} else {
-				$val = (string) $val;
+				$val = HtmLawed::filter(
```
cagp-dev-mtl: This code allows to input html tag and removes the "style" attribute if present, I think this is the intended behavoir.

connerbw: Yeah. It's intended. I went with this recommendation:

> Admins, however, may still want to completely deny the style attribute, e.g., with code like
> `$processed = htmLawed($text, array('safe'=>1, 'deny_attribute'=>'style'));`
> Permitting the style attribute brings in risks of click-jacking, etc. CSS property values can render a page non-functional or be used to deface it. 

Source:
http://www.bioinformatics.org/phplabware/internal_utilities/htmLawed/htmLawed_README.htm#s3.6

Issue #1822, opened 2019-11-03
Make 'Save Book Information' box sticky


SteelWagstaff:
The book info page contains a 'Save Book Information' box. When users have set screen options to 2 column and are viewing the page at sufficient width, this box sits at the top right of the page (CSS rules target `#post-body.columns-2 #postbox-container-1`). 

![Screenshot from 2019-11-03 14-28-16](https://user-images.githubusercontent.com/13485451/68091501-390f9880-fe46-11e9-87a5-4a5c20d5798a.png)

For better UX, it should be made sticky as the user scrolls down a very long page. See https://alligator.io/css/position-sticky/ and https://caniuse.com/#feat=css-sticky for some implementation ideas.

SteelWagstaff:
Quick & dirty CSS (for wide page views): 
`position: fixed; z-index: 1; right: 20px; margin-right: 0; float: none;`

mcgratay:
👍

PR #1830, opened 2019-11-12
Update book-info column 2 to be fixed on large screen

ho-man-chan:
1. Book info's right side column 2 is fixed for large screens.
2. Tested on the following browsers: Chrome, Safari, Firefox and Edge.
3. All tests have passed locally.


ho-man-chan:
Add Issue id #1822 for tracking

connerbw:
> Are we intentionally versioning content under "dist" folder?

When we package a plugin we delete the `assets/src` dir. Pressbooks has chosen to be responsible for the compiled files in `assets/dist` Some other companies will let the CI pipeline handle that. We don't. The files are intentionally versioned.

PR #1833, opened 2019-11-13
Pb 1823 bulk add users

cagp-dev-mtl:
### PR description
This PR allows to bulk add users to a book through a new submenu available in the book context.

### Context
Some Pressbooks users create collaborative books and is time consuming adding users individually, e.g., professors assigning several students as book contributors.

### :shipit: How can this PR be tested?

- Checkout branch: pb_1823_bulk_add_users
- Navigate to any book
- Click on the 'Bulk add' submenu located under the 'Users' menu.
Bulk Add ‹ The onboarding process — Pressbooks

**Test case 1:**

1.  In the textbox, enter emails as follows:
- 1 email address of an existing user that is not linked to the current book.
- 1 new email address.
- 1 new email with an username that is less than 3 characters long.

**Note:** New email addresses should not be previously used or waiting for user email confirmation.
 
2.  Select a role from the dropdown and click "Add users".

:ok_hand:  **Expected results:**

1.  A WordPress confirmation message should appear displaying the list of email addresses that were associated to the book, e.g.:
Success
2. The user associated to the existing email address should be added to the book.
3. 1 email confirmation should be sent to every new user. After activating the accounts by clicking on the links of the confirmation emails, new users should be added to the book.
4.  In the case of the user email that had less than 3 characters, the account associated to the book should have an username that is 4 characters long, the email address remains the same.
5. All users added in bulk must have the role that was selected in the dropdown when submitting the form.

**Test case 2:**

1.  Add a new email with an existing username but different domain, submit the form and activate the user by clicking the activation link that was sent for that account. 

:ok_hand:  **Expected results:**
1.  A new user account is created and associated to the book, the username has been disambiguated and the email address remains the same.

**Test case 3**
1.  Add a new email  but do not activate it.
2.  Add the same email.

:ok_hand:  **Expected results:**
1.  You will receive an activation email.
2.  As long as you have not activated the email, you would not be able to add emails that contains the same username.
3.  You should see an error message similar to the one below:
Error-message

**More tests welcome!**

### Considerations / Specs
If a user with one of the email addresses already exists in the network, that user should be added to the book. If one does not exist, a new user invitation should be sent, using the text string that precedes the @ sign for the user name. If a username already exists for the string or the string is not 4 characters long, extra characters should be added for disambiguation/padding. Following the completion of the activity, some kind of status message should be shown to the user who added the new users.

### Unit tests
Confirm that all test pass by running the following command inside your vagrant machine :
```/usr/bin/php /srv/www/pressbooks.test/current/vendor/phpunit/phpunit/phpunit --group userbulk --configuration /srv/www/pressbooks.test/current/web/app/plugins/pressbooks/phpunit.xml```

### Related tickets
[#1823](https://github.com/pressbooks/pressbooks/issues/1823)

### PR type
- [x] Feature
- [ ] Bug

### Other Related PRs:
- [ ] Depends on another PR (List dependencies below)

### DIFF
--- a/hooks-admin.php
+++ b/hooks-admin.php
@@ -356,3 +356,7 @@ add_filter( 'gettext', '\Pressbooks\Utility\change_recommendations_sentence', 10
 // Theme check
 add_action( 'admin_init', '\Pressbooks\Theme\check_required_themes' );
 add_action( 'admin_init', '\Pressbooks\Theme\check_upgraded_customcss' );
+
+if ($is_book) {
+	add_action( 'init', [ '\Pressbooks\Admin\Users\UserBulk', 'init' ] );
+}
--- a/hooks-admin.php
+++ b/hooks-admin.php
@@ -357,6 +357,6 @@ add_filter( 'gettext', '\Pressbooks\Utility\change_recommendations_sentence', 10
 add_action( 'admin_init', '\Pressbooks\Theme\check_required_themes' );
 add_action( 'admin_init', '\Pressbooks\Theme\check_upgraded_customcss' );
 
-if ( $is_book ) {
-	add_action( 'init', [ '\Pressbooks\Admin\Users\UserBulk', 'init' ] );
-}
+// Bulk add users
+add_action( 'init', [ '\Pressbooks\Admin\Users\UserBulk', 'init' ] );
+
--- a/hooks-admin.php
+++ b/hooks-admin.php
@@ -357,6 +357,6 @@ add_filter( 'gettext', '\Pressbooks\Utility\change_recommendations_sentence', 10
 add_action( 'admin_init', '\Pressbooks\Theme\check_required_themes' );
 add_action( 'admin_init', '\Pressbooks\Theme\check_upgraded_customcss' );
 
-if ($is_book) {
+if ( $is_book ) {
 	add_action( 'init', [ '\Pressbooks\Admin\Users\UserBulk', 'init' ] );
 }
--- a/templates/admin/user_bulk_new.blade.php
+++ b/templates/admin/user_bulk_new.blade.php
@@ -1,0 +1,30 @@
+<div class="wrap">
+	<h1>{{ __( 'Bulk Add New User', 'user' ) }}</h1>
+	<form method="POST" action="{{ $form_url }}" method="post">
+		<?php echo wp_nonce_field( $nonce ); ?>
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: templates/admin/user_bulk_new.blade.php [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+					<div class="form-field term-description-wrap">
+						<textarea name="users" id="users" rows="5" cols="50" spellcheck="false"></textarea>
+						<p>{{ __('One entry per line', 'user') }}</p>
+					</div>
+				</td>
+			</tr>
+			<tr class="form-field">
+				<th scope="row"><label for="adduser-role"><?php _e( 'Role' ); ?></label></th>
+				<td><select name="role" id="adduser-role">
+						<?php wp_dropdown_roles( get_option( 'default_role' ) ); ?>
```
connerbw: Avoid using `<?php` in Blade template files. Instead use `@php` ...`@endphp`

(Because you don't want a user going to https://x/y/z.blade.php and running the code.)

hooks-admin.php
```diff
@@ -356,3 +356,7 @@
 // Theme check
 add_action( 'admin_init', '\Pressbooks\Theme\check_required_themes' );
 add_action( 'admin_init', '\Pressbooks\Theme\check_upgraded_customcss' );
+
+if ( $is_book ) {
```
connerbw: I know this is my fault, but `$is_book` logic should be moved into `hooks()`. Example: 

https://github.com/pressbooks/pressbooks-saml-sso/blob/6c0fa5d1154d7d5cdeafbfd9ad400711c8659413/inc/class-updates.php#L27

This is so the class can decide where to print the menu or not.  Useful if we extend this feature to be in both Book and Network Admin.

templates/admin/user_bulk_new.blade.php
cagp-dev-mtl: `<?php` replaced by `@php` and  `?>` replaced by `@endphp`

hooks-admin.php
cagp-dev-mtl: `$is_book` check moved into `hooks`

cagp-dev-mtl:
Thank you @ho-man-chan, I have addressed the issues you pointed by adding $_POST sanitization to every email received in the post argument and by enforcing the the role parameter.

PR #1853, opened 2019-11-25
PB-1823 - Swap invited and new added user success messages

cagp-dev-mtl:
Update new user and added user success messages.

### DIFF
--- a/inc/admin/users/class-userbulk.php
+++ b/inc/admin/users/class-userbulk.php
@@ -58,8 +58,8 @@ class UserBulk {
 	public function addMenu() {
 		add_submenu_page(
 			self::PARENT_SLUG,
-			__( 'Bulk Add', 'user' ),
-			__( 'Bulk Add', 'user' ),
+			__( 'Bulk Add', 'pressbooks' ),
+			__( 'Bulk Add', 'pressbooks' ),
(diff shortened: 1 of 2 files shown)


inc/admin/users/class-userbulk.php
```diff
@@ -238,10 +238,10 @@ public function getBulkMessageSubtitle( $status ) : string {
 				$subtitle = __( 'The following user(s) could not be added.', 'users' );
 				break;
 			case self::USER_STATUS_INVITED:
-				$subtitle = __( 'An invitation email has been sent to the user(s) below. A confirmation link must be clicked before their account is created.', 'users' );
+				$subtitle = __( 'User(s) successfully added to this book.', 'users' );
```
connerbw: The translations functions should always have the text domain as the 2nd param. For this plugin it's `pressbooks`

https://github.com/pressbooks/pressbooks/blob/dd283c3728ef124fa328ae1b0120e064630a8f5f/pressbooks.php#L11

connerbw: See: https://developer.wordpress.org/reference/functions/__/

cagp-dev-mtl:
PB-1823 - Update unit get subtitle message test

PR #1860, opened 2019-11-27
Update version and localizations, build assets

cagp-dev-mtl:
First tandem release @ho-man-chan & @cagp-dev-mtl


.tx/config
```diff
@@ -2,8 +2,9 @@
 host = https://www.transifex.com
 
 [pressbooks.pressbooks]
+file_filter = languages/pressbooks-<lang>.po
```
connerbw: Put this back, no diff please.

cagp-dev-mtl: Sure, file has been reverted.

Issue #1845, opened 2019-11-23
Fix styling changed in WP 5.3 to look like PB


SteelWagstaff:
Rollback unwanted changes by overriding the affected classes with the styles of the previous WP version (primarily color)

connerbw:
@cagp-dev-mtl 

Can you check this page? It's new in 5.3.0

![image](https://user-images.githubusercontent.com/812192/69759636-a1455780-1130-11ea-8706-5bab44c1bcd8.png)

 + https://github.com/WordPress/WordPress/blob/6f86cedb09f4ad42c5443cf5cc0c46b30bd41010/wp-login.php#L611
+ https://pressbooks.test/wp/wp-login.php?redirect_to=https%3A%2F%2Fpressbooks.test%2Fwp%2Fwp-admin%2F&action=confirm_admin_email

@SteelWagstaff @mcgratay Do we want to suppress this feature? Anyone with `[manage_options](https://wordpress.org/support/article/roles-and-capabilities/#manage_options)` will see this page...

mcgratay:
Can verify that network managers would see this page. I'd recommend we disable it since the preferred default is that all network and site admins are set to ops@pressbooks.com. It also looks like it may pop up on book logins. 

Not finding any unwanted blue, so other than the above, everything looks good to go.

cagp-dev-mtl:
@mcgratay ,  I have addressed your comments + most of Dac's notes attached in the [PR](https://github.com/pressbooks/pressbooks/pull/1867). 

Admin email screen confirmation page should not appear anymore, I added a hook that Dac sent me to block those emails. (Locally did not see the email confirmation window after adding the hook)

I could not change the color of the checkboxes arrows, I tried a couple of solutions including replacing the content attribute by an awesome icon, but the arrows appear misplaced out of the checkbox. I think if we want to address that particular issue we need to reasearch more....

mcgratay:
Two hopefully quick fixes: 

1) The page size dropdown on the book list was fixed but that apparently didn't affect the user list one: 

![Screen Shot 2019-12-05 at 11 34 33 AM](https://user-images.githubusercontent.com/39064028/70254523-365cd780-1753-11ea-8ccb-1414a1123f9f.png)

And 2) Now that the highlights that appear on selected objects has been switched from blue to red, there are some discrepancies through different features. The most obvious is on the Media Library page where any selected media or search box is highlighted blue: 

[...]

PR #1861, opened 2019-11-27
PB-1845 - Revert some a11y styles introduced in WP 5.3

cagp-dev-mtl:
## Reset some of the a11y styles introduced in WP 5.3

This PR resets some of the a11y style changes introduced for the Admin interface in WordPress 5.3  

## Context
WordPress 5.3 introduced a new set of css rules intended to improve a11y, some of these changes do not match our brand colors and Pressbooks already has a a11y mode. 

A decision has been made to change the border color to red for all buttons and to change the text color to dark grey for drop downs. When pressbooks a11y mode is enabled, al buttons must have a black border and dark fonts when the button has no background.

Previous versions of WordPress already introduced a blue focus border, we already have this border in production, thus this is out of the scope of this ticket. 

## Related tickets
Fix styling changed in WP 5.3 to look like PB [#1845](https://github.com/pressbooks/pressbooks/issues/1845)


cagp-dev-mtl:
@connerbw, @ho-man-chan , I just commited the fix for the admin email confirmation buttons, notice that the green color comes from my customized theme:

**Before:**
Admin-email-confirmation

**After:**  
Admin-email-confirmation png-fixed

cagp-dev-mtl:
Also removed border radius from login inputs.

connerbw:
# Steel mentioned this in a call:

image


# This seems odd:

Fixme2a
Fixme2b

# Too bunched up, can we do something better?

Can we do something better here 1

# New CSS affects Network Analytics plugin in a weird way

Network Analytics Plugin

## I am going to approve the PR. I think testing will come back with more than these issues. I think there will need to be a second pass with a goal on making things look better. See civilized flame war at bottom of this thread for different points of view.

https://make.wordpress.org/core/2019/10/18/noteworthy-admin-css-changes-in-wordpress-5-3/#comment-37328

PR #1868, opened 2019-12-04
Remove fix for table role (see #1641)

greatislander:
#1657 implemented a fix to compensate for WordPress's use of tables for layout without an appropriate ARIA role indicating that they were presentational. My core patch (https://core.trac.wordpress.org/ticket/46899) was merged and released in WordPress 5.2.3, so this JavaScript is no longer required as all tables generated via the Settings API include an appropriate ARIA role indicating that they are presentational in nature.


SteelWagstaff:
Thanks @greatislander!

PR #1873, opened 2019-12-10

cagp-dev-mtl:

### DIFF
--- a/compatibility.php
+++ b/compatibility.php
@@ -33,7 +33,7 @@ function pb_meets_minimum_requirements() {
 
 	// WordPress Version
 	global $pb_minimum_wp;
-	$pb_minimum_wp = '5.2.4';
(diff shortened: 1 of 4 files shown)


compatibility.php
```diff
@@ -33,7 +33,7 @@ function pb_meets_minimum_requirements() {
 
 	// WordPress Version
 	global $pb_minimum_wp;
-	$pb_minimum_wp = '5.2.4';
+	$pb_minimum_wp = '5.3';
```
connerbw: Does this work?

https://stackoverflow.com/questions/10997482/php-version-compare-returns-1-when-comparing-5-2-and-5-2-0

cagp-dev-mtl: It worked by I did update it to '5.3.0' to be consistent with the format.

PR #1935, opened 2020-04-29
Add book directory deletion verification

cagp-dev-mtl:
@ho-man-chan , this PR addresses the deletion validation requirement for the book directory.

There is a new parameter 'sid' in the POST body for the deletion triggers that is sent to the delete endpoint of the book fetcher:
```
$data = [
		'sid'       => $sid,
		'network'   => 'https://' . $_SERVER['HTTP_HOST'],
		'book_id'   => $book_id,
	];
```
Once you receive parameter 'sid', you need to send it back to the network in a **GET** request to the validation deletion endpoint with the following format:

```
https://[ORIGIN NETWORK URL]/wp-json/pressbooks/v2/directory/verify_deletion?sid=[SID VALUE]
```

If you receive **true** as response then you can delete the book otherwise don't.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ coverage.xml
 node_modules
 tests/logs/
 vendor
+coverage-reports
--- a/composer.json
+++ b/composer.json
@@ -78,6 +78,10 @@
 			"vendor/bin/phpunit --configuration phpunit.xml --coverage-clover coverage.xml",
 			"@standards"
 		],
+		"test-coverage": [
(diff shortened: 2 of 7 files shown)


cagp-dev-mtl:
@ho-man-chan , I have added support for an env. variable 'CUSTOM_DELETE_BOOK_ENDPOINT' that allows to customize the deletion endpoint. This env variable will allow to set a custom endpoint for  deleting books from textopress and integrations.

cc: @SteelWagstaff , @ricardopressbooks

Issue #1797, opened 2019-09-25
Build v3 of API for books 


SteelWagstaff:
Goal: include data/structure as needed for public book list/catalogue improvements and speed/performance optimization

Information to expose for each book in v3 of API:
1. Word count
2. Storage size
3. number of H5P activities
4. Book in catalog (Boolean)

We also want to be able to store/query the network root URL (so that users can see only books from a given network/networks). Don’t know if we need to get from API, as we could pretty easily extrapolate from the book’s URL. 

Running queries could be a soft task (could extract basic API book query) and then manually join with these three data points. See #1800

TO DO: attach benchmarks for queries & API responses

- [x] Add data to v3 of API
- [x] Document as needed
- [ ] Update directory data fetcher to store new information in Algolia

SteelWagstaff:
Info not currently available from book API but which we would like to have in v3:
1. Word count
2. Storage size
3. # of H5P activities

SteelWagstaff:
Running queries could be a soft task (could extract basic API book query) and then manually join with these three data points. See https://github.com/pressbooks/pressbooks/pull/1800

SteelWagstaff:
TO DO: attach benchmarks for queries & API responses

ricardopressbooks:
PRs:
- https://github.com/pressbooks/pressbooks/pull/1938
- https://github.com/pressbooks/book-directory-fetcher/pull/15

Scope:
I discussed with @SteelWagstaff about the scope for this tasks and it includes:
- Add the 3 fields required for API v3
- Update docs in https://docs.pressbooks.org/rest-api/, that means `wp-json/pressbooks/v2?context=help` endpoint should expose the new fields as well.

PR #1938, opened 2020-05-06
Pb 1797 new fields

ricardopressbooks:
https://github.com/pressbooks/pressbooks/issues/1797


cagp-dev-mtl:
@ricardopressbooks , let's spend a couple of minutes to explore the single meta array approach you mentioned this mornign, thanks.

PR #1948, opened 2020-05-22
Remove directory endpoint test

cagp-dev-mtl:


cagp-dev-mtl:
The book directory deletion endpoint test passes locally however when running in travis it breaks. I am removing the endpoint check.

Issue #1918, opened 2020-04-10
XHTML & PDF exports can't handle LaTeX expressions which include > (greater than sign)


SteelWagstaff:
When users create a math expression that includes the `>` sign, there are some situations in which `>` character is not properly escaped in the image alt text and/or title attribute (no matter whether the user is using MathJax or WP QuickLaTeX). As a result, the image element closes prematurely in the XHTML output, and the rest of the image attributes following the alt text are displayed in the resulting PDF.  Compare https://integrations.pressbooks.network/latextest/format/xhtml?debug=prince with https://integrations.pressbooks.network/latextest/format/xhtml or see examples of PDF exports at https://integrations.pressbooks.network/latextest/wp-admin/options.php?page=pb_export or the screenshots below:
![Screenshot from 2020-04-14 19-53-59](https://user-images.githubusercontent.com/13485451/79287545-bc1f2b00-7e89-11ea-84cc-4422f7875bc4.png)

![Screenshot from 2020-04-09 22-02-24](https://user-images.githubusercontent.com/13485451/78958513-d8981d80-7aad-11ea-9299-6110e3bc8c6f.png)

I'm pretty confused by this bug because I can't reliably reproduce it even on the same network. For example, see the attached XHTML file export:
[Math-Book-1593636886.zip](https://github.com/pressbooks/pressbooks/files/4860501/Math-Book-1593636886.zip)

The _identical_ LaTeX expression `x > 11` appears in the first two chapters. In one chapter it renders just fine and in the second it breaks. I have no idea how to explain this variation.

SteelWagstaff:
Appears related to https://github.com/pressbooks/pb-mathjax/issues/21

SteelWagstaff:
Can be fixed by adding something like
`
$old_alt = $image->getAttribute( 'alt' );
			$new_alt = htmlspecialchars( $old_alt_);
			if ( $old_alt !== $new_alt ) {
				$image->setAttribute( 'alt', $new_alt );
				$changed = true;
			}
` 
to the foreach loop here: https://github.com/pressbooks/pressbooks/blob/bbe60a188d4ae73182fb14e22dc233c1a82e48d4/inc/modules/export/xhtml/class-xhtml11.php#L731-L751

SteelWagstaff:
[...]

PR #1969, opened 2020-06-23
Escape special characters in image alt and title properties

SteelWagstaff:
Escape special characters in image alt and title properties as part of XHTML export routine. Proposed fix for #1918.


SteelWagstaff:
Caveat: I was unable to test completely with PDF exports because of local MathJax configuration issues.

Issue #1972, opened 2020-06-30
New data-url attribute isn't added to front & back matter


SteelWagstaff:
The new method which is supposed to add data-url attributes to all links during the XHTML export routine misses front and back matter. I haven't looked very closely yet, but suspect the problem may be related to this loop: https://github.com/pressbooks/pressbooks/blob/66429e41f0cb567c2a1b5ef71a3a55d1b207c404/inc/modules/export/xhtml/class-xhtml11.php#L540-L581 ?

SteelWagstaff:
The issue ended up being with the 'changed' statements we were previously using to tell the export routine whether to use an altered version of the source content or not.

PR #1973, opened 2020-07-01
Improve logic for fixInternalLinks function

SteelWagstaff:
Fix for https://github.com/pressbooks/pressbooks/issues/1972


inc/modules/export/xhtml/class-xhtml11.php
```diff
@@ -665,18 +663,13 @@ protected function fixInternalLinks( $source_content, $id = null ) {
...
 				}
 			}
 		}
 
-		if ( ! $changed ) {
-			return $source_content;
-		} else {
-			$content = $html5->saveHTML( $dom );
+		$content = $html5->saveHTML( $dom );
 			return $content;
```
cagp-dev-mtl, line 672: Please align `return $content;` at the same indentation level than $content

Issue #1831, opened 2019-11-13
Build PDF footnote fix for block elements using JS hack


SteelWagstaff:
When a user creates a footnote which includes one or more HTML block elements, the footnote breaks in the XHTML and PDF exports. This happens because the footnote is currently being rendered as a `<span>`, an inline element which cannot contain block elements and remain valid HTML. For example, if an author puts a second paragraph or a block quote within a footnote, the footnote will display normally in the webbook, but not the PDF exports. Here's how the HTML in webbook looks: 
![Screenshot from 2019-12-09 13-54-04](https://user-images.githubusercontent.com/13485451/70468056-fc117400-1a8b-11ea-8fa5-b5370f167a28.png)

Here's how that same footnote renders in a PDF export:
![Screenshot from 2019-12-09 13-52-21](https://user-images.githubusercontent.com/13485451/70467640-32022880-1a8b-11ea-8e6d-242fb705dbce.png)

See footnote 1 here: https://integrations.pressbooks.network/steelstest/chapter/footnote-test/ and page 48 (54) here: https://integrations.pressbooks.network/steelstest/open/download?type=print_pdf

We cannot leave the footnote element in place and process as a `<div>` because these elements cannot live within `<p>` elements either. See https://stackoverflow.com/a/10763952. 

We've discussed in the [PrinceXML forums](https://www.princexml.com/forum/topic/4382/block-elements-in-paragraphs-footnotes?p=1#22060), where Mike Day told us:
> The tricky part is that Prince allows this via the CSS display property, but HTML syntax does not allow nesting blocks within spans and is unaware of the CSS, so it breaks the markup.

> I think it would be possible to solve this with JavaScript, in which the footnote block content is elsewhere in the document and a script moves it into place, allowing a block to be inside a span without breaking everything.

A sample implementation that we'd like to try in Pressbooks was provided by Mark Brown:

> Here's an example. The footnote elements are placed outside the `<p>` elements so as not to break the html, and are linked by another element with class "indirect". The script moves any footnotes into their correct places.

```
<script>
ind = document.getElementsByClassName("indirect");
for (var i=0; i<ind.length; i++) {
    var e = document.getElementById(ind[i].getAttribute("ref"));
    if (e) ind[i].appendChild(e);
}
</script>

<p>
A paragraph containing a footnote.<span style="float: footnote;">
    <span class="indirect" ref="fn1"></span>
</span>
Content in the same paragraph.
</p>

<div id="fn1">
    <ul>
[...]

PR #1974, opened 2020-07-02
Pb 1831 footnote issue refactor

ricardopressbooks:
https://github.com/pressbooks/pressbooks/issues/1831

## How to test
- Create a footnote in a chapter with different elements like lists, images, quotes, etc
- Create more than 1 footnote in the same paragraph

Check in web and PDF export footnotes have the expected behavior.


assets/src/scripts/export-footnotes.js
```diff
@@ -0,0 +1,10 @@
+function moveFootnotes() {
+	var footnotes = document.getElementsByClassName( 'footnote-indirect' );
```
cagp-dev-mtl, line 2: Pls change all `var` for `let`

inc/modules/export/prince/class-pdf.php
```diff
@@ -119,6 +120,7 @@ function convert() {
 		// Save PDF as file in exports folder
 
 		$prince = new \PrinceXMLPhp\PrinceWrapper( PB_PRINCE_COMMAND );
+
```
cagp-dev-mtl, line 123: remove trailing new line.

inc/modules/export/xhtml/class-xhtml11.php
```diff
@@ -520,6 +529,32 @@ function doEndnotes( $id ) {
...
+	 * Content for footnotes.
+	 *
+	 * @see footnoteShortCode
+	 *
+	 * @param $id
+	 *
+	 * @return string
+	 */
+	function doFootnotes( $id ) {
+
```
cagp-dev-mtl, line 542: trailing newline (remove)

inc/modules/export/xhtml/class-xhtml11.php
```diff
@@ -353,14 +359,14 @@ public function transformGenerator() : \Generator {
...
 		}
-
 		echo "</head>\n<body lang='{$this->lang}' ";
 		if ( ! empty( $_GET['optimize-for-print'] ) ) {
 			echo "class='print' ";
 		}
 		echo ">\n";
 		$replace_token = uniqid( 'PB_REPLACE_INNER_HTML_', true );
 		echo $replace_token;
+
```
cagp-dev-mtl, line 369: Trailing newline (Remove)

inc/modules/export/prince/class-pdf.php
```diff
@@ -136,6 +138,12 @@ function convert() {
 		}
 
 		$prince->addStyleSheet( $css_file );
+
+		// Footnote hack to allow more complex DOM
```
cagp-dev-mtl, line 142: pls remove comment

cagp-dev-mtl:
In the future try to use jQuery, e.g. `export.js`

`var footnotes = document.getElementsByClassName( 'footnote-indirect' );`

becomes

`let $footnotes = $( '.footnote-indirect' );`

etc

if you are looping a jQuery object you can use:

`$footnotes.each( function() { .................. } )`

ricardopressbooks:
@cagp-dev-mtl thank you for your feedback.
I did some fixes according to your requests.

About JS Script I tried before to use jQuery with this script (instead the current):
```js
function moveFootnotes() {
	let footnotes = $( '.footnote-indirect' );
	for ( let i = 0; i < footnotes.length; i++ ) {
		let ref = $('#' + footnotes[i].attr('data-fnref'));
		if (ref) footnotes[i].append(ref);
	}
}
if ( typeof Prince != 'undefined' ) {
	moveFootnotes();
}
```

But seems to be jQuery is not available for Prince in our current configuration. I guess doesn't makes sense for this small and simple script, what do you think?

SteelWagstaff:
May be helpful? https://www.princexml.com/doc/javascript/ + https://www.princexml.com/doc/prince-input/#applying-javascript-in-prince

ricardopressbooks:
> May be helpful? https://www.princexml.com/doc/javascript/ + https://www.princexml.com/doc/prince-input/#applying-javascript-in-prince

Yes I actually used that approach for this solution.

ricardopressbooks:
@cagp-dev-mtl thank you for the feedback.
I tried with both (using `$` and `jQuery`) and doesn't work. This is my logs locally trying to use for example jQuery (applying same script you shared):
```
Mon Jul 13 16:51:54 2020: ---- begin
Mon Jul 13 16:51:54 2020: https://pressbooks.test/app/plugins/pressbooks/assets/dist/scripts/export-footnotes.js?id=ef3dd8e81c30fbc37b41:1: error: ReferenceError: no such property: jQuery
Mon Jul 13 16:51:54 2020: finished: success
Mon Jul 13 16:51:54 2020: ---- end
```
I fixed the syntax issue and I applied brackets correctly.

ricardopressbooks:
@SteelWagstaff could you test if it works for Prince in some network? Otherwise we should revert this PR as well according we talked with @cagp-dev-mtl during our planning meeting.

PR #1984, opened 2020-07-17
Pb 1831 footnote issue refactor

ricardopressbooks:
Footnote fix: Docraptor support


ricardopressbooks:
Note: Docraptor wrapper JS Engine for Prince doesn't allow `let` keywork (ES6) or `jQuery`.

ricardopressbooks:
@cagp-dev-mtl docraptor engine doesn't allow it (including Prince JS engine wrapper). Docraptor logs show the following error:
`Running scripts... SyntaxError: unexpected token footnotes` when I using `let`.

PR #2031, opened 2020-09-07
Avoiding null or false value in about field in PB API

ricardopressbooks:
https://github.com/pressbooks/pressbooks-book-directory-fe/issues/96


SteelWagstaff (approved):
LGTM

PR #2037, opened 2020-09-25
Update book api to filter by epoch time using modified_since param

ho-man-chan:
issue: pressbooks/book-directory-fetcher#34

Fix & tests:

1. In the book API endpoint (https://pressbooks.test/wp-json/pressbooks/v2/books?modified_since=1600992000), the modified_since param now takes in the epoch time instead of the YYYY-MM-DD. The book api will show the books that were last updated after the epoch time. This gives the user more control over.

   Test:
   1. Go to the book API endpoint URL 
ex: https://pressbooks.test/wp-json/pressbooks/v2/books?modified_since=1600992000
   2. Then type in the desired epoch in the URL param `modified_since=<some epoch time>` and press enter.
   3.  A new list should appear with the books that were last modified since the epoch time.


richard015ar:
LGTM Ho Man, but I think you should increment the coverage to cover the new lines.

richard015ar (approved):
LGTM!

Issue #2044, opened 2020-10-22
Send main site domain as network for subdomain book networks


ho-man-chan:
Book directory fetcher does not recognize networks that use book URL in their network URL.
The network URL we sent to the book directory fetcher should not include the book complete book URL but only the main site URL.

Example: 
BAD: https://bookname.textopress.com
GOOD: https://integrations.pressbooks.network/bookname

All network should give only the main site
https://bookname.textopress.com -> https://textopress.com
https://integrations.pressbooks.network/bookname -> https://integrations.pressbooks.network

solution
update pressbooks to submit only the main network domain. We can probably find out if the network uses subdomain or not as their book url. if yes, remove the book url part from the host. 

https://github.com/pressbooks/pressbooks/blob/04bd5de1f2a2fbb11042cc3296daf4db05526c29/inc/class-bookdirectory.php#L135

context
When textopress was set to display all public books and I manually excluded an individual book from textopress from the directory using the book's sharing & privacy setting, it was not immediately removed from the directory listing at dev.pressbooks.directory, nor was it removed during the next cron job/sync. See https://portuguese.textopress.com/wp/wp-admin/options-general.php?page=pressbooks_sharingandprivacy_options

SteelWagstaff:
@ho-man-chan would one of these functions get us the value you need? https://developer.wordpress.org/reference/functions/network_home_url/ or https://developer.wordpress.org/reference/functions/get_home_url/

ho-man-chan:
> @ho-man-chan would one of these functions get us the value we need? https://developer.wordpress.org/reference/functions/network_home_url/ or https://developer.wordpress.org/reference/functions/get_home_url/

Looks promising! Thanks @SteelWagstaff

Issue #2043, opened 2020-10-22
Reverting the network directory exclude only updates the public non catalog books


ho-man-chan:
Reverting the network directory exclude only updates the public non catalog books
solution
include all public books and not just catalog books during while reverting network book directory exclude
https://github.com/pressbooks/pressbooks/blob/b0d1e87dcf4af5ff9e30524e29297e7a98b44c52/inc/admin/network/class-sharingandprivacyoptions.php#L223
context
When I changed textopress so that it did not 'exclude uncatalogued books' from the directory, I would expect all public books with >1500 words that don't have 'test' etc in the title. to be added to the directory the next time the fetcher runs. This did not happen. There should be ~15 textopress books in the directory, but only 7 are listed (the ones that are in the catalog).

PR #2047, opened 2020-10-26
Fix bugs for book directory exclude (2043)

ho-man-chan:
Fixes the following issues:
https://github.com/pressbooks/pressbooks/issues/2043
https://github.com/pressbooks/pressbooks/issues/2044

Test:
1. Reverting the network book exclude will update all public books.
2. In textopress.com (or any network with subdomain as their book URL configuration), at the book level, excluding the book will delete it from the directory.


richard015ar (changes requested):
LGTM Ho man, I just leave a quick question for you.

richard015ar (approved):
LGTM

PR #2049, opened 2020-10-29
Create individual book exclude option if not exist

ho-man-chan:
Problem:
When individual books are excluded for the very first time, the pb_book_directory_exclude is added but does not trigger the function to exclude because it listens to an update of the option. 

solution: The previous implementation only triggers when the option is updated and not when added. This PR adds the option when viewing the sharing and privacy menu. By doing so, when updating the option, it will always trigger the exclude function since WordPress treats it as an updated and not an addition of an option.


SteelWagstaff (approved):
LGTM!

PR #2052, opened 2020-11-05
Replacing whitelist for allowlist in template

richard015ar:
Related issue: https://github.com/pressbooks/ideas/issues/323
Small change in template to replace "Whitelist" string.


SteelWagstaff (commented):
@richard015ar I added a commit which also updates the label.

SteelWagstaff (approved):
LGTM

Issue #2061, opened 2020-12-03
'Lost Password' page is broken when accessed directly from a book (instead of homepage)


pb-amy:
### Description

When someone decides to log-in or reset their password, they can do so by navigating from the network's home page or from the book site, both using the top menu bar.

From the home page, the "Lost Password" module works perfectly from the network's home page. ex: https://cod.pressbooks.pub/wp/wp-login.php?action=lostpassword

However, when you are on the book site, and you navigate to Log In -> Lost Password, the page is broken. Ex: https://cod.pressbooks.pub/**wp**/introductiontopsychology/wp-login.php?action=lostpassword

This happens on Aldine and the Publisher theme. The problem only appears on subdirectory installations (subdomains appear to be unaffected).

### Steps to Reproduce

1. Go onto a book site
2. Click on Log In on the top menu bar
3. Click on "Lost your Password?" in the login page.

**Expected behavior:** All 'Lost your Password?' pages should take you to working password reset pages.

**Actual behavior:** When 'Lost Password' module is accessed through the book site, the user is taken to a page which doesn't exist: `https://[institution].pressbooks.pub/**wp**/booktitle/wp-login.php?action=lostpassword` rather than `https://[institution].pressbooks.pub/booktitle/wp-login.php?action=lostpassword`

pb-amy:
Bug discovered on this ticket: https://pressbooks.zendesk.com/agent/tickets/3066

arzola:
@SteelWagstaff / @pb-amy 

We have two reset password valid URLs the "network one" and the "book one" 

**Network**

https://cod.pressbooks.pub/wp-login.php?action=lostpassword

**Book**

https://cod.pressbooks.pub/actionresearch/wp-login.php?action=lostpassword

Should we redirect to the "network one" ? as the expected behavior says? Those are slightly different logins, the network one it will take you to the general dashboard and the book one to the book administration directly

SteelWagstaff:
@arzola thanks for checking. When they attempt to reset their login from the book site login, we should send them to the book one. I.e. https://cod.pressbooks.pub/actionresearch/wp-login.php?action=lostpassword. It looks like the problem is that we're adding /wp/ to the path?

arzola:
Yes that is the problem I just asked because the expected behavior, but yeah thanks I will remove the Wp for this link I was thinking to use a filter for the reset password link but first I need to confirm the nginx rewrite rules

SteelWagstaff:
[...]

PR #2068, opened 2021-01-06
Added filter to remove unnecessary prefix

arzola:
Solves #2061 

Ready for PR @SteelWagstaff @richard015ar @ho-man-chan 

I think we should apply this unharmful fix and later work on the `wp/` Bedrock's stuff (for instance naming a book (wp), review the possibility to get rid of `/wp` entirely but this would require a full regression testing, this change is only focused on the lost password url

### DIFF
--- a/tests/test-login.php
+++ b/tests/test-login.php
@@ -7,17 +7,25 @@ class Login extends \WP_UnitTestCase {
 	 */
 	public function test_if_wp_prefix_is_removed() {
 
-		$scenario1 = 'https://institution.pressbooks.pub/wp/booktitle/wp-login.php?action=lostpassword';
-		$expected = 'https://institution.pressbooks.pub/booktitle/wp-login.php?action=lostpassword';
+		$scenario1 = 'https://network.pressbooks.pub/wp/booktitle/wp-login.php?action=lostpassword';
+		$expected = 'https://network.pressbooks.pub/booktitle/wp-login.php?action=lostpassword';
 
 		$actual = \Pressbooks\Registration\remove_wp_prefix($scenario1);
 
 		$this->assertEquals($expected, $actual);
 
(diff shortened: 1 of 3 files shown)


tests/test-login.php
```diff
@@ -0,0 +1,35 @@
...
+	 * @group Login
+	 */
+	public function test_if_wp_prefix_is_removed() {
+
+		$scenario1 = 'https://network.pressbooks.pub/wp/booktitle/wp-login.php?action=lostpassword';
+		$expected = 'https://network.pressbooks.pub/booktitle/wp-login.php?action=lostpassword';
+
+		$actual = \Pressbooks\Registration\remove_wp_prefix($scenario1);
+
+		$this->assertEquals($expected, $actual);
```
richard015ar, line 15: @arzola I think we could summarize it as:
```
$this->assertEquals( 'https://network.pressbooks.pub/wp/booktitle/wp-login.php?action=lostpassword', \Pressbooks\Registration\remove_wp_prefix( 'https://network.pressbooks.pub/wp/booktitle/wp-login.php?action=lostpassword' ) );
```
Same for other assertions.

richard015ar (changes requested):
@arzola thank you for applying this fix. The solution LGTM.
I left a single suggestion to simplify the test and avoid some extra variables.
I also noticed some small issues related to coding standard. I noticed we are not running phpcs in tests folder but would be good to keep the coding standards for new unit tests. 
You can run `vendor/bin/phpcs --standard=phpcs.ruleset.xml *.php tests/test-login.php` in the root path to see the coding standards issue.

richard015ar (dismissed):
LGTM! thank you for applying the suggestions.

richard015ar (approved):

Issue #2067, opened 2021-01-05
Update minimum required PHP version to 7.3


SteelWagstaff:
- [x] Update min version in readme, compatibility.php https://github.com/pressbooks/pressbooks/blob/d3275b9c903363cf66d1a344d75b2d7069c7f146/compatibility.php#L27-L36, etc.
- [x] Remove 7.1 and 7.2 from testing matrix https://github.com/pressbooks/pressbooks/blob/d3275b9c903363cf66d1a344d75b2d7069c7f146/.travis.yml#L17-L20

See https://github.com/pressbooks/pressbooks/commit/15f946b1a976bf0082d50bbcc33047f2a9be0679 & https://github.com/pressbooks/pressbooks/issues/1231 for last time this was done. Partial attempt from Dac in https://github.com/pressbooks/pressbooks/pull/1707/files + https://github.com/pressbooks/pressbooks/issues/1745

PR #2070, opened 2021-01-06
Update Pressbooks to PHP 7.3 and WordPress 5.5.3

SteelWagstaff:
This PR bumps the PB version to 5.17.4-dev and updates Pressbooks to require PHP 7.4 and WordPress 5.5.3 as minimum versions. It also removes previous versions of PHP from the test matrix used by Travis.

I used https://github.com/pressbooks/pressbooks/commit/15f946b1a976bf0082d50bbcc33047f2a9be0679 as a reference to understand what changes needed to be made.

Fix for https://github.com/pressbooks/pressbooks/issues/2067


readme.txt
```diff
@@ -1,9 +1,9 @@
 === Pressbooks ===
 Contributors: Pressbooks <code@pressbooks.com>
 Tags: ebooks, publishing, webbooks
-Requires at least: 5.4
-Tested up to: 5.5.2
-Requires PHP: 7.1
+Requires at least: 5.5.3
```
ho-man-chan, line 4: I believe we can set the requires at least to 5.5 so users don't get a nag if their wordpress is not up-to-date.

ho-man-chan (changes requested):
looks good. I think we can lower the request at least to 5.5 for the wp version.

readme.txt
SteelWagstaff: I'd prefer to set it to 5.5.3 as it's a stable release and the one we use in our production bedrocks.

ho-man-chan (approved):
LGTM

Issue #2066, opened 2020-12-29
Where can I report a 0-day Vulnerablity in this plugin?


OlivierLaflamme:
I reproduced a Pressbooks vulnerability or a misconfiguration from a client and  would like to disclose it if it's a zero-day bug.

Setup Pressbooks and recreation were compleated
Reproduce in 3.9.8
Reproduce in latest version

I would like to know were I have to go to to discuss this vulnerability.

SteelWagstaff:
Hi @OlivierLaflamme you're welcome to send details to our dev team at code@pressbooks.com. Thanks for checking with us -- we appreciate the courtesy and will attempt to review and remediate the issue in the new year.

OlivierLaflamme:
E-mail has been sent thank you @SteelWagstaff

Issue #404, opened 2016-05-01
Export formats


ausapa:
Hello
After updating Pressbook Textbook download formats options disappeared. I'm using the pluging with wordpress with the last version.
Does anybody know how to solve this issue?
Thanks so  Much
Aurelian

greatislander:
Hi @ausapa, Pressbooks Textbook is not supported here—please post your issues on their repository: https://github.com/BCCampus/pressbooks-textbook. (FYI @bdolor.)

PR #2072, opened 2021-01-12
Fix XSS security vulnerability

arzola:
Related issues #2066 and [#404](https://github.com/pressbooks/private/issues/404)

This security fix clean and sanitize metadata book info metaboxes to prevent XSS attacks on fields that allows HTML input, this uses [Htmlawed](https://www.bioinformatics.org/phplabware/internal_utilities/htmLawed/htmLawed_README.htm) to filter and sanitize the input values.

#### Requirements:

* Access to the latest Pressbooks admin interface

#### Steps:

* Enter to the admin interface
* Select any book and go to the book info page
* Try to fill any **wysiwyg** and **textareas** with malicious XSS strings and open the book front page of that book to see the safe sanitized values printed
* TIP you can use the following [XSS Filter Evasion Cheat Sheet](https://owasp.org/www-community/xss-filter-evasion-cheatsheet)
* You can try to test any input field metabox in the book info with the previous step, those values are already being properly escaped in the cover front page

### Notes

Two tests were added

[Functional test](https://github.com/pressbooks/pressbooks/pull/2072/files#diff-26db17a6755a6258d67c7522a3695bc626a34b87eadc4ded774722e22410ac9aR354)

[Unit test](https://github.com/pressbooks/pressbooks/pull/2072/files#diff-2cfd1565eaadffbe2d162d17535363549cd763bb62396d88845dcc9a2e9a96b9R573)

### DIFF
--- a/inc/sanitize/namespace.php
+++ b/inc/sanitize/namespace.php
@@ -6,6 +6,8 @@
 
 namespace Pressbooks\Sanitize;
 
+use Pressbooks\HtmLawed;
+
 /**
  * Convert HTML5 tags to XHTML11 divs.
  *
@@ -814,4 +816,27 @@ function maybe_safer_unserialize( $original ) {
 		return safer_unserialize( $original );
 	}
 	return $original;
+}
+
+
+/**
+ * Sanitize an string for undesired XSS attacks allowing HTML but removing malicious code
+ * @param $value
+ * @param bool $allow_html
+ * @return string
+ */
+function sanitize_string( $value, $allow_html = false ) {
+
+	if ( $allow_html ) {
+
+		return HtmLawed::filter(
+			pb_decode( $value ), [
+				'safe' => 1,
+			]
+		);
+
+	}
+
+	return strip_tags( pb_decode ( $value ) );
+
 }
--- a/inc/sanitize/namespace.php
+++ b/inc/sanitize/namespace.php
@@ -837,6 +837,6 @@ function sanitize_string( $value, $allow_html = false ) {
 
 	}
 
-	return strip_tags( pb_decode ( $value ) );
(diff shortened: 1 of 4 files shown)


SteelWagstaff:
@arzola will you add a reference to the issue addressed by this PR in the PR description as well as any information the reviewer/tester might need to review/test this PR? See https://github.com/pressbooks/pressbooks-lti-provider-1p3/pull/126 or https://github.com/pressbooks/pressbooks/pull/2070 for two previous examples.

arzola:
@SteelWagstaff @richard015ar I added the description I think now it's ready for review, thanks 👍

inc/sanitize/namespace.php
```diff
@@ -815,3 +817,26 @@ function maybe_safer_unserialize( $original ) {
...
+		return HtmLawed::filter(
+			pb_decode( $value ), [
+				'safe' => 1,
+			]
+		);
+
+	}
+
+	return strip_tags( pb_decode( $value ) );
+
```
richard015ar, line 841: Oscar, those 13 lines could be replaced by:
`return $allow_html ? HtmLawed::filter( pb_decode( $value ), [ 'safe' => 1 ] ) : strip_tags( pb_decode( $value ) );`

It is not mandatory, it is just a suggestion. The current implementation is cleaner that one-line solution in terms of readability but it still concentrate the single action (2 cases) simple to read.

If you decide to keep it in multiple lines, you can reduce it by deleting the empty lines to keep consistence between other functions in the same file.

arzola: I sent the improvement, I think for this case the ternary operator is fine 👍 thanks

richard015ar (approved):
LGTM! thank you Oscar!

PR #2083, opened 2021-01-29
fixed post type scope during callback

arzola:
I debug different scenarios of this "simple" change and this was causing a side effect in the importing routine because `$this->post_type` could be undefined some times and also the link could be lost because is being used as a filter callback and not as a regular php object

### Solution

I replaced the usage of `$this->post_type` inside the filter callback to do it in a more WP_Rest API way `$args['post_type']` and it works as expected both things :)

### Expected results

* Books are properly imported with glossary 
* Books could be cloned with full glossary
* Full Glossary is being exposed correctly through the API


richard015ar (approved):
Thank you for catching it Oscar! At the code level it LGTM. Happy to help testing it on Integrations

PR #2092, opened 2021-02-10
H5P Activities anchor

arzola:
Related Issue https://github.com/pressbooks/ideas/issues/356

This change adds a custom anchor name instead the default one #pb-interactive-content for h5p activities, with this change users will be able to link a specific activity from anywhere.

## Note

This change will require an upstream change into the h5p plugin I already sent a [PR](https://github.com/h5p/h5p-wordpress-plugin/pull/122)

### How to test it

1. Meanwhile the upstream change gets merged you should modify directly the plugin in your vendor directory in order to add the custom hook to append the desired anchor into the output, doing the changes the upstream [PR](https://github.com/h5p/h5p-wordpress-plugin/pull/122)

2. Create some h5p activities and export your book, see your new anchors and review if they link you directly to the activity
3. Remember you can't add the same h5p activities in the same chapter or page because both are going to have the same anchor because are the same thing.
4. All the anchors will follow the following structure: `#h5p-ID` where ID is the activity identifier.

### DIFF
--- a/inc/interactive/class-h5p.php
+++ b/inc/interactive/class-h5p.php
@@ -224,7 +224,7 @@ class H5P {
 	}
 
 	/**
-	 * This hook add a HTML wrapper to identify each hp5 activity
+	 * This hook adds a HTML wrapper to identify each hp5 activity
 	 * @param $html
 	 * @param $content array this array holds the custom post type information (h5p)
 	 * @return string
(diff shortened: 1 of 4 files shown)


SteelWagstaff:
@arzola the upstream PR has been merged and is included in H5P 1.15.2

arzola:
Ready to go

This PR is ready to be tested, remember you should bump the h5p plugin to have this working properly

Issue #2088, opened 2021-01-28
Links cannot be included in the Short & Long descriptions


pb-amy:
### Description
Found on the following Zendesk tickets: https://pressbooks.zendesk.com/agent/tickets/3909 (VT) and https://pressbooks.zendesk.com/agent/tickets/3976 (OpenOregon)

When adding a **Short or Long Description** in **Book Info**, you cannot add a link to any text; the `<a href="link">text</a>` will revert to  `<a href="">text</a>`. The hyperlink will still be there, but upon clicking on it, it will take you to the top of the page.

View what's happening on this video: https://drive.google.com/file/d/1gW95iEtR3lE0BHfvCNO7jfFcxrylU4eO/view

https://user-images.githubusercontent.com/61845107/106620824-f1dbf000-653f-11eb-9f12-b4602d197caa.mp4

### Steps to Reproduce

1. Create a hyperlink in the short / long description and save.
2. Go on the webbook, go down to the long description and click on it.
3. Look back in the book info and see that the link itself is gone.

**Expected behavior:** Hyperlinks should work, and should take you to the link.

**Actual behavior:** The text is linked but the field is empty.

richard015ar:
### Tests on integrations
Those are the following use cases I tried in both fields: Short and Long descriptions:
- `I am a 'description' with "a'" <a href="link.com">link</a>`
- `<a href="link.co">link'</a><a href="link.com">'link'</a>` -> Result: Both links works
- `<a href="link.com" target='_link'>link</a>` -> Result: link with target attribute works.
- `<a href='link.com' target="_link">link</a> <script>alert("hi");</script>` -> Result: it works, and single quotes are replaced by double quotes.
- `<div>I am div</div><a href="link.com" target="_link">link</a>` -> Result: it works.
- `<a fakeattr='asd'>link</a>` -> Result: link works with any href and extra attribute is deleted.

All cases works as expected. 
Thank you @arzola! it was a nice job!

PR #2099, opened 2021-02-25
Fixed anchor sanitization and the failing fonts test

arzola:
Related issue #2088 

### Problem

When links are being saved all the attributes quotes are being scaped this was not allowing`HtmLawed` to recognize a valid href schema.

### Solution

I added a `wp_kses_stripslashes` on the sanitization function that will only remove scaped quotes in html attributes that needs to be sanitized

Also two unit test assertions were added

### How to test

1. Try to save an anchor link into the fields allows HTML with some XSS strings
2. Wellformed urls will be saved in the href attribute other malicious tags should be striped

### DIFF
--- a/inc/admin/metaboxes/namespace.php
+++ b/inc/admin/metaboxes/namespace.php
@@ -435,7 +435,7 @@ function add_meta_boxes() {
 			'label' => __( 'Short Description', 'pressbooks' ),
 			'description' => __( 'A short paragraph about your book, for catalogs, reviewers etc. to quote.', 'pressbooks' ),
 			'sanitize_callback' => function ( ...$args ) {
-				return sanitize_string( $args[ METADATA_CALLBACK_INDEX ] );
+				return sanitize_string( $args[ METADATA_CALLBACK_INDEX ], true );
 			},
 		]
 	);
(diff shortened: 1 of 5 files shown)


richard015ar:
@arzola I tested this, and I had some problems by adding some links surrounded by `'` instead `"`.
For example: `<a href='https://google.com'>I am a link</a>`. It doesn't work as expected since it removes the `href` value completely.
Also, we need define if we will allow links in the short description as Amy mentioned in the ticket. If that the case, I guess you need to implement it in this PR.

There are a few names of variables that could be improved in the new tests you added, but it is not mandatory, just to keep your policy about naming conventions.

richard015ar (changes requested):
@arzola I tested this, and I had some problems by adding some links surrounded by `'` instead `"`.
For example: `<a href='https://google.com'>I am a link</a>`. It doesn't work as expected since it removes the `href` value completely.
Also, we need define if we will allow links in the short description as Amy mentioned in the ticket. If that the case, I guess you need to implement it in this PR.

There are a few names of variables that could be improved in the new tests you added, but it is not mandatory, just to keep your recommendations about naming conventions.

arzola:
@richard015ar hey bro, I fixed the issue with single quotes, just one thing the library is going to convert single quotes to double quotes.

I added support for HTML in the short description field.

For the naming conventions I was trying to follow the current "wordpress" standards snake_case according to https://github.com/pressbooks/coding-standards is there something that I'm doing wrong?

Thank you for your feedback

richard015ar:
LGTM! easy and elegant solution!
Thank you Oscar!

richard015ar (approved):

PR #2103, opened 2021-03-03
test: try GitHub actions for PHPUnit

greatislander:
This PR adds a simple Github Action for running PHPUnit tests. It does not change the current Travis CI configuration at all, but could be seen as a starting point for migrating to Github Actions for all CI.

Workflow results can be seen here: https://github.com/greatislander/pressbooks/actions/workflows/phpunit.yml

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -2,7 +2,7 @@ name: PHPUnit
 
 on:
   push:
-    branches: [ dev, master ]
+    branches: [ dev, master, github-actions ]
   pull_request:
     branches: [ dev ]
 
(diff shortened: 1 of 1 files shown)


SteelWagstaff:
@greatislander I think @ho-man-chan is still planning to review this PR this coming week. I have an slightly unrelated question for you. Have you tried the workflow without locking Composer at v1? https://github.com/greatislander/pressbooks/blob/b798cdbbc3d8efa334ad65564b8db88cf022f40c/.github/workflows/phpunit.yml#L48-L56. I'm trying to understand more about what was blocking us from updating to Composer 2 -- I think it was the composer/installer version declared in Travis, but am not sure. Would be curious to see if build/tests still pass.

greatislander:
> Have you tried the workflow without locking Composer at v1?

I didn't because I think it makes sense to move the CI to Github Actions as is and then work on the range of improvements we've discussed. It may work with an update to `composer/installers` but I think there's a trick there that involves not setting a strict version dependency for `composer/installers` (otherwise all the plugins and themes would need to be updated to the new dependency at once).

ho-man-chan (commented):
@greatislander Looks good!

There this cache message in your workflow, is of concern?
image

greatislander:
@ho-man-chan That will only be present if the composer dependencies haven't changed (which allows the cached dependencies to be used instead of running composer install). I think that's because something had changed prior to my run.

ho-man-chan:
> @ho-man-chan That will only be present if the composer dependencies haven't changed (which allows the cached dependencies to be used instead of running composer install). I think that's because something had changed prior to my run.

Hi Ned, so this is NOT an issue?

greatislander:
> Hi Ned, so this is NOT an issue?

Correct!

ho-man-chan (approved):
LGTM

Issue #2106, opened 2021-03-09
Import changes embedded base64 encoded images


BenGig:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [ ] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)? Pressbooks is up to date
* [x] Are you experiencing this issue with one of our built-in root or book themes?
* [x] Did you [perform a cursory search](https://github.com/issues?q=+is%3Aissue+repo%3Apressbooks%2Fpressbooks) to see if your bug or enhancement is already listed?

### Description

The import function scans for IMG tags. If an import contains embedded base64 encoded images, the import function treats them as broken and adds a "#fixme" at the end of the src attribute. In the case of base64, this leads to issues, particularly with Safari on iOS.

### Steps to Reproduce

1. Create an XML containing base64 encoded images
2. Import
3. Check in browser: base64 string ends with "#fixme". Safari on iOS shows no images.

**Expected behavior:** base64 encoded images should pass without being changed

**Actual behavior:** base64 string is changed

#### Root Blog Info

Root Blog ID: 1
Root Blog URL: http://pressbooks.benet.local/

#### Browser

Platform: Windows
Browser Name: Firefox
Browser Version: 78.0
User Agent String: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:78.0) Gecko/20100101 Firefox/78.0 Waterfox/78.8.0

#### WordPress Configuration

Network URL: http://pressbooks.benet.local/
Network Type: Subdirectory
Version: 5.6
Language: en_US
WP_ENV: Not set
WP_DEBUG: Disabled
Memory Limit: 64M

#### Pressbooks Configuration

Version: 5.18.2
Root Theme: Aldine
Root Theme Version: 1.9.0

#### Pressbooks Dependencies

Epubcheck: Not Installed
Kindlegen: Not Installed
xmllint: Not Installed
PrinceXML: Not Installed
Saxon-HE: Not Installed

#### Must-Use Plugins

hm-autoloader.php: n/a

#### Network Active Plugins

Hello Dolly: 1.7.2
Pressbooks: 5.18.2
WordPress Importer: 0.7

#### Root Blog Active Plugins

[...]

PR #2107, opened 2021-03-09
Don't process images if they are embedded base64 strings

BenGig:
refers to issue #2106

### DIFF
--- a/inc/modules/import/wordpress/class-downloads.php
+++ b/inc/modules/import/wordpress/class-downloads.php
@@ -67,7 +67,7 @@ class Downloads {
 			/** @var \DOMElement $image */
 			// Fetch image, change src
 			$src_old = $image->getAttribute( 'src' );
-			if ( str_contains($src_old, ";base64," )) {
+			if ( str_contains( $src_old, ';base64,' ) ) {
 				// Do nothing because image is embedded as base64 string
 				continue;
 			}
--- a/inc/modules/import/wordpress/class-downloads.php
+++ b/inc/modules/import/wordpress/class-downloads.php
@@ -67,6 +67,10 @@ class Downloads {
 			/** @var \DOMElement $image */
 			// Fetch image, change src
 			$src_old = $image->getAttribute( 'src' );
+			if ( str_contains($src_old, ";base64," )) {
+				// Do nothing because image is embedded as base64 string
+				continue;
+			}
 			$attachment_id = $this->fetchAndSaveUniqueImage( $src_old );
 			if ( $attachment_id === -1 ) {
 				// Do nothing because image is not hosted on the source Pb network


BenGig:
I fixed the code style issues in my changes. Sadly I cannot see where the style issues in the other file (not touched by me) come from.

SteelWagstaff:
Thanks @BenGig -- we're seeing the same test failures in another PR and think it may be a Travis CI issue unrelated to any changes you're proposing. Hopefully we'll get that squared away soon and will be able to review this properly before long.

SteelWagstaff:
@BenGig We've rerun the build -- all tests pass now, but we now have a CodeCov nag because no unit test was submitted with the code change. Do you have any experience using [PHPUnit](https://phpunit.de/) to write unit tests for PHP functions? If so, would you be willing to write a small test for the change you're proposing? See https://docs.pressbooks.org/unit-testing/. If not, our team should be able to assist at some point soon.

BenGig:
I have no experience with PHPUnit, but with (unit) testing as such, with other frameworks like pytest and molecule. Can you point me to the test unit which already does tests on this part of the code?

SteelWagstaff:
@BenGig All tests for the repo live in https://github.com/pressbooks/pressbooks/tree/dev/tests. You'd probably want to put unit tests for this new behavior in https://github.com/pressbooks/pressbooks/blob/dev/tests/test-modules-import.php? Upon my own cursory initial review of what you're proposing, I suspect that we'll have some concerns about accepting unvalidated base64 code into a Pressbooks import without stronger sanitization/validation than the simple `str_contains` check proposed here. See https://stackoverflow.com/questions/12658661/validating-base64-encoded-images. All this to say that it'll take a bit longer for us to review what probably appears to you to be a simple PR, as our devs and security team will need to take a closer look. You're welcome to use your proposed approach in your own local fork in the meantime?

SteelWagstaff:
@BenGig turns out I didn't properly understand what your PR was doing and gave some misleading answers previously. I don't think you need to do anything else at present re: security improvements or unit tests. I'll ask the dev team to take a closer look and see if we can get your PR reviewed properly soon. Sorry for all the back and forth on this.

arzola (changes requested):
@SteelWagstaff @richard015ar @fdalcin  

I tested it locally and works properly but creating a "unit" or feature test for this could be time consuming, we need to create a wordpress xml template that contains some base64 images embeded and then create the test using that file.

My thoughts

1. I think it's not a common thing to have images embeded like that and users usually don't import wordpress content using this method but still there are a few users that could be benefit of this

I suggest to ignore coverage for this block, and create a ticket to test the **class-downloads.php** file because this file is not being tested at all (just if this feature is highly used)

```
// @codeCoverageIgnoreStart
if ( str_contains( $src_old, ';base64,' ) ) {
continue;
}
// @codeCoverageIgnoreEnd
```

2. We could also discard this change if we think we won't want to support this change

SteelWagstaff:
@BenGig Closing this PR in favor of https://github.com/pressbooks/pressbooks/pull/2132, which has been merged to dev. Our next release of Pressbooks (early-mid next week) should include these changes.

PR #2108, opened 2021-03-10
Fixed linter issues

arzola:
Related issue https://github.com/pressbooks/pressbooks/pull/2095

This PR just fix a linter issue introduced by that change

PR #2114, opened 2021-03-12
Fix WP version for testing

arzola:
Wordpress 5.7 has some breaking changes in your test suite, we fixed the version to 5.6.2

> We've determined that the test failures are related to the WordPress 5.7 release -- we will fix temporarily by changing travis file to use WP 5.6.2 for test suite (the version of WordPress we use in production). Later we will seek to understand why tests break with WP 5.7 and address.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ branches:
 php:
 - 7.3
 env:
-- WP_VERSION=latest WP_MULTISITE=1 TRAVIS_NODE_VERSION="10"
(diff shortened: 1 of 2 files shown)


SteelWagstaff:
I'm going to run `composer fix` and commit change to branch. Should fix things for us and duplicates https://github.com/pressbooks/pressbooks/pull/2108

SteelWagstaff (approved):
Yay! Glad this passed.

PR #2117, opened 2021-03-17
Enable redirection plugin for super admins

arzola:
Related Issue [#436](https://github.com/pressbooks/private/issues/436)

### Solution

I added a filter hook that enable the redirection plugin only for super admin according to [official docs](https://redirection.me/developer/permissions/).


arzola:
Plugin for this created here

https://github.com/pressbooks/pressbooks-plugins-config

PR #2133, opened 2021-04-09
Fix google fonts path

arzola:
Google Fonts for Japanese, Korean and Chinese fonts changed the fonts directory structure

https://github.com/googlefonts/noto-cjk/pull/189

This PR fix the fonts urls


SteelWagstaff (approved):
LGTM!

PR #2138, opened 2021-04-14
Complete Github Actions Steps

arzola:
This PR includes full support for github Actions replacing travis


.github/workflows/phpunit.yml
```diff
@@ -42,42 +44,73 @@ jobs:
...
+          ${{github.workspace}}/*.zip
+
+    - name: Slack Notification
+      uses: rtCamp/action-slack-notify@v2
+      env:
+        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
+        SLACK_USERNAME: 'Github Actions'
+        SLACK_ICON_EMOJI: ':octocat:'
+        SLACK_COLOR: ${{job.status}}
+      if: always()
```
ho-man-chan, line 116: Hi @arzola ! After reviewing the slack messages sent by this slack action, it seems a little messy. It looks like our slack already has a GitHub integration as well, so this would be redundant. Good to know it works and it has potential for another task in the future! For now, I think it would be best to remove it. What do you think?

ho-man-chan (changes requested):
LGTM. Nice export! A note for the Slack integration:
After reviewing the slack messages sent by this slack action, it seems a little messy. It looks like our slack already has a GitHub integration as well, so this would be redundant. Good to know it works and it has potential for another task in the future! For now, I think it would be best to remove it. What do you think?

.github/workflows/phpunit.yml
arzola: Yeah I think that slack integration doesn't add value

arzola: I'll remove it

ho-man-chan (approved):
lgtm

PR #2139, opened 2021-04-16
fix: announce search results on catalog (resolves #1778)

### LINKED ISSUE #1778: No Announcement of Search Results for Non-Visual Users
### Steps to Reproduce

1. Navigate to My Catalog page in admin interface.
2. Navigate to search field, type query, execute search
3. Notice that the page is refreshed and focus is returned to the start of the page.
4. Notice that there is no aria-live or equivalent event to notify non visual users of the results of their search
5. Use the keyboard to navigate to the search results, notice that this takes a significant time, even when using an assistive technology like a screen reader.

**Expected behavior:** 
Assistive tech users should be given an update about the results of their search. Visual users can tell immediately if their search returned results while non-visual users have to browse through the entire page to see if the results table is populated. It is recommended that an aria-live region notifies users with a message like "Search returned 4 items" or "Search return no items".
**Actual behavior:** 
Currently non-visual users are given no notification and must navigate to the table of results to find out if their search was successful.

greatislander:
Resolves #1778. To test, perform a search on the catalog page while using a screen reader. Search results should be announced (screen reader behaviour is [inconsistent](https://github.com/w3c/aria/issues/1360#issuecomment-739319614), though— your mileage may vary).

See also: https://github.com/pressbooks/pressbooks/issues/1778#issuecomment-821294887

### DIFF
--- a/inc/admin/class-catalog-list-table.php
+++ b/inc/admin/class-catalog-list-table.php
@@ -535,7 +535,8 @@ class Catalog_List_Table extends \WP_List_Table {
 					/* translators: %s: search keywords, %d: total items found */
 					$search_results = sprintf( __( 'Search results for &#8220;%1$s&#8221; returned %2$d items', 'pressbooks' ), esc_html( wp_unslash( $_REQUEST['s'] ) ), $total_items );
 				}
-				echo '<span class="subtitle">' . $search_results . '</span>';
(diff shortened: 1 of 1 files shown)


SteelWagstaff (approved):
Results are ready by screen reading software following search execution as desired.

PR #2141, opened 2021-04-21
Update phpunit.yml

SteelWagstaff:
Add WordPress 5.7 to testing matrix

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -17,7 +17,7 @@ jobs:
         os: [ubuntu-18.04]
         epubcheck: [4.2.4]
         prince: [12.5-1]
-        wordpress: [5.6.2, 5.7.0]
+        wordpress: [5.6.2, 5.7]
 
     name: PHPUnit - ${{ matrix.php }}
 
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -19,7 +19,7 @@ jobs:
         prince: [12.5-1]
         wordpress: [5.6.2, 5.7]
 
-    name: PHPUnit - ${{ matrix.php }}
+    name: PHPUnit - ${{ matrix.php }} - ${{ matrix.wordpress }}
 
     steps:
     - name: Checkout code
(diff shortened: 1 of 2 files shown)


.github/workflows/phpunit.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         os: [ubuntu-18.04]
         epubcheck: [4.2.4]
         prince: [12.5-1]
-        wordpress: [5.6.2]
+        wordpress: [5.6.2, 5.7.0]
```
greatislander, line 20: @SteelWagstaff The reason this isn't working is that WordPress "major" versions aren't versioned using "real" SemVer (they omit the final `.0`). This should be:

```suggestion
        wordpress: [5.6.2, 5.7]
```

greatislander (changes requested):

.github/workflows/phpunit.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         os: [ubuntu-18.04]
         epubcheck: [4.2.4]
         prince: [12.5-1]
-        wordpress: [5.6.2]
+        wordpress: [5.6.2, 5.7]
 
     name: PHPUnit - ${{ matrix.php }}
```
greatislander, line 22: For clarity, it would be good to show the WordPress version here too so that the jobs are distinguishable in the checks list, e.g.:

```suggestion
    name: PHPUnit - ${{ matrix.php }} - ${{ matrix.wordpress }}
```

greatislander (changes requested):

arzola (approved):

fdalcin:
Wordpress 5.6 used to suppress some warnings, see [here](https://github.com/WordPress/WordPress/blob/8eea4444d91e641a664ffc024bf52bda9b92bf17/wp-admin/includes/image.php#L782). Wordpress 5.7 now relies on defining a `WP_RUN_CORE_TESTS` variable to suppress those warnings, see [here](https://github.com/WordPress/WordPress/blob/5ba41bd4f097c9bcd699736fc560686c0cad621e/wp-admin/includes/image.php#L796).

Running tests without defining `WP_RUN_CORE_TESTS` caused some tests to break (#2143). Adding the definition on `phpunit.xml` informs Wordpress to continue suppressing when running tests.

Also, as we run tests with code coverage we have to [define](https://core.trac.wordpress.org/ticket/50267#comment:4) the variable with a `false` value.

arzola:
LGTM, Thank you @fdalcin

PR #2145, opened 2021-04-23
ci: remove unnecessary conditional

greatislander:
This PR removes an invalid and unnecessary conditional check from the localization GitHub Action.

Follow-up to #2140.


SteelWagstaff (approved):
LGTM. Thanks Ned

PR #2161, opened 2021-05-19
Change label for LTI weblink export

richard015ar:
Change the label for CC LTI weblink export to a more user friendly name. Partial fix for https://github.com/pressbooks/pressbooks-lti-provider-1p3/issues/143


SteelWagstaff (approved):
LGTM. Thanks!

Issue #2162, opened 2021-05-20
Update labels for file export downloads


SteelWagstaff:
When book admins turn on file downloads to public visitors, a list of file formats is listed on the book's homepage. 
![Screenshot from 2021-05-20 12-57-10](https://user-images.githubusercontent.com/13485451/119040985-e84d8b00-b96a-11eb-849a-e725cb263a14.png)

Some label names are formatted inconsistently or confusing to most visitors. We should make the following changes:
'Print_pdf' -> 'Print PDF'
'Web links' -> Common Cartridge (web links)
'Thincc13' -> Common Cartridge 1.3 (LTI links)
'Thincc12' -> Common Cartridge 1.2 (LTI links)
'Thincc11' -> Common Cartridge 1.1 (LTI links)

See https://github.com/pressbooks/pressbooks-book/blob/dca488ac2ed41e1f9ef3eee48d280b95c85daeeb/partials/content-cover-book-header.php#L46-L58
 https://github.com/pressbooks/pressbooks/blob/67ecb4367319521460d62d8c2201f308ee61c0d0/inc/utility/namespace.php#L185-L212

https://github.com/pressbooks/pressbooks/blob/67ecb4367319521460d62d8c2201f308ee61c0d0/inc/modules/export/namespace.php#L194-L249

PR #2163, opened 2021-05-20
Fix display name for print PDF and CC with weblink exports

SteelWagstaff:
This PR fixes the display name for Print PDF and Common Cartridge with weblink exports as displayed on the book homepage when books are made available for download. It is a partial fix for https://github.com/pressbooks/pressbooks/issues/2162

It also adjust the test matrix to include WordPress 5.7.2 (the latest release of WordPress).

## To test:
1. checkout this branch
2. allow redistribution for books on your network at `NETWORKURL/wp-admin/network/settings.php?page=pressbooks_sharingandprivacy_options`
3. go to an individual book and turn on 'Share Latest Export Files' at `NETWORKURL/BOOKURL/wp-admin/options-general.php?page=pressbooks_sharingandprivacy_options`
4. Go to the exports menu in the book dashboard and produce file exports for PDF (for print) and Common Cartridge with Web Links  
5. Visit the book home page and confirm that you can see the updated labels and download the correct files

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -17,7 +17,7 @@ jobs:
         os: [ubuntu-18.04]
         epubcheck: [4.2.4]
         prince: [12.5-1]
-        wordpress: [5.6.2, 5.7, 5.7.1]
+        wordpress: [5.6.2, 5.7.2]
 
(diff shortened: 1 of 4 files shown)


richard015ar (approved):
LGTM, thank you Steel! now it looks much better to the user level!

PR #2170, opened 2021-05-26
Fix version and localizations

ho-man-chan:


SteelWagstaff (approved):
Looks good. Sorry I didn’t catch the version discrepancy earlier.

PR #2174, opened 2021-05-31
Generate Cover Thumbnail & Expose its URL to Books API Endpoint

arzola:
This PR is part of https://github.com/pressbooks/pressbooks-book-directory-fe/issues/269

This PR creates a new getCoverThumbnail function in `datacollector/class-book.php` which checks to see if a book has replaced the default cover image. If so, it returns the path for 'large' size thumbnail (max height/width of 1024px) of that image.

The URL of the cover image thumbnail is then exposed as part of the books API endpoint using the `thumbnailUrl` attribute 
recommended by Schema.org (https://schema.org/thumbnailUrl).

### How to test

1. Review the API endpoint has the new thumbnailUrl attr
2. Upload a new cover image and check to see if the thumbnail version is returned
3. Review if images with no cover images return the same default cover as the `image` attribute

### DIFF
--- a/inc/api/endpoints/controller/class-books.php
+++ b/inc/api/endpoints/controller/class-books.php
@@ -261,7 +261,9 @@ class Books extends \WP_REST_Controller {
 			'last_updated' => strtotime( get_blog_details( $id )->last_updated ),
 		];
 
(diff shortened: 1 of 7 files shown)


richard015ar (approved):
LGTM, I tested locally reviewing current images and uploading new ones.
Thank you Os!

Issue #2203, opened 2021-05-20
Store download event data in Pressbooks database when files are downloaded from webbook homepage


SteelWagstaff:
As a book author, I want to know how many times each of my book exports have been downloaded (by file type) from my book homepage without having to install Google Analytics or other tracking software so that I can understand how readers are interacting with my book and which file exports I should focus on improving.

As a network manager, I want to know which books have their export files downloaded most frequently, and to know which file types are most popular so that I can advise book admins on which file types users tend to download from our network.

As a product owner, I want to know which file types are most frequently downloaded across all networks so we can prioritize those export routines. I also want to know which files types are downloaded least often, so we can consider retiring or replacing them with more useful export options.

We currently have an event tracker built into Pressbooks-book which relays download information to Google Analytics: https://github.com/pressbooks/pressbooks-book/blob/dca488ac2ed41e1f9ef3eee48d280b95c85daeeb/partials/content-cover-book-header.php#L64-L74

I'd like us to store download events in the Pressbooks database so that it can be aggregated and displayed to book admins, network managers, and Pressbooks product team (for our hosted networks). Information we want to store in our database is the date of the action, the file type downloaded, and whether or not the person who downloaded the file was logged into Pressbooks at the time of download. We don't want to collect or store IP address or other identifying information about the user. Ultimately, our goal will be to count and display the number of file downloads by type over a given duration to book authors, network managers, and Pressbooks product team (for our hosted networks only).

Possible information to store in a new database table:
- is_logged_in
- date
- file type
- blog_id

Displaying this information is NOT in scope for this task. Collecting it in the Pressbooks database IS in scope for this task.

SteelWagstaff:
@ho-man-chan We've added a new feature which should add a new database table called 'wp_pressbooks_tracking' when book download events are performed from the webbook homepage (see https://github.com/pressbooks/pressbooks/pull/2204). Will you check in Integrations database to see if this table exists for Making Open Textbooks with Students (book id #3), and that it contains events that look like the ones in Ricardo's PR comment?

ho-man-chan:
@SteelWagstaff 
Looks Good!
Heres the result:
![image](https://user-images.githubusercontent.com/21694293/121703839-34d74280-caa1-11eb-9ab4-75f76fe63d6b.png)

PR #2204, opened 2021-06-08
Store download event data

fdalcin:
Issue: #2203 

This PR adds a new feature to record book download events. It will not record any sensitive information.
It creates a generic `pressbooks_tracking` table and adds a `BookDownload.php` class that is responsible for registering the download tracking action.

#### How to test

With the option to download books enabled
* Go to the book home page and download any file format available
* The download should work like before
* Check the database and see if there is a new table `wp_pressbooks_tracking`
* Check if the table has download records


inc/redirect/namespace.php
```diff
@@ -198,7 +198,11 @@ function rewrite_rules_for_sitemap() {
 function rewrite_rules_for_open() {
 
 	add_rewrite_endpoint( 'open', EP_ROOT );
-	add_filter( 'template_redirect', __NAMESPACE__ . '\do_open', 0 );
+	add_filter( 'template_redirect', function () {
+		do_open( function ( $filepath ) {
+			force_download( $filepath );
+		} );
+	}, 0 );
```
fdalcin, line 205: This was required to "mock" the download functionality. We now pass a callback to the `do_open` function and it will call it passing the file to be downloaded.

inc/redirect/namespace.php
```diff
@@ -208,7 +212,7 @@ function rewrite_rules_for_open() {
  * @copyright 2014 Brad Payne
  * @since 3.8.0
  */
-function do_open() {
+function do_open( $do_download = null ) {
```
fdalcin, line 215: Added a callback function to be able to swap download implementation for testing purposes.

inc/redirect/namespace.php
```diff
@@ -217,15 +221,17 @@ function do_open() {
...
-			if ( isset( $files[ $_GET['type'] ] ) ) {
-				$filepath = \Pressbooks\Modules\Export\Export::getExportFolder() . $files[ $_GET['type'] ];
-				force_download( $filepath );
-				exit;
-			}
+		$format = $_GET['type'];
+		$files = \Pressbooks\Utility\latest_exports();
+
+		if ( isset( $files[ $format ] ) ) {
+			do_action( 'store_download_data', $format );
```
fdalcin, line 230: Here we call the `store_download_data` action that will record that event.

inc/tracking/class-tracking.php
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * Set up the database table.
+	 *
+	 * @return void
+	 */
+	protected function setup() {
+		require_once( ABSPATH . 'wp-admin/includes/upgrade.php' );
+
+		$sql = "CREATE TABLE IF NOT EXISTS `$this->dbTable` (
```
fdalcin, line 34: Setup only created a new table if that doesn't already exists.

inc/redirect/namespace.php
```diff
@@ -217,15 +221,17 @@ function do_open() {
...
+
+			$filepath = \Pressbooks\Modules\Export\Export::getExportFolder() . $files[ $format ];
+			$do_download( $filepath );
```
richard015ar, line 233: Cara, should we ensure the callback function is not null?
`if ( ! is_null( $do_download ) )`, just in case we need to use this function and we don't send any callback function

fdalcin: Yes, we could add something like this ` $do_download && $do_download( $filepath );` maybe?

richard015ar: yes, makes sense to me!

inc/tracking/class-tracking.php
```diff
@@ -0,0 +1,76 @@
...
+				`id` bigint(20) NOT NULL AUTO_INCREMENT,
+				`blog_id` bigint(20) NOT NULL,
+				`track_type` varchar(30) NOT NULL,
+				`track_value` varchar(255),
```
richard015ar, line 50: I wonder if we want to store serialized data in the future for another tracking, makes more sense create this field as a `text` type. We are not sure at the moment, so keep it as varchar(255) is ok to me.

inc/tracking/class-tracking.php
```diff
@@ -0,0 +1,76 @@
...
+
+	/**
+	 * Set up the database table.
+	 *
+	 * @return void
+	 */
+	protected function setup() {
+		require_once( ABSPATH . 'wp-admin/includes/upgrade.php' );
+
+		$sql = "CREATE TABLE IF NOT EXISTS `$this->dbTable` (
```
richard015ar, line 46: I think we could store the timezone as well, since some networks is in different timezone. So, when we want to get data from different networks, we can process the data in a single timezone. I see in line 69 we are storing data in the server timezone (which makes sense), but timezone that indicates the in which timezone was the date stored would be good. What do you think?

richard015ar (changes requested):
@fdalcin the OOP design is really good! I just left a few notes with ideas, looking forward your feedback on those.

inc/tracking/class-tracking.php
fdalcin: Like you said, I think we can keep that as `varchar` for now and possibly change to `text` as needed.

inc/tracking/class-tracking.php
fdalcin: I see no problem in storing the timezone as well, just not sure about use cases for that field. Maybe we could ask @SteelWagstaff ?

richard015ar: I am agree, let's ask and we can decide after that.
I was thinking if we want to consume the data across networks in retool for example, if you see in the ticket Steel mentioned:
> As a product owner, I want to know which file types are most frequently downloaded across all networks

Which means if I want to see all books downloaded this week (or today) -for example- for all networks, could be an issue, since networks are in different timezone, the same issue we had with the lastUpdated filter feature in the PB Directory app. 
If we store the timezone, we will be able to get the info avoiding this issue by displacing the time with a centralized timezone (i.e UTC 0)

inc/tracking/class-tracking.php
```diff
@@ -0,0 +1,76 @@
...
+	 */
+	public function store( $value ) {
+		global $wpdb;
+
+		$wpdb->insert( $this->dbTable, [
+			'blog_id' => get_current_blog_id(),
+			'track_type' => $this->type,
+			'track_value' => $value,
+			'logged_in' => is_user_logged_in(),
+			'created_at' => date( 'Y-m-d h:i:s' ),
```
richard015ar, line 73: Cara, in case the timezone is an issue or required, option 2 could be store here all the tracking info in UTC 0, so we know all the data is stored in a single timezone independently of the network server setting.

inc/tracking/class-tracking.php
fdalcin: Solved by considering our [second option](https://github.com/pressbooks/pressbooks/pull/2204#discussion_r649183895)

inc/tracking/class-tracking.php
fdalcin: Applied changes to enforce UTC time when storing dates.

inc/tracking/class-tracking.php
```diff
@@ -0,0 +1,78 @@
...
+		global $wpdb;
+
+		$date = new \DateTime( 'now', new \DateTimeZone( 'UTC' ) );
+
+		$wpdb->insert( $this->dbTable, [
+			'blog_id' => get_current_blog_id(),
+			'track_type' => $this->type,
+			'track_value' => $value,
+			'logged_in' => is_user_logged_in(),
+			'created_at' => $date->format( 'Y-m-d h:i:s' ),
```
richard015ar, line 75: @fdalcin you should add `H` instead 'h', otherwhise it will contains 0-12 for hours. It should be: `$date->format( 'Y-m-d H:i:s' )`

I just tested and here's the result in my db table:
```sql
mysql> select * from wp_pressbooks_tracking;
+----+---------+---------------+-------------+-----------+---------------------+
| id | blog_id | track_type    | track_value | logged_in | created_at          |
+----+---------+---------------+-------------+-----------+---------------------+
|  8 |       2 | book_download | epub        |         1 | 2021-06-10 03:00:46 |
|  9 |       2 | book_download | wxr         |         1 | 2021-06-10 03:00:49 |
| 10 |       2 | book_download | thincc12    |         0 | 2021-06-10 03:00:55 |
| 11 |       2 | book_download | thincc11    |         0 | 2021-06-10 03:01:21 |
| 12 |       2 | book_download | xhtml       |         0 | 2021-06-10 03:01:23 |
+----+---------+---------------+-------------+-----------+---------------------+

```
If you see now it 3 PM ish in UTC timezone, it's 15:00 ish in mysql datetime format

fdalcin: My bad. Fixed now.

richard015ar:
## Test notes
- I allow downloads in a test public book in my local environment
- I create different file formats ready to download from the main book's page
- I downloaded all available files logged in and logged out
- The DB Table contains all the info required:
```sql
mysql> select * from wp_pressbooks_tracking;
+----+---------+---------------+-------------+-----------+---------------------+
| id | blog_id | track_type    | track_value | logged_in | created_at          |
+----+---------+---------------+-------------+-----------+---------------------+
| 20 |       2 | book_download | htmlbook    |         1 | 2021-06-10 15:34:38 |
| 21 |       2 | book_download | xhtml       |         1 | 2021-06-10 15:34:39 |
| 22 |       2 | book_download | wxr         |         1 | 2021-06-10 15:34:40 |
| 23 |       2 | book_download | vanillawxr  |         0 | 2021-06-10 15:34:46 |
| 24 |       2 | book_download | thincc12    |         0 | 2021-06-10 15:34:47 |
| 25 |       2 | book_download | thincc11    |         0 | 2021-06-10 15:34:48 |
+----+---------+---------------+-------------+-----------+---------------------+

```
- I run the unit tests created and the phpcs to check linting issues

Everything works as expected, and the OOP design is flexible to add new tracking data pretty easy in the future if needed. Thank you @fdalcin, it's a great feature!

richard015ar (approved):

PR #2214, opened 2021-06-14
Use pb_cover_large on API for thumbnails

arzola:
This PR will make our directory super fast, because we are going to load really optimized thumbnails

### DIFF
--- a/inc/datacollector/class-book.php
+++ b/inc/datacollector/class-book.php
@@ -471,7 +471,7 @@ class Book {
 		$cover_id = $attachment_id ? $attachment_id : attachment_id_from_url( $cover_path );
 
 		if ( $cover_id ) {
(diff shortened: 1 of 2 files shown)


SteelWagstaff (commented):
@arzola These changes look good for me locally. I noticed that we're not touching the default cover image, however. It appears that we do have a few sizes of the default book cover in our dist/images folder. How difficult would it be to reference one of these (like https://github.com/pressbooks/pressbooks/blob/dev/assets/dist/images/default-book-cover-225x0%402x.jpg or https://github.com/pressbooks/pressbooks/blob/dev/assets/dist/images/default-book-cover-225x0.jpg) as the thumbnail URL when a book hasn't change it's cover image?

SteelWagstaff (approved):
Approved. We'll love at improving behaviour for default cover image thumbnails once we have a replacement cover image to use.

PR #2219, opened 2021-06-17
Suppress mobi 8 generation for kindlegen

ho-man-chan:
**Description**
This Pull Request suppresses the .mobi8 generation for Kindlegen in the site's export folder by removing the file after generation in Ubuntu 20.04.

To suppress the .mobi8 file generation define php constant PB_KINDLEGEN_SUPRRESS_MOBI8 to true.
By default PB_KINDLEGEN_SUPRRESS_MOBI8 is false.

**Test**
Export a book in Mobi format in an OS Ubuntu 20.04 with Pressbooks

**Result**
Only Mobi file is generated and Mobi8 does not exist.


SteelWagstaff:
Closing in favor of #2222

PR #2224, opened 2021-06-22

ho-man-chan:


SteelWagstaff (changes requested):
@ho-man-chan this all looks good to me, but checks fail because of missing `sudo apt update` line. Will you add to this PR and we'll merge it here?

SteelWagstaff (approved):
:+1:

PR #2226, opened 2021-06-23
Replace default book cover images

### LINKED ISSUE #2195: Change the default book cover image
### Feature Description

Please switch the [default book cover](https://github.com/pressbooks/pressbooks/blob/dev/assets/dist/images/default-book-cover.jpg) (the red type writer and the word Pressbooks) to an updated version that does not feature a type writer. This is a bandaid solution until we can do something more dynamic (e.g. a cover generator, Cesar once made one for Rebus that was super cool). 

### Feature Use Case

Now that the search results display sorted by most recently updated, a lot more red type writers are showing up. The red type writers conflict with our innovative brand and make the search results look less impressive than they are.

### Other Notes

One option for the cover would be to have the same graphic designer [Leah Concepcion-Vanderbyl](https://www.leahcv.com/) who did the website and some of the assets do a standard book cover. I would be happy to facilitate the conversation.

SteelWagstaff:
This PR replaces the default cover image (red typewriter) with a new default cover image which matches the visual elements that were part of our recent rebrand. It fixes #2195. 

One note: these images are 2x larger than previous ones. Partially I think this is because the jpg has more information in it, but may also be because I don't have a copy of Adobe Illustrator to make web optimized jpg versions of these files from the .ai source file. If someone does have Illustrator and would like to do this to optimize image size, let me know ... Image can be found here: https://drive.google.com/file/d/11jyFqP3G2bHMqbXlrr0CQq_-RI_B6toB/view?usp=sharing

## To Test
1. Checkout this branch locally
2. Look at a previously created book with a default cover image. Check to see that the image has been replaced with the new image
3. Create a new book. Check to see that its cover image it the new default image


arzola (approved):
LGTM

I tested it and those images works properly I just optimized the images to strip any EXIF metadata and also reduce the file size apply the google recommended `jpgoptim` settings

Issue #2154, opened 2021-05-10
Ensure that cover image URLs are always https (instead of http) in API


SteelWagstaff:
## Description

The URL for the cover image provided via Pressbooks API for some books use http instead of https. We should use https whereever possible.

**Expected behavior:**
Visit https://uw.pressbooks.pub/wp-json/pressbooks/v2/books?page=3 and look at URLs in `image:` field for various books. All of them should be https. 

**Actual behavior:**
Instead, several of them use http. This causes problems in the Directory and the marketing site, which reads the book cover link to display cover images on each of these sites.

## Notes
See https://github.com/pressbooks/pressbooks-book-directory-fe/issues/180 and https://github.com/pressbooks/pressbooks-book-directory-fe/issues/269 for related issues.

We'd like network URL, site URL, home address, and image values to have is_ssl checks and then store the https value if true.

See https://github.com/pressbooks/private/issues/100

hughmcguire:
The marketing site is pulling from Directory/Staging fyi.

SteelWagstaff:
We've determined that the 'Network URL' for uw.pressbooks.pub and ecampus.pressbooks.pub are both set to http:// instead of https://. There may be other networks where this is the case as well. We need to investigate further.

LeighKP:
@SteelWagstaff I'm meeting with Kayla today. Is there anything more for Plank to be concerned with here, or is it on us now?

SteelWagstaff:
Fixed on integrations and textopress

PR #2230, opened 2021-06-28
Pb 2154 Use https when available on API books endpoint resources

arzola:
This PR solves #2154 

This enforces https for book urls and images exposed through the API

## How to test

Create a new book on your network, change the site_url and home_url in your wp_*_options to http://, review the book on the API and should have https on the link and image attributes.

I'm not sure about hooking the `wp_insert_site` I couldn't find any modifications inside the default multisite blog (book) creation, but if we want to proceed we could use that hook to update `site_url` and `home_url` id those values don't have https and the main has `ssl` enabled.

### DIFF
--- a/inc/api/endpoints/controller/class-books.php
+++ b/inc/api/endpoints/controller/class-books.php
@@ -2,10 +2,10 @@
 
 namespace Pressbooks\Api\Endpoints\Controller;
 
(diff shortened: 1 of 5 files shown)


fdalcin (approved):
LGMT @arzola, tested and works as expected

image

image

image

Issue #2233, opened 2021-06-30
Fix cannot modify header issue with book download tracking


SteelWagstaff:
This warning in New Relic is related to the new downloads feature. It is triggered every time someone downloads the book

![image (8).png](https://images.zenhubusercontent.com/6012398f80b7cb0f8d1385bc/9f1732e5-a739-43cf-9428-6fcffedc10d2)

Maybe it's related to: https://github.com/pressbooks/pressbooks/blob/1457ab067706bf0ccb92fa3560a65ecac0cb010b/inc/redirect/namespace.php#L234

Value was previously `exit;`

SteelWagstaff:
Tested. Works as expected on Integrations.

PR #2237, opened 2021-07-05
Fix Cannot modify header warning

fdalcin:
Fix for #2233 

This PR removes the "Cannot modify header" warning when downloading books.

### How to test

1. Perform any book download
1. The file `/srv/www/pressbooks.test/logs/error.log` shouldn't have "Cannot modify header information" warnings.


SteelWagstaff (approved):
Lgtm

PR #2241, opened 2021-07-10
chore: add Dependabot configuration file (fix #2236)

### LINKED ISSUE #2236: Set Dependabot allow rules to direct only
Make sure dependabot rules are set to top level only, so that package dependencies are handled in build-tools repo not core repos: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#allow

Once merged, use this as model for other repos.

greatislander:
This PR adds a configuration file for Dependabot, specifying that only direct (top-level) dependencies should be checked for updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+version: 2
+updates:
+  # Maintain dependencies for npm
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


SteelWagstaff (approved):
Approved!

PR #2258, opened 2021-07-14
Break line after Contact link

fdalcin:
This PR adds a line break after the Pressbooks footer text, so that it will display more cleanly when other plugins (like Koko Analytics) append additional text in this area.


SteelWagstaff (approved):
LGTM

Issue #2193, opened 2021-03-13
Remove MOBI exports from Pressbooks


SteelWagstaff:
TODO:
- [x] Remove MOBI export option from export menu, all references to MOBI export routines and tests (including MOBI hacks from EPUB: https://github.com/pressbooks/pressbooks/issues/1859), and references to KindleGen. 

    - https://github.com/pressbooks/pressbooks/pull/2260

- [ ] Remove KindleGen dependencies from all active infra repos. 

    - See https://github.com/pressbooks/textopress-infrastructure/pull/22 
    - https://github.com/pressbooks/trellis/pull/35 
    - https://github.com/pressbooks/ca-central-1-infrastructure/pull/126 
    - https://github.com/pressbooks/us-east-1-infrastructure/pull/3 
    - https://github.com/pressbooks/pressbooks-infrastructure/pull/12 
    - https://github.com/pressbooks/bibliolabs-infrastructure/pull/9 
    - https://github.com/pressbooks/ecampus-infrastructure/pull/6 
    - https://github.com/pressbooks/client-infrastructure/pull/17
    - https://github.com/pressbooks/unizin-infrastructure/pull/25 
    - https://github.com/pressbooks/unizin-b-infrastructure/pull/4 
    - https://github.com/pressbooks/sample-infrastructure/pull/1 
    - https://github.com/pressbooks/rebus-infrastructure/pull/2 
    - https://github.com/pressbooks/audit-infrastructure/pull/1

- [ ] Update documentation as needed

Amazon has retired the KindleGen tool that Pressbooks has historically used to generate MOBI exports. See [note](https://www.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211): 
> KindleGen is no longer available for download. Please use Kindle Previewer to convert, preview, and validate your eBooks. Kindle Previewer provides the same functionality of KindleGen. ... We recommend using EPUB format for publishing new reflowable titles and updating previously published titles. MOBI should only be used when testing on older devices that do not support Enhanced Typesetting.

From [Amazon's KDP recommendations for publishers](https://kdp.amazon.com/en_US/help/topic/G79CTKR8BX79E96L):
> Note: We recommend submitting EPUB files for publishing reflowable content. MOBI is not recommended for reflowable content. Kindle Previewer is the best way to validate your EPUB content before publishing. 

For conversion houses, they note: 
[...]

Issue #1859, opened 2019-11-27
Mobi hacks XSLT breaks JavaScript in Epub3


SteelWagstaff:
Old Asana ticket from @connerbw :

Played around with trying to get JavaScript working in EPUB3 exports and found out this:
`\Pressbooks\Modules\Export\Epub\Epub201::kneadHtml`
`$html = $this->transformXML( ""{$utf8_hack }<html>{$html}</html>"", $this->dir . '/templates/epub201/mobi-hacks.xsl' );`
Will treat CDATA as a text node.

> The XPath data model that XSLT uses doesn't allow to distinguish any CDATA sections -- any of these are represented just as (part of) a text node. So CDATA preservation cannot be achieved with XSLT or XPath  [...]

Got it working by 
1. Fixing this [ https://github.com/Masterminds/html5-php/issues/139 ] with an unpleasant hack, ie.
```
\Pressbooks\Modules\Export\Epub\Epub201::kneadHtml
$html = $html5parser->saveHTML( $dom );	
$html = \Pressbooks\Sanitize\strip_container_tags( $html );		
$html = \Pressbooks\HtmLawed::filter( $html, [ 'valid_xhtml' => 1 ] ); // HTML5 back to XHTML 1.1 strict. again
```
2. Using cdata-section-elements for script tags
```
// web/app/plugins/pressbooks/inc/modules/export/epub/templates/epub201/mobi-hacks.xsl
<xsl:output encoding=""UTF-8"" indent=""yes"" method=""xml"" omit-xml-declaration=""yes"" cdata-section-elements=""script""/>
```
The whole mobi7 hacks using XSLT should be abandoned.

SteelWagstaff:
Dac explained this to me in 1-on-1 meeting. My understanding is that we include a bunch of extra XSLT info in epub exports so that we can use later in producing a mobi export. This cruft causes problems for EPUB3 exports. Ideally, we'd only include these hacks if/when they were needed in mobi export routine OR decide that they're only needed by old mobi readers that are no longer in support (like how we've decided to no longer ensure our html output works in IE 8, for example).

PR #2260, opened 2021-07-16
Remove MOBI export and references

SteelWagstaff:
This PR removes the MOBI export option and most MOBI related references from Pressbooks. Partial fix for https://github.com/pressbooks/pressbooks/issues/2193 and fix for https://github.com/pressbooks/pressbooks/issues/1859. 

It also makes a small change to the export format menu, moving 
PressbooksXML and Thin Common Cartridge with weblinks to the 'supported formats' column and renaming the EPUB 2.01 export option.

Screenshot from 2021-07-15 19-38-14

**To test:**
1. Turn on ability to distribute files from home page
1. Open a book and create a MOBI export
1. Create an EPUB export and confirm that it does not have validation errors
1. Switch to this branch and open the export menu in any book
1. Check to see that MOBI export no longer appears in the menu
1. Check to see that the previously created MOBI file still appears in the list of previously exported files & can still be downloaded from the book's home page.
1. Confirm that EPUB 2.01 exports do appear in the menu 
1. Create an EPUB export and confirm that it still does not have validation errors


fdalcin (commented):
@SteelWagstaff LGTM. My only question is, should we also remove styling related to MOBI ([see here](https://github.com/pressbooks/pressbooks/blob/dev/assets/src/styles/export.scss#L127-L133))?

Confirming that MOBI option isn't present anymore
image

MOBI previously generated is still available
Screen Shot 2021-07-19 at 14 36 34

Still possible to download MOBI files previously generated 
Screen Shot 2021-07-19 at 14 37 25

SteelWagstaff:
@fdalcin those images are displayed on the export page after exports have been produced 
Screenshot from 2021-07-19 11-03-05. i think we should keep it so the display page works as expected

fdalcin:
Makes sense, I thought they were related to the exporting process. My bad.

fdalcin (approved):

SteelWagstaff:
@greatislander are you willing to take a quick look at this before we merge to make sure we're not missing anything?

PR #2265, opened 2021-07-20
Pb 2263 trim book info

### LINKED ISSUE #2263: Trim book info information before save
Occasionally, book info values will be stored with extra spaces before after a value. 

This can causes when these fields need to be sorted in alphabetical order, as in the Pressbooks Directory. For example, when the title of a book has extra spaces before the title begins, the book will appear too early in alphabetical order. We should trim/sanitize these values upon save in Pressbooks 'book info'.

SteelWagstaff:
This PR fixes #2263 by sanitizing the text entry fields for several of the book info metadata fields (including book title).

To test:
1. Open the book info page for a book
2. Enter values text entry fields which include superfluous spaces at the beginning or end of the string, html tags, `<` characters or other characters removed by the WP sanitize text field function: https://developer.wordpress.org/reference/functions/sanitize_text_field/
3. Add extra space at the beginning and end of the copyright notice, short and long description fields
4. Save book info and observe that extra white space at beginning and end of string has been removed on save for all fields
5. Observe that HTML tags and other characters have been sanitized or escaped upon save for all text entry fields except copyright notice, short and long description fields (which permit HTML).

### DIFF
--- a/inc/admin/metaboxes/namespace.php
+++ b/inc/admin/metaboxes/namespace.php
@@ -190,6 +190,9 @@ function add_meta_boxes() {
 		'pb_title', 'metadata', [
 			'group' => 'general-book-information',
 			'label' => __( 'Title', 'pressbooks' ),
+			'sanitize_callback' => function ( ...$args ) {
+				return sanitize_text_field( $args[ METADATA_CALLBACK_INDEX ] );
+			},
 		]
 	);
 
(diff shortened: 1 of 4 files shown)


arzola (approved):
LGTM Thanks @SteelWagstaff I added tests to test the sanitization callbacks

Issue #2264, opened 2021-07-19
Print PDF icon image missing on export page


SteelWagstaff:
### Description

The print PDF icon image is missing on the export page. 

### Steps to Reproduce

1. Produce a print PDF export.
2. Look at the file listing on the export page

**Expected behavior:** 
I expect to see an icon that says printPDF

**Actual behavior:** 
There is no icon for the print PDF export.

![Screenshot from 2021-07-19 11-06-32](https://user-images.githubusercontent.com/13485451/126206542-e55fd076-9560-4295-b2e2-4e6de22eed6f.png)

gavinpressbooks:
Test on Integrations with the following book: https://integrations.pressbooks.network/usaskphystest

Works as expected. Red banner 'Print PDF' icon is now visible for print PDF exports.

See also: #2266

Issue #2266, opened 2019-11-20
Add "Print" to "Print PDF" icon


SteelWagstaff:
When we generate PDF exports, the icons used for digital and print PDFs are visually disambiguated by color only (red & greeen). 
![PDF_export](https://user-images.githubusercontent.com/13485451/69245864-a412dc00-0b75-11ea-9ae9-f9d30b2fc692.png)

We should improve this by adding the word 'print' to the print PDF export icon.

The code for the current print PDF icon can be found here: https://github.com/pressbooks/pressbooks/blob/8da7f180f8e6187b82d6c64d49f6c8aa339c5a8c/assets/src/styles/export.scss#L135-L141. I don't know how these PNGs were originally generated (I think that was @greatislander's work?), but there are four sizes of this icon here: https://github.com/pressbooks/pressbooks/tree/master/assets/src/images

gavinpressbooks:
Test on Integrations with the following book: https://integrations.pressbooks.network/usaskphystest

Works as expected. Red banner 'Print PDF' icon is now visible for print PDF exports.

gavinpressbooks:
Test on Integrations with the following book: https://integrations.pressbooks.network/usaskphystest

Works as expected. Red banner 'Print PDF' icon is now visible for print PDF exports.

See also: https://github.com/pressbooks/pressbooks/issues/2266

PR #2267, opened 2021-07-20
Replace print pdf icon

SteelWagstaff:
This PR fixes https://github.com/pressbooks/pressbooks/issues/2264 and https://github.com/pressbooks/pressbooks/issues/2266

To test:
1. Produce a Print PDF export
2. Check to see that the new icon appears on the export page. The icon should have a red banner and should include the text 'Print PDF'

**Note:** I did not replace the 36 and 72 px versions of this image as they are not used anywhere in Pressbooks. The large size is explicitly specified for the export table, which is the only place these icons are currently used in Pressbooks: https://github.com/pressbooks/pressbooks/blob/86b921b5c3ede76e3fd94732e56cce50c710ded8/inc/modules/export/class-table.php#L414-L416


SteelWagstaff:
@greatislander Is there a historical reason why we need to leave the small sizes of export icons SCSS https://github.com/pressbooks/pressbooks/blob/86b921b5c3ede76e3fd94732e56cce50c710ded8/assets/src/styles/export.scss#L96-L198 and the corresponding image assets in Pressbooks?

fdalcin (dismissed):
LGTM @SteelWagstaff 

Screen Shot 2021-07-21 at 10 04 56

greatislander:
> @greatislander Is there a historical reason why we need to leave the small sizes of export icons SCSS and the corresponding image assets in Pressbooks?

@SteelWagstaff Nope. It used to be that the exports were displayed in a grid, with the most recent of each format as large icons and then a more compact grid with smaller icons below. These can and should be removed.

SteelWagstaff:
@fdalcin Just pushed a commit which removes the unused small sizes of file export icons. If it still looks good to you, please squash and merge at your convenience.

PR #2271, opened 2021-07-26
Escape special characters for alt image attribute

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/1918

When an image is upload to a chapter, part or other taxonomies with alternative text containing unsupported html characters, it breaks the XHTML export formats affecting PDF export formats.

This change checks the `alt` and `title` attributes for images and applies the `htmlspecialchars` PHP function to convert those special characters.

### Test notes
- Add an image to a chapter, and add a unsupported character as alternative text for that image, for example: `I am > an alternative text`.
- Export the book to PDF and confirm the book looks as expected


SteelWagstaff (commented):
@richard015ar we also need to escape HTML characters in the 'Title' attribute as well, so I added that and renamed the function. The biggest question I have about this change is whether it wouldn't make sense to try to combine this function with the fixImages function which already exists in this class. I tried to do something like that in those old PR: https://github.com/pressbooks/pressbooks/pull/1969/files but could never quite get it working correctly ...

SteelWagstaff (approved):
LGTM!

Issue #2096, opened 2021-02-16
[Hackathon] Surface more information about contributors in the book and book homepage


SteelWagstaff:
Will hold a 2 1/2 day hackathon from August 2-4 to focus on the following:

**Feature description:**
DONE:
- [x] Change contributor page to allow user entry (and sanitize inputs) for Name, square profile picture (upload image or choose from media library) with size contraints, text bio (rich text input -- allows links, italics, bold, but not other html tags for now), Website (URL input), Twitter (URL input), LinkedIn (URL input), GitHub (URL input). Any text that has been already entered in to the current 'description' field provided for contributors should be preserved and moved to the text bio field, if possible.
- [x] Change user profile page to match inputs described above. Remove unneccesary fields. https://github.com/pressbooks/pressbooks/pull/2295
- [x] When an existing user is added to a book with the role of (or has their role increased to that of) contributor or higher, make sure a 'contributor' entry is created for them and matches at time of creation. 
- [x] Create an 'about the contributor' stub/page for each contributor which displays the info provided above in attractive way

TODO:
- [ ] #2312
- [ ] #2311
- [ ] https://github.com/pressbooks/pressbooks-book/issues/832
- [ ] #2314
- [ ] https://github.com/pressbooks/pressbooks-book/issues/831
- [ ] Allow users to associate their Pressbooks profile with an ORCID ID (stretch goal): https://github.com/pressbooks/pressbooks/issues/1500

Context: https://docs.google.com/document/d/1b-YsRqRERAEsz3hwNWAT24CGgt-lL84Ee5BQ3PsEp9s/edit#bookmark=id.85aqxibgch6s

hughmcguire:
* generate a "Contributor" page in the book that is a styled list of all contributors & their info.

NMSalmon:
I'm very excited for this revamp! Two related questions/ideas, if they're welcome here (otherwise please ignore!):

1.  Would this affect the way that the dashboard view of the contributors feature operates? (Currently, my "contributors" list will display the number of pages affiliated with each author, but if you click on that count hyperlink, it just takes you to a blank page that says "0 posts" rather than showing you who authored which chapter.)

[...]

PR #2295, opened 2021-08-02
Remove unused contact fields and add new ones to user profile page

arzola:
This PR removes unwanted contact info fields from the user profile page (AIM, Yahoo IM, Jabber / Google Talk) and adds three new contact fields: Twitter URL, LinkedIn URL, and GitHub URL. It also checks to ensure that the values entered for these fields and the existing website field are valid URLs. If not, the page throws an error and does not save any changes.

Partial fix for https://github.com/pressbooks/pressbooks/issues/2096

**To Test:**
1. Checkout this branch and attempt to edit your user page
2. Observe that the previous contact fields are no longer visible and the three new fields are visible
3. Enter invalid text in the website field and save. Observe an error message and unchanged values
4. Enter a valid URL in the website field and save. Observe success!
5. Repeat 3 & 4 for each of the other new fields.
6. Login as an admin user and repeat steps 1-5 while editing the user profile for a user that is not you.

### DIFF
--- a/hooks-admin.php
+++ b/hooks-admin.php
@@ -359,3 +359,6 @@ add_action( 'admin_init', '\Pressbooks\Theme\check_upgraded_customcss' );
 // Bulk add users
 add_action( 'init', [ '\Pressbooks\Admin\Users\UserBulk', 'init' ] );
 
+// Add & sanitize additional contact methods to user profile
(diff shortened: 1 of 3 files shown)


SteelWagstaff (approved):
LGTM

PR #2298, opened 2021-08-04
ci: set wp-cli package index to canonical = false

greatislander:
This should fix the install issues with the POT update GitHub Action.


SteelWagstaff (approved):
LGTM

PR #2302, opened 2021-08-06
Include institution

arzola:
This PR Is part of #2096


inc/class-taxonomy.php
```diff
@@ -56,6 +56,8 @@ static public function hooks( Taxonomy $obj ) {
 			add_action( 'init', [ $obj, 'registerTaxonomies' ] );
 			add_action( 'init', [ $obj, 'maybeUpgrade' ], 1000 );
 			add_action( 'user_register', [ $obj->contributors, 'addBlogUser' ] );
+			add_action( 'add_user_to_blog', [ $obj, 'registerTaxonomies' ]); // This is a workaround because newbloguser does not fire init action.
```
richard015ar, line 59: I think we could remove the comment in the code, and let's leave a note here in the change history:

> This is a workaround because newbloguser does not fire init action

PR #2327, opened 2021-08-18
ci: adjust Composer configuration for WP-CLI packages to resolve installation in workflow

greatislander:
This PR modifies the Composer config for WP-CLI packages to allow PB-CLI to be properly installed. Follow-up to #2298.


SteelWagstaff (approved):
LGTM. Thanks Ned.

PR #2328, opened 2021-08-19
Prefix and suffix added

arzola:
This PR solves https://github.com/pressbooks/pressbooks-book/issues/833

We added two new fields into the contributors form to be able to append a prefix and a suffix for contributors, this will be shown on every place the contributor is printed.

### How to test

1. Create a new contributor and append the prefix and suffix or only one of them
2. Review if the prefix or suffix are being added
3. Review if those are optional

### DIFF
--- a/inc/class-contributors.php
+++ b/inc/class-contributors.php
@@ -426,7 +426,8 @@ class Contributors {
 			$last_name = get_term_meta( $term->term_id, 'contributor_last_name', true );
 			if ( ! empty( $first_name ) && ! empty( $last_name ) ) {
 				$name = $prefix ? "{$prefix} {$first_name} {$last_name}" : "{$first_name} {$last_name}";
-				$name = $suffix ? "${name} $suffix" : $name;
+				$suffix = ! empty( $suffix ) ? ( preg_match( '/^[MDCLXVI]+$/', $suffix ) ? " $suffix" : ", $suffix" ) : '';
+				$name = $suffix ? "${name}${suffix}" : $name;
 			} elseif ( ! empty( $term->name ) ) {
 				$name = $term->name;
 			}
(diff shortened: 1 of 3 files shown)


SteelWagstaff (changes requested):
We should add a comma between the last name and suffix in most cases. See note below. Otherwise, this looks great and I'd be happy to merge.

inc/class-contributors.php
```diff
@@ -406,10 +420,13 @@ public function personalName( $slug ) {
...
 		$term = get_term_by( 'slug', $slug, self::TAXONOMY );
 		if ( $term ) {
+			$prefix = get_term_meta( $term->term_id, 'contributor_prefix', true );
+			$suffix = get_term_meta( $term->term_id, 'contributor_suffix', true );
 			$first_name = get_term_meta( $term->term_id, 'contributor_first_name', true );
 			$last_name = get_term_meta( $term->term_id, 'contributor_last_name', true );
 			if ( ! empty( $first_name ) && ! empty( $last_name ) ) {
-				$name = "{$first_name} {$last_name}";
+				$name = $prefix ? "{$prefix} {$first_name} {$last_name}" : "{$first_name} {$last_name}";
+				$name = $suffix ? "${name} $suffix" : $name;
```
SteelWagstaff, line 429: The convention is to add a comma between name and suffix, unless it's a roman numeral. For example, we should write `Ken Griffey, Jr.` and `Ted Logan, Esq.` but `Alec Baldwin III`. Can we create rule to handle this that looks for `III` or `IV` or strings that start with `V` and contain 1-3 `I`s? Should cover most use cases.

SteelWagstaff (approved):
LGTM

Issue #2311, opened 2021-08-11
Add and handle profile picture field for contributors


SteelWagstaff:
As a book author, I'd like to upload small profile images for each of my book's contributors so that they can be displayed as part of their 'contributor profile' on chapters, contributor front matter, and the book's home page.

Users should either be prompted/required to upload a square image (no bigger than 500px), or should be allowed to crop an image that they've uploaded/chosen from the media library. 

If they upload a very large image, we should resize it before saving.

Contributor images should be displayed as part of the user's profile in the 'About the Authors' block.

Progress on profile picture in `hackathon-contributorProfilePicture` branch:

https://user-images.githubusercontent.com/13248424/128247043-b52358ee-3290-410c-85ef-7b47bbfe384f.mp4

Pending:
- make sure that input fields are cleared after the user's profile picture is added and a new contributor is saved 
- add user profile picture for `contributor_picture` taxonomy metadata for existing users
- Force size (square). Now it's using thumbnail by default (there is some progress at the backend level, but it needs to be improved)
- Styling

**Note**: _the plugin that Ned suggested won't work because it's for avatars, and it's a custom field (each user will have different pictures for different books, instead a user metadata - avatar for all books)_

arzola:
I tested it on integrations and works perfectly, thanks

PR #2332, opened 2021-08-20
Contributors hackathon

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2311

This PR add the `contributor_picture` field for contributors, and uses [wp_media_upload_handler](https://developer.wordpress.org/reference/functions/wp_media_upload_handler/) to the upload picture feature through the [contributor.js](https://github.com/pressbooks/pressbooks/pull/2332/files#diff-d32533293436ca4842065cc7ad39b393f588b2a33792ffcdf060a47f4b71da62) file. 
Some small JS lines were moved there to avoid php/js mixed code within in contributors forms.

By requirement the minimum picture size is 400px and should be square, so the wp media element will require cropping the image when the image is larger or it is not square.

### Test
- Under Book Info > Contributors (Book admin section) add a new contributor with its picture.
- Make sure the image was uploaded correctly by accessing to the contributor editing form (on the right side of the contributor add form).


tests/test-contributors.php
```diff
@@ -223,15 +226,17 @@ public function test_editContributorForm() {
 		$results = $this->contributor->addBlogUser( $user_id );
 
 		$term = get_term_by( 'slug', $user->user_nicename, 'contributor' );
-
 		ob_start();
 		\Pressbooks\Admin\Metaboxes\contributor_edit_form( $term );
 		$buffer = ob_get_clean();
 
-		$this->assertContains( "jQuery('.term-description-wrap').remove()", $buffer );
```
richard015ar, line 231: JS was moved to `contributors.js` file.

tests/test-metaboxes.php
```diff
@@ -283,8 +291,8 @@ public function test_contributor_metaboxes() {
 		\Pressbooks\Admin\Metaboxes\contributor_add_form();
 		$buffer = ob_get_clean();
 
-		$this->assertContains( 'window.tinyMCE.activeEditor.setContent', $buffer );
-		$this->assertContains( '.term-description-wrap', $buffer );
```
richard015ar, line 287: JS moved to `contributors.js` file

assets/src/scripts/contributors.js
```diff
@@ -0,0 +1,195 @@
+jQuery( function ( $ ) {
+	// Media
+	$( document ).ready( function () {
```
arzola, line 3: I think we can avoid the double document.ready callback because the following is the shorthand 
`
jQuery( function ( $ ) { } 
`
Now we are nesting two document ready callbacks

arzola: Probably it could be a good idea to cache in a variable some selectors like:

`jQuery( '#contributor-picture' )` to avoid querying the DOM every-time and also making it easy to refactor.

`const pictureField = jQuery( '#contributor-picture' )`

Then you can use it everywhere but the DOM lookup will only happen once.

`pictureField..val( '' )`

arzola (changes requested):
In general looks very good, just a couple of comments, and a little visual detail when the user performs a crop of the profile picture

richard015ar:
@arzola thank you for the suggestions. I already fixed those and I also fixed the cropping handler issue you reported in the latest commit.

arzola (approved):
LGTM

arzola (approved):
LGTM, thanks Ricardo

Issue #2231, opened 2021-06-30
Broken Links in Admin Screen- Contributors Page


beckej13820:
### Description

In the Admin panel, on the contributor page, there is a table that includes Name, Description, Slug, and the Count of posts.

The count of posts is a link to /wp-admin/edit.php?contributor=slug

However that page is blank and doesn't display any posts, even if the contributor has posts connected to them.

### Steps to Reproduce

1. Create a book, a contributor and assign chapters to the contribtor
2. Go to /wp-admin/edit-tags.php?taxonomy=contributor
3. Confirm that the counter works, but the links lead to an unhelpful page

**Expected behavior:** 

I'd expect the link to lead to a page with all the contributor's posts OR for the link to be removed.

**Actual behavior:** 

There is a link to a page with no useful information or purpose.

TO DO:
- [x] remove the description, slug, and count fields 
- [x] add profile picture, biographical info, and institution fields to this view so that they can be seen at a glance

gavinpressbooks:
This issue was reproduced following the steps in the original bug report without issue. Anywhere a contributor is added, whether in the Book Info or Chapter Info pages, the counter increases accordingly but following the linked number provides an empty list. A screen recording is attached.

https://user-images.githubusercontent.com/85123231/126491266-56a2b5dc-99ec-4f57-83c8-d1375cf335ec.mov

SteelWagstaff:
If possible, we should overhaul the contributor menu by removing the description, slug, and count fields and adding profile picture, biographical info, and institution fields so they can be seen at a glance.

We also want to create new bulk actions for contributors (export/import). See https://github.com/pressbooks/pressbooks/issues/1824

fdalcin:
Tested on integrations and works as expected

PR #2352, opened 2021-09-01
Change contributors list

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2231

This PR changes the Contributors list in Admin Book > Book Info > Contributors, and displays 4 columns:
- Picture
- Name (sortable)
- Institution (sortable)
- Biographical Info (sortable): Max 180 characteres

It also fixes: https://github.com/pressbooks/pressbooks/issues/2355

### Test
- Go to Admin Book > Book Info > Contributors and make sure the tables mentioned above are displayed in the order proposed.
- Make sure the sort actions works for all columns (except the Picture one, it should be disabled).

#### For https://github.com/pressbooks/pressbooks/issues/2355
- Edit a contributor with a profile picture.
- Change one or more fields, except for the picture one and save it.
- Make sure the contributor still has the profile picture in place.


inc/admin/metaboxes/namespace.php
```diff
@@ -1273,12 +1273,20 @@ function contributor_edit_form( $term ) {
 						<td>
 							<?php if ( $value ) : ?>
 								<img src="<?php echo $value; ?>" id="<?php echo $meta_tags['tag']; ?>-thumbnail" width="120" /> <br />
+							<?php else : ?>
+								<img style="display: none" id="<?php echo $meta_tags['tag']; ?>-thumbnail" width="120" />
 							<?php endif; ?>
+							<br />
```
richard015ar, line 1279: This will fix: https://github.com/pressbooks/pressbooks/issues/2355

fdalcin: Tested and it's working as expected now

inc/metadata/namespace.php
```diff
@@ -1016,6 +1017,36 @@ function register_contributor_meta() {
...
+
+	if (
+		$pagenow === 'edit-tags.php' &&
+		isset( $args['taxonomy'] ) &&
+		isset( $args['orderby'] ) &&
+		in_array( Contributors::TAXONOMY, $args['taxonomy'], true ) &&
+		in_array( $args['orderby'], $order_field_allowed, true )
+	) {
+		global $wpdb;
+		$pieces['join'] .= ' INNER JOIN ' . $wpdb->termmeta . ' AS tm ON t.term_id = tm.term_id ';
```
fdalcin, line 1043: Maybe we should use a LEFT JOIN here to avoid removing contributors that don't have a field.

fdalcin (changes requested):
LGTM @richard015ar 

Two things that I noticed.

1. We could reduce the picture size to 64px maybe and have borders rounded to have a better UI.
image

1. Sorting the list works properly, except when the field is empty. When sorting all contributors that don't have that field will be removed from the list, see below:

https://user-images.githubusercontent.com/1761690/132014095-a070ff49-6780-401c-90ad-17c5b77445ce.mov

inc/metadata/namespace.php
richard015ar: oh yes, you're right! I will try with LEFT JOIN and that should solve the issue. Thanks for catching it!

richard015ar:
@fdalcin regarding to point 1, would be nice to ask to @SteelWagstaff to confirm the UI. Honestly, I don't know what's better to UX and UI level, your suggestion looks nicer to me, but I would like to confirm this suggestion with Steel.

Thank you for catching the 2 point, I will try with the LEFT JOIN to solve it.

richard015ar:
@fdalcin the sorting feature for custom columns in the contributors list has been removed, since a custom query in a different kind of hook would be required. LEFT JOIN is useful but not enough, since the WHERE condition restricts the ordering.
Only way to order by a specific meta_value is by using WP_Query object, which requires other handling.
@SteelWagstaff mentioned this is not a required feature, although I tried to add it, after 2 hours I decided to keep it out of the scope given the priorities of this sprint.  

I added the style you suggested, now look nicer.

fdalcin (approved):

Issue #1824, opened 2019-11-03
Allow import/export of contributor lists


SteelWagstaff:
As a network manager, I want to be able to quickly export a list of contributors from one book and import them into another book so that I can attribute them on my book info page in that second book. This will be especially helpful in situations where a large group of people were involved in multiple book projects which are not clones of each other.

To make this possible, we should 1) add a 'Download' option to 'Bulk Actions' on the contributors page (would ideally produce a CSV or JSON file) and 2) add an import option that allows users to upload downloaded contributor lists and bulk create contributors from existing files. See pressbooks/ideas#82 and pressbooks/ideas#205 for similar ideas.

fdalcin:
### Update

- [ ] add a 'Download' option to 'Bulk Actions' on the contributors page
- [ ] generate a CSV or JSON file when downloading contributors
- [ ] add an import option that allows users to upload downloaded contributor lists and bulk create contributors from existing files

fdalcin:
### Update

It's now possible to import a list of contributors. The only remaining step is to add some tests to keep code coverage.

https://images.zenhubusercontent.com/9209369/e0c2ced3-de93-484d-95d9-9e2b1a089973/screen_recording_2021_09_02_at_18_13_53.mov

richard015ar:
I tested on integrations using the same cases than https://github.com/pressbooks/pressbooks/pull/2356#issuecomment-915430224
and it works as expected. It is ready to deploy to Production. Thank you @fdalcin!

PR #2356, opened 2021-09-02
Allow import/export of contributors list

### LINKED ISSUE #2360: Contributors' list does not allow selecting items to apply bulk actions
While working on #1824 I realized that #2352 has removed checkboxes that allowed selecting a list of contributors to apply bulk actions. See below:
image

Previous behavior
image

fdalcin:
Issue #1824 

This PR adds a new bulk action to export the list of contributors as a JSON file. It also adds a new form that allows users to import a list of contributors with all meta information.

This PR should also fix #2360.

#### How to test

1. Select a book that has contributors.
1. Go to Admin Book > Book Info > Contributors.
1. There should be a new form 'Import Contributors' and a new bulk action 'Download JSON'.
1. Select the list of contributors to export and download the JSON.
1. Check if the file has all contributors and all the metadata.
1. Go to another book's contributors list and import the previously download file.
1. Make sure that all contributors are imported and all metadata is present.

Things to notice: if a contributor with the same slug already exists, all the metadata will be updated (If not required, this could be removed).

#### TODO
- [ ] Define the 'Our Guide' URL that is displayed below the file input field.
- [ ] Improve message for invalid contributors.


inc/utility/class-handlestransfers.php
```diff
@@ -0,0 +1,348 @@
...
+ */
+trait HandlesTransfers {
+	/**
+	 * Adds the required filters to handle exporting and importing data.
+	 *
+	 * @param $obj
+	 * @return void
+	 */
+	public function bootExportable( Transferable $obj ) {
+		add_action( self::TAXONOMY . '_pre_add_form', [ $obj, 'renderImportForm' ] );
```
richard015ar, line 25: @fdalcin I wonder if it would make sense to add the `TAXONOMY` constant as part of the interface, given it is used here. What do you think, it is necessary?

inc/utility/class-handlestransfers.php
```diff
@@ -0,0 +1,348 @@
...
+	/**
+	 * @return array|false
+	 */
+	public function handleUpload() {
+		if ( empty( $_FILES['import_file']['name'] ) ) {
+			return false;
+		}
+
+		if ( $_FILES['import_file']['type'] !== 'text/csv' ) {
+			$_SESSION['pb_errors'][] = __( 'Sorry, this file type is not permitted for security reasons.' );
```
richard015ar, line 222: Please, confirm this message with others with Steel, just in case some English syntax issue. 
Also, I think for translation reasons, it should contains `'pressbooks'` as a second parameter, example:
```
__( 'Sorry, this file type is not permitted for security reasons.' , 'pressbooks');
```

richard015ar (changes requested):
@fdalcin I exported the following file: 
[contributor-list-1630693041.csv](https://github.com/pressbooks/pressbooks/files/7107839/contributor-list-1630693041.csv)

Then I went to another book and when I tried to import it, this happens:
Screenshot from 2021-09-03 14-20-16
When I went back to the contributors list, the following message was displayed: 
```
Successfully imported.
```
But the contributors were not imported.

Am I missing some setting?

_Note: I exported a partial list of a contributors, which means I selected 2 of 6 contributors I had. But I also tried exporting the whole list and it didn't work neither_

Also please, check the test-contributors.php file looks like it is in conflict now since another PR was merged recently.

inc/utility/class-handlestransfers.php
SteelWagstaff: This string is perfect. I think it's already included in the WordPress project: https://github.com/WordPress/wordpress-develop/blob/5f5895bf61dfe62890e296ad7acd81aad9bde88b/src/wp-admin/includes/file.php#L934. Not sure if we need to take special action to 'inherit' localized strings from WP or if it happens already ... See https://developer.wordpress.org/themes/functionality/internationalization/#loading-translations for ideas?

fdalcin: I think it happens already. I used the same approach as in here https://github.com/pressbooks/pressbooks/blob/dev/inc/modules/import/class-import.php#L421-L424

inc/utility/class-handlestransfers.php
fdalcin: I don't think it's possible to add a constant a constant as part of the interface because we won't be able to override it.

fdalcin:
@richard015ar I think that bug is fixed now and a resolved the conflicts, but it seems that codecov is complaining about coverage now.

inc/utility/class-handlestransfers.php
richard015ar: you're right Dalcin!

richard015ar:
@fdalcin this PR is an elegant OOP poetry :smile: 
Everything works on my end, here's my test notes:

- I exported Contributors with different fields filled, and others one with all fields filled.
- I used strange descriptions with multiple strings, including <script>, special chars like `\" <i-almost-a-tag` and others. The `\` c
- I exported all contributors from a book, and partially from others by selecting some of them
- I imported all of the files, and all worked

My only concern is about the description, as we discussed those are exported without HTML format. Which means URL, and other styles are not exported/imported. In understood it is to preserve the CSV format. Let's confirm it with @SteelWagstaff , since probably it should be informed to the user.
Other note is about some characters, for example this description:
```
Hey he is <strong>buddy</strong>!
I am trying to use the following strings:
" hey1
\' hey2
\" hey3
```
is imported as:
```
Hey he is Manuel!I am trying to use the following strings:" hey1' hey2" hey3
```
Looks like `\` and lines break are escaped, which means multiple lines in the description will be imported in a single line.

inc/utility/class-handlestransfers.php
```diff
@@ -0,0 +1,347 @@
...
+				$invalid_rows[] = $key + 1;
+
+				continue;
+			}
+
+			$items[] = $item;
+		}
+
+		if ( ! empty( $invalid_rows ) ) {
+			$message = __( 'One or more rows were not imported because they have missing information. Please, check rows %s', 'pressbooks' );
```
fdalcin, line 238: This is a temporary message. 

If the json structure does not contain `name` or `slug`, the contributor won't be imported and we could let the user know. Once functionality is in place we could think of a proper message here.

inc/utility/class-handlestransfers.php
```diff
@@ -0,0 +1,347 @@
...
+				}
+
+				$term
+					? update_term_meta( $results['term_id'], $field, sanitize_text_field( $item[ $field ] ) )
+					: add_term_meta( $results['term_id'], sanitize_text_field( $field ), $item[ $field ] );
+			}
+		}
+
+		if ( $changed ) {
+			$_SESSION['pb_notices'][] = __( 'Values for one or more of the imported contributors were altered by our validation routine. You may want to check the values for the newly imported contributors.', 'pressbooks' );
```
fdalcin, line 301: If we have invalid URL fields we're going to inform the user that values were changed.

inc/utility/class-handlestransfers.php
```diff
@@ -0,0 +1,347 @@
...
+		foreach ( $items as $item ) {
+			$term = get_term_by( 'slug', $item['slug'], self::TAXONOMY );
+
+			if ( ! $term ) {
+				$results = wp_insert_term( sanitize_text_field( $item['name'] ), self::TAXONOMY, [
+					'slug' => sanitize_text_field( $item['slug'] ),
+				] );
+			} else {
+				$results = wp_update_term( $term->term_id, self::TAXONOMY, [
+					'name' => sanitize_text_field( $item['name'] ),
```
richard015ar, line 264: ### Not a blocker, but a question

Dalcin, for this validation, and https://github.com/pressbooks/pressbooks/pull/2356/files#diff-9d961ff1ac28242af01296e4dd0a4f7c45f1130737cf9a11f4636add55cf6b8aR278 and https://github.com/pressbooks/pressbooks/pull/2356/files#diff-9d961ff1ac28242af01296e4dd0a4f7c45f1130737cf9a11f4636add55cf6b8aR278 and https://github.com/pressbooks/pressbooks/pull/2356/files#diff-3e98df1fc7a6e8edfe8f3516f827ab9eba589a08edc9a7ab97981fca0b80e57eR394

Would make sense to use a validator (could be a method maybe) that uses: https://github.com/pressbooks/pressbooks/blob/dev/inc/class-contributors.php#L289
Something like iterate on that array and use the unify validation methods we declared there.

inc/utility/class-handlestransfers.php
richard015ar: Makes sense, @SteelWagstaff could you confirm this message?

inc/utility/class-handlestransfers.php
```diff
@@ -0,0 +1,347 @@
...
+	 *
+	 * @param string $url
+	 * @return false|string
+	 */
+	public function handleImage( $url ) {
+		$parts = explode( '?', $url );
+		$parts = explode( '#', $parts[0] );
+		$parts = explode( '/', $parts[0] );
+
+		$filename = sanitize_file_name( end( $parts ) );
```
richard015ar, line 316: Cara, I think we want to make sure it is a valid URL as well: https://github.com/pressbooks/pressbooks/blob/dev/inc/class-contributors.php#L320

inc/utility/class-handlestransfers.php
fdalcin: I think we could have a method that handles input sanitization for sure. Only thinking how to approach this here https://github.com/pressbooks/pressbooks/pull/2356/files#diff-9d961ff1ac28242af01296e4dd0a4f7c45f1130737cf9a11f4636add55cf6b8aR300-R301

richard015ar: yes, I would consider it out of the scope. Just wondering if it is easy to add, otherwise the validation is already performed and works.

richard015ar:
Elegant solution @fdalcin thank you! 
I performed the same tests mentioned in https://github.com/pressbooks/pressbooks/pull/2356#issuecomment-915430224 and I reviewed the changes.
It is ready to go for me. After @SteelWagstaff confirmation about messages, I am ready to approve and merge it.

inc/utility/class-handlestransfers.php
SteelWagstaff: I'd prefer `The following rows were not imported because they are missing the name or slug: (%s)`

fdalcin:
@richard015ar I updated messages after talking with Steel. I think we are good to go now.

We decided to keep those warning messages as `pb_errors` for now, with the possibility to change that to warnings in a future PR.

richard015ar (approved):
LGTM!

Issue #2353, opened 2021-09-01
Contributor image crop fails with vertical images


SteelWagstaff:
### Description

I'm not able to upload and crop an image that is taller than it is wide when adding a profile picture for a contributor. 

### Steps to Reproduce

1. Create a new contributor and add a profile picture
2. Upload an image that is taller than it is wide
3. Try to crop the image
4. Try to save the image

https://user-images.githubusercontent.com/13485451/131746446-ef93727b-759f-46ef-af73-fa9bc0ab75ca.mp4

**Expected behavior:** As long as the image is larger than 400x400, I should be able to upload it and crop to expected dimensions.

**Actual behavior:** I could not see the cropping interface, nor could I apply the crop and use an image if it was taller than it is wide.

fdalcin:
Tested on integrations and works as expected.

PR #2357, opened 2021-09-03
Pb 2353 contributor picture cropping

richard015ar:
Related issue: https://github.com/pressbooks/pressbooks/issues/2353

This change fixes the WP media cropping settings for contributors' picture feature.
When the image is taller than larger and it's more than 400px of larger, the cropper fails due a pixel falls out of the image.
Previously there is not a check for minimal size (400px), there was just a suggestion. Now the the checking is performed from the back end and front end sides. Now, images smaller then 400px are not allowed to upload. If a image smaller than 400px is selected from the media library, it is not inserted for the picture and a error is displayed to inform the user.

### Test
- Add or edit a contributor in a book (Book Info > Contributors)
- Add an image larger than 400px and taller than larger. The cropper should work.
- Try to upload a image smaller than 400px (width and/or height). A message should be displayed indicating the minimum size is 400px.
- Try to add an image smaller than 400px from the media library (instead uploading from the contributors section)
- Try to add that image in any contributor's profile picture. An error message must be displayed.


assets/src/scripts/contributors.js
```diff
@@ -164,6 +164,9 @@ jQuery( function ( $ ) {
 		} );
 		pictureLibrary.on( 'select', function () {
 			const attachment = pictureLibrary.state().get( 'selection' ).first().toJSON();
+			if ( attachment.width < minPictureSize || attachment.height < minPictureSize ) {
+				return;
```
richard015ar, line 168: here an alert would be possible before closing the wp media

inc/admin/metaboxes/namespace.php
```diff
@@ -1223,7 +1223,7 @@ function contributor_add_form() {
 					<button name="dispatch-media-picture" id="btn-media">Upload Picture</button>
 					<input type="hidden" name="<?php echo $term; ?>" id="<?php echo $meta_tags['tag']; ?>">
 					<p>
-						<?php echo __( 'Images should be square and at least 400px wide. Very large images will be resized upon upload.', 'pressbooks' ); ?>
+						<?php echo __( 'Images should be square (400px x 400px). You will be allowed to crop images after upload. The maximum file size permitted is 1 MB.', 'pressbooks' ); ?>
```
richard015ar, line 1226: Confirm this with Steel

assets/src/scripts/contributors.js
```diff
@@ -77,9 +77,9 @@ jQuery( function ( $ ) {
 			minWidth: xImg > xInit ? xInit : xImg,
 			minHeight: yImg > yInit ? yInit : yImg,
 			x1: x1,
```
richard015ar, line 79: Here we can use `maxHeight` and `maxWidth` if we want to limit the cropped image resolution to a specific number of pixel. Ask to Steel.

fdalcin (approved):
@richard015ar LGTM. 

- [x] Images smaller than 400px are not allowed to be uploaded nor selected. 
- [x] Images that are greater than 400px show a cropping option with a minimum size of 400px.
- [x] Images that are taller than wider are now showing the cropping option.

Issue #2347, opened 2021-09-01
About the authors information is not being exported


richard015ar:
### Prerequisites

* [x] Can you reproduce the behavior you wish to address with [all other plugins deactivated](http://codex.wordpress.org/Multisite_Network_Administration#Plugins)?
* [x] Are you running the [latest version of Pressbooks](http://github.com/pressbooks/pressbooks/releases/latest/) and the [latest version of WordPress](http://codex.wordpress.org/Upgrading_WordPress)?

### Description

From https://github.com/pressbooks/pressbooks-book/issues/832 we know that the book's admin can display contributors information at the end of each chapter/back matter / front matter, etc through a new global option.
When the book is exported in any format, that information is not included when the book is exported.

### Steps to Reproduce

1. For a particular book go to Appearance > Theme options > Global options
2. Make sure a new About the Author checkbox option is present. It should be unchecked by default
3. Open any chapter (or back matter, front matter, etc) and add a one or more contributors if there are any. You can do it at the bottom of the editing page, in "Chapter Metadata" section.
4. Export the book in any format, for example in PDF.
5. About the authors information must be displayed for those chapter with contributors assigned.

**Expected behavior:** About the authors section must be displayed in those chapters/front/back matters, etc with contributors assigned for all export formats.

**Actual behavior:** Contributors information is not present in any export format.

PR #2362, opened 2021-09-07
Add about the authors information to XHTML export routine

richard015ar:
Related issue: https://github.com/pressbooks/pressbooks/issues/2347

This PR adds the **About the Authors** section at the end of each chapter for XHTML and EPUB (using [the pb-books PR](https://github.com/pressbooks/pressbooks-book/pull/853)) exports, including PDFs through a new method that adds the HTML at the end of the string of each chapter .

This PR requires pressbooks-book change, see: https://github.com/pressbooks/pressbooks-book/pull/853

### Test
- For a random book, assign different authors to a different chapters. Make sure those authors are book's contributor in the Book Info section.
- Make sure Appearance > Theme Options > About the authors options is enabled.
- To test it better, you may add picture, url, description and/or other fields to the contributors in Book Info > Contributors section.
- Export the book to EPUB, PDF and XHTML and make sure the authors are added at the end of those chapter with authors.


inc/modules/export/namespace.php
```diff
@@ -279,3 +281,47 @@ function update_pins() {
...
+	if ( empty( $chapter_contributors ) ) {
+		return '';
+	}
+	$title = sprintf( _n( '%s Author', '%s Authors', count( $chapter_contributors ), 'pressbooks' ), 'About the' );
+	$print = '<div class="contributors">';
+	$print .= "<h3 class=\"about-authors\">{$title}</h3>";
+	foreach ( $chapter_contributors as $contributor ) {
+		$print .= '<div class="contributor_name_and_links">';
+		if ( $contributor['contributor_picture'] ) {
+			$print .= "<img class=\"contributor_profile_picture\" src=\"{$contributor['contributor_picture']}\" />";
```
fdalcin, line 303: EPUB validation is complaining about missing alt tags here

fdalcin (changes requested):
Hey @richard015ar LGTM. I just noticed that when exporting the EPUB 2.01, it gives a warning about validation errors.

image

Logs are here:
image

Other minor things I believe are related to the styling portion of this PR.

I feel like the authors printed below the chapter name could have a smaller size, not much though.
image

For the links, I have the same feeling that we could reduce the font size.
image

inc/modules/export/namespace.php
SteelWagstaff: For alt text, we should attempt to include/display the image's alt text and fall back to `alt=""` if none has been set for the image in question. https://webaim.org/techniques/alttext/#decorative

richard015ar:
@fdalcin thank you for reviewing it.
I added the alt attribute for the img tag, now EPUB validation is not throwing any issues.
Take in consideration (which is not the previous case you mentioned in the PR comment) where some validation issues are thrown, it will depend of the book structure itself.

Regarding the authors in the index of the book is not something was added in this PR, it is another feature.

For the links size, as you mentioned to me, it is for [the related PR](pressbooks/pressbooks-book#853) but looks like a smaller size would looks nicer, what do you think @SteelWagstaff? Let's discuss it in the review meeting we have today.

SteelWagstaff:
> For the links size, as you mentioned to me, it is for [the related PR](pressbooks/pressbooks-book#853) but looks like a smaller size would looks nicer, what do you think @SteelWagstaff? Let's discuss it in the review meeting we have today.

Link size could be adjusted in the styling PR, it's true, but I think it's outside the scope of _this_ particular task. I don't think it should be a blocker for this pressbooks PR.

fdalcin (approved):
Thanks @richard015ar, LGTM. I'll wait to merge until the styling PR is approved. 

Agreed, minor styling changes shouldn't be blocker for this.

PR #2364, opened 2021-09-08
About the authors: 

richard015ar:
Related issue: https://github.com/pressbooks/pressbooks/issues/2347

This PR adds the **About the Authors** section at the end of each chapter to the HTMLBook export routine.
It also remove the contributors picture size validation at the backend level in order to avoid the validation in other media uploading.
Now the validation is done using the wp media library as recommended by WP.

### Test
- For a random book, assign different authors to a different chapters. Make sure those authors are book's contributor in the Book Info section.
- Make sure Appearance > Theme Options > About the authors options is enabled.
- To test it better, you may add picture, url, description and/or other fields to the contributors in Book Info > Contributors section.
- Export the book to HTMLBook and make sure the authors are added at the end of those chapter with authors. Take into account styling is not supported for this format.


assets/src/scripts/contributors.js
```diff
@@ -79,6 +79,8 @@ jQuery( function ( $ ) {
 			imageHeight: realHeight,
 			minWidth: xImg > xInit ? xInit : xImg,
 			minHeight: yImg > yInit ? yInit : yImg,
+			maxHeight: minPictureSize * 2,
+			maxWidth: minPictureSize * 2,
```
richard015ar, line 83: @SteelWagstaff I defined in this PR the maximum allowed size for a cropped image to 800x800, which means at the moment to crop a contributor's picture, the maximum size allowed to cover is 800x800 px.
I can remove or change this limit, but I think keep it relatively small is better for performance purposes.

hooks-admin.php
```diff
@@ -199,7 +199,6 @@
 	add_filter( 'attachment_fields_to_edit', '\Pressbooks\Admin\Attachments\add_metadata_attachment', 10, 2 );
 	add_filter( 'attachment_fields_to_save', '\Pressbooks\Admin\Attachments\save_metadata_attachment', 10, 2 );
 	add_filter( 'edit_form_top', [ '\Pressbooks\Book', 'notifyBisacCodesRemoved' ] );
-	add_filter( 'wp_handle_upload_prefilter', '\Pressbooks\Admin\Metaboxes\validate_contributor_picture_size', 10, 1 );
```
richard015ar, line 202: It is validated using wp media here: https://github.com/pressbooks/pressbooks/blob/eeeaa48b18ee796517bf6349fa73a20405dbc789/assets/src/scripts/contributors.js#L172

assets/src/scripts/contributors.js
```diff
@@ -79,6 +80,8 @@ jQuery( function ( $ ) {
 			imageHeight: realHeight,
 			minWidth: xImg > xInit ? xInit : xImg,
 			minHeight: yImg > yInit ? yInit : yImg,
+			maxHeight: maxSize,
```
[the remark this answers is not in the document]
richard015ar: @SteelWagstaff  I defined in this PR the maximum allowed size for a cropped image to 800x800, which means at the moment to crop a contributor's picture, the maximum size allowed to cover is 800x800 px.
I can remove or change this limit, but I think keep it relatively small is better for performance purposes.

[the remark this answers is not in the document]
SteelWagstaff: So you're saying if I uploaded an image with 4000 x 6000 pixels, I'd only be able to make a selection of 800 x 800? I think what I'd prefer, if not too difficult would be to let the user make a square selection on the original image and have the resulting crop resized to be 400x400. So for example, I could choose a 4000x4000 crop rectangle on a 4000x6000 pixel image but once I confirmed the crop, we'd resize the cropped image to 400x400 and use that smaller image as the contributor profile pic. Does that make sense?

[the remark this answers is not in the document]
SteelWagstaff: In other words, the cropping selection should be about placement of the square -- and then we silently drop pixels/downsize the image after the selection is confirmed.

assets/src/scripts/contributors.js
```diff
@@ -125,9 +126,9 @@ jQuery( function ( $ ) {
 			 */
 			doCrop: function ( attachment ) {
 				const cropDetails = attachment.get( 'cropDetails' );
-
-				cropDetails.dstWidth  = minPictureSize;
-				cropDetails.dstHeight = minPictureSize;
+				const minSize = parseInt( minPictureSize );
+				cropDetails.dst_width  = minSize;
+				cropDetails.dst_height = minSize;
```
richard015ar, line 131: @SteelWagstaff those properties will resize the image right after to being cropped and will save it with 400x400 pixels.

tests/test-modules-export-export.php
```diff
@@ -577,4 +577,24 @@ public function test_getContributorsForSectionXHTML() {
...
+	 * @group export
+	 */
+	public function test_endnoteShortcode() {
+		$html_book = new Pressbooks\Modules\Export\HTMLBook\HTMLBook( [ 'endnotes' => true ] );
+		$end_note = $html_book->endnoteShortcode( [] , 'I am a endnote, see you!');
+		$attributes = $end_note->getAttributes();
+		$this->assertArrayHasKey( 'class', $attributes );
+		$this->assertEquals( 'endnote', $attributes['class'] );
+	}
+
```
richard015ar, line 599: Out of scope tests. I added those for the code coverage. The line I added in the HTMLBook class is almost impossible to test without mocking half of methods.

SteelWagstaff (approved):
LGTM

Issue #2324, opened 2021-08-13
Cloned chapters are not being associated with correct parent part


pb-amy:
### Description

Chapters and their content were left out of the cloning routine. The cloning routine also failed to clone media and other elements attached to the missing chapters. However, cloning message shows 'success' when the book is cloned, and shows the number of chapters cloned over accurately.

An instance of this bug is visible:
- Original: https://openoregon.pressbooks.pub/chemeketacollegetransition/ 
- Clone: https://kpu.pressbooks.pub/testfordebbiebaker/

This was tested on University and Integrations. On University, here's the failed clone: https://university.pressbooks.pub/chemektatest/wp-admin/. On Integrations, I was able to replicate the cloning error (none of the chapters are cloned over).

The importing function is working so it’s a sufficient workaround for now.

### Steps to Reproduce

1. Clone a book
2. Check for missing chapters
 
**Expected behavior:** Cloning should bring over all content.

**Actual behavior:** Chapters are missing.

### Notes
Two institutions ([KPU](https://pressbooks.zendesk.com/agent/tickets/6789) and [Clemson](https://pressbooks.zendesk.com/agent/tickets/5763)) reported issues related to cloning

pb-amy:
Virginia Tech has also noted this issue: https://pressbooks.zendesk.com/agent/tickets/6866

SteelWagstaff:
Books where we had cloning problems:
https://btugman2021.pressbooks.com/
https://anne1.pressbooks.com/
https://pressbooks.lib.vt.edu/aerodynamics/
https://openoregon.pressbooks.pub/chemeketacollegetransition/

ho-man-chan:
integration errors during cloning
[18-Aug-2021 14:36:39 UTC] PHP Notice:  Undefined index: sizes in /srv/www/integrations.pressbooks.network/releases/20210817141820/web/app/plugins/pressbooks/inc/cloner/class-cloner.php on line 745
[18-Aug-2021 14:36:39 UTC] PHP Warning:  Invalid argument supplied for foreach() in /srv/www/integrations.pressbooks.network/releases/20210817141820/web/app/plugins/pressbooks/inc/cloner/class-cloner.php on line 745
[18-Aug-2021 14:36:51 UTC] PHP Warning:  Illegal string offset 'slug' in /srv/www/integrations.pressbooks.network/releases/20210817141820/web/app/plugins/pressbooks/inc/cloner/class-cloner.php on line 608
[...]

PR #2269, opened 2021-07-22
Add clone book page for lowly users

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/1899

The clone book page now is under `admin.php?page=pb_cloner`, and is accessible to all users, including 'lowly users' since this change gives 'lowly users' the `subscriber` role on the root site.

The cloner page is now accessible for lowly users at the top menu under `My Books`.

### Testing notes
-  Create a lowly user through: `Network Admin > Users > Add new`, it's in: `wp-admin/network/user-new.php`
- Update or check the password for the new user to make sure you'll have access to log in.
- Log in with the new user and check the dashboard remains the same as any lowly user, including top and sidebar menus, widgets, etc.
- Check at the top menu in `My Books` to confirm `Clone a book` sub menu is present. You should be able to access to the page and be able to clone a book successfully.
- Once you have cloned a book, you should still be able to access the book's dashboard using the top menu under `My Books`.
- Now the user can clone additional books under `<BOOK_URL>/wp-admin/admin.php?page=pb_cloner` or `/wp-admin/admin.php?page=pb_cloner`, depending on whether the user accesses the link from the book's dashboard or the main dashboard.

PR #2366, opened 2021-09-13
Cloning routine fix

richard015ar:
Related issue: https://github.com/pressbooks/pressbooks/issues/2324

This PR fixes 2 different problems in the cloning routine:
- `pb_part_invisible` API field in the `_embed` section for the `pressbooks/v2/toc?_embed=1` endpoint. This field is defined as a boolean. Boolean must be managed differently in WP API, see: https://core.trac.wordpress.org/ticket/48363. When there is not meta_value, instead display a boolean the API will expose a `null` value by default. In this PR the value was moved to string, which does not impact to other uses and keep consistency with other similar fields, like `pb_show_title`, see: https://github.com/pressbooks/pressbooks/blob/b9ee332180f04ae1e5a669e084c3348d9e36d6a9/inc/posttype/namespace.php#L321 . When the value is exposed as `null` it thrown an error in the internal API request in: https://github.com/pressbooks/pressbooks/blob/db69900e65e066186a20a1cdf6db06e9a367b137/inc/cloner/class-cloner.php#L1441, failing the cloning routing without any notification to the user.
- From: https://github.com/pressbooks/pressbooks/pull/2269 the cloning action was independently from a previous created book, which means this variable: https://github.com/pressbooks/pressbooks/blob/db69900e65e066186a20a1cdf6db06e9a367b137/hooks.php#L56 is false always for the cloning routine, avoiding to add the `\Pressbooks\Api\init_book` action, necessary to add the necessary API endpoints and arguments in https://github.com/pressbooks/pressbooks/blob/db69900e65e066186a20a1cdf6db06e9a367b137/inc/api/namespace.php#L55-L76, necessary for the internal API request for create chapters with the `part` parameter to relate the chapter with its part, and add other taxonomies.

### Test
- Clone the following books:
  - https://open.oregonstate.education/aandp/
  - https://openoregon.pressbooks.pub/chemeketacollegetransition/ 
  - https://wtcs.pressbooks.pub/nursingskills/
  - https://pressbooks.umn.edu/guidelines/
  - https://btugman2021.pressbooks.com/
  - https://anne1.pressbooks.com/
  - https://pressbooks.lib.vt.edu/aerodynamics/
  - https://openoregon.pressbooks.pub/chemeketacollegetransition/
 
All of those books should be cloned with the expected content.


hooks.php
```diff
@@ -53,15 +53,13 @@
 // -------------------------------------------------------------------------------------------------------------------
 
 add_filter( 'rest_index', '\Pressbooks\Api\add_help_link' );
-
 if ( $is_book ) {
 	add_action( 'rest_api_init', '\Pressbooks\Api\init_book' );
-	add_filter( 'rest_endpoints', '\Pressbooks\Api\hide_endpoints_from_book' );
```
richard015ar, line 59: This hook is obsolete, it does not add any functionality.

hooks.php
```diff
@@ -53,15 +53,13 @@
...
 	add_filter( 'rest_url', '\Pressbooks\Api\fix_book_urls', 10, 2 );
 	add_filter( 'rest_prepare_attachment', '\Pressbooks\Api\fix_attachment', 10, 3 );
 } elseif ( $enable_network_api ) {
+	add_action( 'rest_api_init', '\Pressbooks\Api\init_book' );
```
richard015ar, line 61: If API is enabled to network API, we need to allow add the endpoints for cloning including the init_book, since cloning routine now is dispatched from the main site (non-book).

inc/api/namespace.php
```diff
@@ -275,18 +275,6 @@ function hide_endpoints_from_book( $endpoints ) {
...
- *
- * @param array $endpoints
- *
- * @return array
- */
-function hide_endpoints_from_root( $endpoints ) {
-
-	return $endpoints; // Nothing to hide
-}
-
```
richard015ar, line 289: This method was removed since does not add anything, it is obsolete.

inc/cloner/class-cloner.php
```diff
@@ -1431,6 +1431,16 @@ protected function cloneSection( $section_id, $post_type, $parent_id = null ) {
...
+		if ( isset( $section['meta']['pb_part_invisible'] ) ) {
+			// pb_part_invisible metadata compatibility with previous boolean type (false | null)
+			if ( is_bool( $section['meta']['pb_part_invisible'] ) && $section['meta']['pb_part_invisible'] === false ) {
+				$section['meta']['pb_part_invisible'] = '';
+			}
+			if ( is_null( $section['meta']['pb_part_invisible'] ) ) {
+				$section['meta']['pb_part_invisible'] = 'on';
+			}
+		}
+
```
richard015ar, line 1443: for those source books with previous version containing `false` or `null` values, instead the new strings `""` or `"on"` just like others binary metadata values

hooks.php
arzola: Nice finding

inc/cloner/class-cloner.php
arzola: Nice detective work @richard015ar

arzola (approved):
@richard015ar @SteelWagstaff I tested the changes on my local and I successfully cloned all of those books, wow nice job @richard015ar  finding the issue with the pb_part_invisible

PR #2369, opened 2021-09-16
Clone contributors information

richard015ar:
This PR changes the PB API in order to include [the new contributor metadata](https://github.com/pressbooks/pressbooks/blob/dev/inc/class-contributors.php#L292-L365).
The changes in the schema to display authors, editors, contributors, illustrators, reviewers and translators is for:
- Include those in the cloning routine. It means, the metadata will be included in the cloned book. Although a new slug for each contributor will be created, if the target book contains that slug for a contributor (the admin user who is cloning the book) the default contributor metadata in the target book will be overwritten.
- Include additional authors and contributors in the export routine for Epub and xhtml formats

Other changes were made, which were agreed with @SteelWagstaff:
- The contributors name in the API are exposed with subfix and suffix, example: https://integrations.pressbooks.network/contributors/wp-json/pressbooks/v2/metadata. If suffix contains any letter from the roman numbers it will avoid to display the comma after the last name to be exposed, see: https://github.com/pressbooks/pressbooks/blob/3217c734b9f3d25d840c247a5ac5b16b6c1698cb/inc/admin/class-sitemap.php#L67, it may be a problem for example for a suffix: `Vic` and a full name: `Carl Thomson.` It will detect the `V` and will display in the API something like: `Carl Thomson Vic` although we want: `Carl Thomson, Vic` since `V` in this case is not a roman number .  It was changed for a fixed subset discussed, since we don't want to add a comma for those letter, see: https://github.com/pressbooks/pressbooks/pull/2369/files#diff-3e98df1fc7a6e8edfe8f3516f827ab9eba589a08edc9a7ab97981fca0b80e57eR638
- Taking into account the previous point, the [oxford_comma](https://github.com/pressbooks/pressbooks/blob/11e40fac4f04096c2ff2e308b6aa0e907d9df828/inc/utility/namespace.php#L1290) method causes some issues when we have contributors name exposed in the API with a comma, i.e.: `Carl Thomson, Vic`. It will be exposed in the API like to different contributors, since [it explode the list by commas](https://github.com/pressbooks/pressbooks/blob/11e40fac4f/inc/metadata/namespace.php#L261-L319). So in this PR the oxford_comma method and the implode one was changed by `;` semicolon, instead comma, to avoid this problem. Nevertheless, if some contributor contains a semicolon (for API version previous to this change) in the name it will cause problems and will split the name in 2 different contributors. It would not be a problem for this API version, since we are exposing and extracting contributors by certain character.
- **If a contributors is not a author of any chapter / front matter / back matter won't be displayed in the API and won't be cloned.**
- The contributor's picture will be cloned to the media of the targeted book, and the picture field will be pointed to that new image cloned.
- If the source book contains previous API version of this change, the contributors metadata won't be exposed in the API, and only the name will be taken, other contributor metadata won't be cloned.
- The ICML format was removed completely. 

### Test
- Create multiple contributors for a book with different fields. Some of them you could use comma in the name and roman numbers like: `II`, `IV` or `XI` por example (Max: `XII`).
- Add Contributors, authors, illustrators, reviewers, editors and translators to the book
- Assign multiple chapters/front/back matter to different contributors
- Clone the book.
- Make sure the new cloned book contains all the contributors presents in the Book Info section of the source book (authors, editors, illustrators, etc) and those that are authors of chapters, back matters and front matters.
- Make sure all the contributor metadata is included in the target book.
- You can test the metadata API endpoints and check the contributors in `<BOOK_URL>/wp-json/pressbooks/v2/metadata` and in the chapter's metadata URL: `<BOOK_URL>/wp-json/pressbooks/v2/chapters/<CHAPTER_ID>/metadata`
- Export the cloned book (or any another you have in your local environment) to PDF and EPUB 2 and 3 and make sure the book is exported correctly. The list of authors and contributors (contributors field in Book Info section, **not** the contributor's list in the Book Info > Contributors section) must be listed in at the beginning of the book.


inc/modules/export/namespace.php
```diff
@@ -164,7 +158,6 @@ function filetypes() {
 			'pdf' => '.pdf',
 			'print_pdf' => '._print.pdf',
 			'mobi' => '.mobi',
-			'icml' => '.icml',
```
SteelWagstaff, line 167: We need to leave this mapping in place so that the export menu displays properly in case a user previously made an ICML export long ago and has one in its export list. Same logic for leaving mobi here ...

inc/modules/export/xhtml/class-xhtml11.php
```diff
@@ -1003,10 +1014,14 @@ protected function echoTitle( $book_contents, $metadata ) {
 			printf( '<h1 class="title">%s</h1>', get_bloginfo( 'name' ) );
 			printf( '<h2 class="subtitle">%s</h2>', ( isset( $metadata['pb_subtitle'] ) ) ? $metadata['pb_subtitle'] : '' );
 			if ( isset( $metadata['pb_authors'] ) ) {
-				printf( '<h3 class="author">%s</h3>', $metadata['pb_authors'] );
+				foreach ( $metadata['pb_authors'] as $author ) {
+					printf( '<h3 class="author">%s</h3>', $author['name'] );
```
SteelWagstaff, line 1018: This is going to result in several H3 elements, rather than one H3 element with several authors. Probably not our desired behaviour.

inc/modules/export/xhtml/class-xhtml11.php
```diff
@@ -1003,10 +1014,14 @@ protected function echoTitle( $book_contents, $metadata ) {
...
+				}
 			}
 			if ( isset( $metadata['pb_contributors'] ) ) {
-				printf( '<h3 class="author">%s</h3>', $metadata['pb_contributors'] );
+				foreach ( $metadata['pb_contributors'] as $contributor ) {
+					printf( '<h3 class="author">%s</h3>', $contributor['name'] );
```
SteelWagstaff, line 1023: Same as previous comment.

tests/test-api.php
```diff
@@ -59,7 +59,14 @@ public function test_booksEndpointMetada() {
 	 */
 	public function test_informationToSchema() {
 		$book_information = [
-			"pb_authors" => "admin",
+			"pb_authors" => [
+				[
+					'name' => 'admin',
+					'contributor_first_name' => 'Pat',
+					'contributor_last_name' => 'Metheny',
+					'contributor_description' => 'The drummer is the leader of any band',
```
SteelWagstaff, line 67: Ha -- nice!

inc/modules/export/xhtml/class-xhtml11.php
richard015ar: Yes, it produces multiple H3 elements in the exported file. What would be the output for multiple authors and contributors? Maybe an "Authors" title (h3) with a list (ul)? and same for contributors.

inc/modules/export/namespace.php
richard015ar: I see, makes sense. I'll revert this line.

richard015ar:
This PR is ready for review. @SteelWagstaff just I would like to confirm the final structure for the list of authors & contributors in the 1st page for a book in the export routine (xhtml format).

inc/cloner/class-cloner.php
```diff
@@ -1335,9 +1341,11 @@ protected function cloneMetadata() {
 			}
 		}
 
-		// Remove the current user from the author field in Book Info
 		$user_data = get_userdata( get_current_user_id() );
-		$this->contributors->unlink( $user_data->user_nicename, $metadata_post_id );
+		if ( ! in_array( $user_data->user_nicename, $authors_slug, true ) ) {
```
richard015ar, line 1345: We don't want to unlink if we find the same contributor is an author in the source book (same contributor means that contains same slug or name). Test case: clone a book whose author contains the same slug or name as the one automatically generated after cloning in the target book.

inc/metadata/namespace.php
```diff
@@ -260,63 +265,26 @@ function book_information_to_schema( $book_information, $network_excluded_direct
...
-
-	if ( isset( $book_information['pb_contributors'] ) ) {
-		$contributing_authors = oxford_comma_explode( $book_information['pb_contributors'] );
-		foreach ( $contributing_authors as $contributor ) {
-			$book_schema['contributor'][] = [
-				'@type' => 'Person',
-				'name' => $contributor,
-			];
+	$contributors = new Contributors();
+	foreach ( $contributors->valid as $contributor_type ) {
```
richard015ar, line 269: Instead multiple if for each type of contributors and repeat the logic, I used the valid public property

inc/metadata/namespace.php
```diff
@@ -260,63 +265,26 @@ function book_information_to_schema( $book_information, $network_excluded_direct
...
+		if ( isset( $book_information[ $contributor_type ] ) ) {
+			// Compatibility with previous version (basic strings contributors)
+			if ( is_string( $book_information[ $contributor_type ] ) ) {
+				$contributors_string = oxford_comma_explode( $book_information[ $contributor_type ] );
+				$book_schema[ $mapped_properties[ $contributor_type ] ] = [];
+				foreach ( $contributors_string as $contributor_name ) {
+					$book_schema[ $mapped_properties[ $contributor_type ] ][] = [
+						'@type' => 'Person',
+						'name' => $contributor_name,
```
richard015ar, line 278: Previous version will contains only the name.

inc/modules/export/epub/class-epub201.php
```diff
@@ -2603,10 +2621,18 @@ protected function createNCX( $book_contents, $metadata ) {
...
+			foreach ( $metadata['pb_authors'] as $author ) {
+				$authors[] = $author['name'];
+			}
+			$authors = oxford_comma( $authors );
+		}
 
 		// Sanitize variables for usage in XML template
 		$vars = [
-			'author' => ! \Pressbooks\Utility\empty_space( $metadata['pb_authors'] ) ? sanitize_xml_attribute( oxford_comma_explode( $metadata['pb_authors'] )[0] ) : '',
+			'author' => sanitize_xml_attribute( $authors ),
```
richard015ar, line 2635: Before we just display the first author listed. Now we are displaying the list of authors.

inc/modules/export/xhtml/class-xhtml11.php
SteelWagstaff: This function displays the text that appears on the 'title page' for export files. Right now we display a series of heading elements -- but this is part of what should be refactored when we improve our exports -- post of these should probably be printed as paragraphs with specific classes that are styled to display as desired. For now, we can simply print a single h3 element with class authors and then the names of each author with commas between them and an 'and' before the last authors name.

SteelWagstaff:
> This PR is ready for review. @SteelWagstaff just I would like to confirm the final structure for the list of authors & contributors in the 1st page for a book in the export routine (xhtml format).

See: https://github.com/pressbooks/pressbooks/pull/2369#discussion_r712195273. We need to make a slight change to this function to preserve the desired functionality (display all authors/contributors within the same `<h3>` element on the title page). Replacing the `<h3>` with `<p>` does not have to be addressed in this PR -- it can/should be part of the XHTML/PDF export refactor job later.

arzola (dismissed):
Excellent work @richard015ar 

I tested this PR and seems working good in general, I just noted @SteelWagstaff notes about the several h3 tags created instead of having a wrapper with an oxford comma separated I think that should be fixed in exports

I tried a lot of different combinations using commas in the name and works good, the API exposes the right values but I think it could be helpful to have some books with real contributor structure to see if everything looks good.

inc/utility/namespace.php
```diff
@@ -1279,7 +1279,7 @@ function rmrdir( $dirname, $only_empty = false ) {
 
 
 /**
- * Comma separated, Oxford comma, localized and between the last two items
+ * Comma separated, Oxford semicolon localized and between the last two items
```
SteelWagstaff, line 1282: Comment should read: 'Semicolon separated list, with a localized 'and' between the last two items'

inc/utility/namespace.php
```diff
@@ -1291,27 +1291,38 @@ function oxford_comma( array $vars ) {
...
+			$output .= $str_implode . __( 'and', 'pressbooks' ) . ' ';
 		}
 		$output .= $last;
 		return $output;
 	}
 }
 
 /**
- * Explode an oxford comma seperated list of items
+ * Explode an oxford semicolon seperated list of items
```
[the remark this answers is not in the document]
richard015ar: Should be it "Explode a semicolon oxford seperated list of items" ? what do you think? considering we are exploding/removing the `and` if present.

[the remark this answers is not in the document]
SteelWagstaff: We could say 'Explode an 'Oxford' semicolon separated list of items', yes. I think the meaning will be clear enough.

fdalcin (commented):
I'm still digging the code and reviewing everything, but it's an awesome work so far @richard015ar.

Found an issue related to saving contributors to chapters or to the book info. See below:
https://user-images.githubusercontent.com/1761690/134223284-b23bd7bc-793f-4ba8-9dc5-963b1a09c621.mov

As we discussed seems to be related to changes on this function https://github.com/pressbooks/pressbooks/blob/e983c11950321102166afec59674cb13dca95f36/functions.php#L97 and other plugins/themes might require changes.

inc/class-contributors.php
```diff
@@ -153,8 +155,19 @@ public function getArray( $post_id, $contributor_type ) {
 		if ( is_array( $meta ) ) {
 			foreach ( $meta as $slug ) {
 				$name = $this->personalName( $slug );
-				if ( $name ) {
-					$contributors[] = $name;
+				if ( ! $include_term_meta ) {
+					if ( $name ) {
+						$contributors[] = $name;
+					}
```
fdalcin, line 161: Is this condition necessary?

inc/class-contributors.php
```diff
@@ -541,7 +633,19 @@ public function personalName( $slug ) {
 			$last_name = get_term_meta( $term->term_id, 'contributor_last_name', true );
 			if ( ! empty( $first_name ) && ! empty( $last_name ) ) {
 				$name = $prefix ? "{$prefix} {$first_name} {$last_name}" : "{$first_name} {$last_name}";
-				$suffix = ! empty( $suffix ) ? ( preg_match( '/^[MDCLXVI]+$/', $suffix ) ? " $suffix" : ", $suffix" ) : '';
+				if ( ! empty( $suffix ) ) {
+					$does_contains_roman_number = false;
+					$roman_numbers_suffix = [ 'II', 'III', 'IV', 'V', 'VI', 'VII', 'VIII', 'IX', 'X', 'XI', 'XII' ];
+					foreach ( $roman_numbers_suffix as $roman_number ) {
+						if ( strpos( $suffix, $roman_number ) ) {
```
fdalcin, line 640: I think that we need to check for `strpos( $suffix, $roman_number ) !== false` otherwise we will miss when position is `0 (zero)` .

inc/modules/export/epub/class-epub201.php
```diff
@@ -1054,10 +1054,16 @@ protected function createTitle( $book_contents, $metadata ) {
 			$html .= sprintf( '<h1 class="title">%s</h1>', get_bloginfo( 'name' ) );
 			$html .= sprintf( '<h2 class="subtitle">%s</h2>', ( isset( $metadata['pb_subtitle'] ) ) ? $metadata['pb_subtitle'] : '' );
 			if ( isset( $metadata['pb_authors'] ) ) {
-				$html .= sprintf( '<h3 class="author">%s</h3>', $metadata['pb_authors'] );
+				if ( isset( $metadata['pb_authors'] ) ) {
```
fdalcin, line 1057: Is this double check required?

inc/class-contributors.php
richard015ar: yes, it is. I improved it a little bit in the latest version. But $name is a result of `personalName` method and it could be `''` which means there are not any contributors with that slug. In normal conditions it should not happen, but there is an issue where we left term metadata information after delete a contributor. I addressed that issue here: https://github.com/pressbooks/pressbooks/issues/2385
In summary, if we found the metadata, but we aren't able to see the name, we don't want to list the contributors because it was deleted.

fdalcin: Thanks for clarifying it

richard015ar:
Note for testers: Now we have a incompatible version on staging environment, which means if you clone a book from integrations or textopress it won't work as expected. Nevertheless, if you clone a book from production, i.e. from university.pressbooks.network, it should work.
The main incompatibility issue is the oxford comma issue + this method: https://github.com/pressbooks/pressbooks/blob/0df3ccf454d7f4724707bb87d698232a11ab78ea/inc/class-contributors.php#L630 which uses commas to append the contributor's suffix in most of the cases.

fdalcin:
@richard015ar testing everything again. 

Some notes that I noticed with this last version:

### Cloning a book from the same API version

Book is cloned properly, however it doesn't look like contributors are being imported properly, see images below:

Contributors page -- original first, clone after
image
image

Edit contributor page -- original first, clone after
image
image

Book info page -- original first, clone after
image
image

Somehow there are some missing information. Also, displaying authors is only showing the first name, see:
image
image

### Cloning a book from previous API version

Book is cloned properly with the correct contributor naming, see:

Contributors page
image

Book info page
image

Book home page
image

### Exporting books

Doesn't matter which API version I cloned a book from, exports fail and give me the following error -- both PDF and EPUB:

```
2021/09/24 02:06:53 [error] 1023#1023: *1626 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Pressbooks\Utility\get_contributors_name_imploded() must be of the type array, string given, called in /srv/www/pressbooks.test/current/web/app/plugins/pressbooks/inc/modules/export/epub/class-epub201.php on line 1058 and defined in /srv/www/pressbooks.test/current/web/app/plugins/pressbooks/inc/utility/namespace.php:1286
```

richard015ar:
@fdalcin thank you for catching it. I fixed it and I also fixed the problem with Epub export routines. The error you displayed was regarding to the compatibility changes I made.

fdalcin (changes requested):
Hi @richard015ar, thanks for all your work so far.

Here are some more notes:

I imported a book from the previous version of the API and this issue happens in PDF exports -- EPUB exports are ok
image

Imported a book using the newest API version and while exporting files I got this issue
```
2021/09/24 19:31:59 [error] 1023#1023: *2001 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Pressbooks\Utility\get_contributors_name_imploded() must be of the type array, string given, called in /srv/www/pressbooks.test/current/web/app/plugins/pressbooks/inc/modules/export/xhtml/class-xhtml11.php on line 1018 and defined in /srv/www/pressbooks.test/current/web/app/plugins/pressbooks/inc/utility/namespace.php:1286
Stack trace:
#0 /srv/www/pressbooks.test/current/web/app/plugins/pressbooks/inc/modules/export/xhtml/class-xhtml11.php(1018): Pressbooks\Utility\get_contributors_name_imploded('John Doe; Jane ...')```

Maybe the fix from the EPUB version should be applied to other export formas as well?

fdalcin (approved):
LGTM @richard015ar. Cloning routine seems to be working properly for previous API version as well as for the new one, and the exports are being generated properly. 

Thanks cara!

Issue #2371, opened 2021-09-16
Fix 'contributor' display throughout webbook and exports


SteelWagstaff:
## Remaining issues
- [ ] #2401
![Screenshot from 2021-09-24 20-21-11](https://user-images.githubusercontent.com/13485451/134756211-e2313764-37f7-4733-866a-a708fbbd7562.png)
![Screenshot from 2021-09-24 20-17-02](https://user-images.githubusercontent.com/13485451/134756218-4540c751-b51f-442d-a259-bcc42261ebff.png)
- [x] Twitter links get turned into oEmbeds in EPUB exports (just like they do in webbooks if you paste them on their own line). We need to turn off oEmbed for these links in our EPUB export routine. i think we have a function called overrideEmbeds: https://github.com/pressbooks/pressbooks/blob/dev/inc/interactive/class-content.php#L430-L443 which calls a function called replaceOembed: https://github.com/pressbooks/pressbooks/blob/dev/inc/interactive/class-content.php#L256-L280 that we use to replace oEmbed content using a Blade tempalte: https://github.com/pressbooks/pressbooks/blob/dev/templates/interactive/oembed.blade.php
Here's the output in the HTML of the EPUB which has a twitter link that has been 'oembedded' into an iframe:
` <p><a class="twitter-timeline" href="https://twitter.com/steelwagstaff?ref_src=twsrc%5Etfw">Tweets by steelwagstaff</a><script async="" src="https://platform.twitter.com/widgets.js" charset="utf-8"></script></p>`. I think there are two problems:
1. we're replacing the actual text inside the link
2. we're including a script which causes validation errors and display problems in certain EBOOK readers
The result of our not overriding these oEmbeds are broken EPUBs and lots of validation errors. 
![Screenshot from 2021-09-24 20-49-20](https://user-images.githubusercontent.com/13485451/134756924-efa73f87-b9fa-48d7-bba6-4d9cc6c515f3.png)
- [x] Don't automatically create 'Contributors' back matter when new books are created
- [x] We need to figure out better way to render HTML text in contributor bio. Current method which uses wp_kses, strips out `<p>` tags and other unwanted html elements: https://github.com/pressbooks/pressbooks/blob/3ab65feca72e43439e143a7ddab3e1fc175f2c0e/templates/posttypes/contributor.blade.php#L83 
[...]

PR #2372, opened 2021-09-16
Fix contributor exports

SteelWagstaff:
Partial fix for #2371

### DIFF
--- a/inc/class-contributors.php
+++ b/inc/class-contributors.php
@@ -82,7 +82,7 @@ class Contributors implements BackMatter, Transferable {
 
 		add_action(
 			'pb_pre_export', function () use ( $obj ) {
-				$this->exporting = true; //only set this variable during exports
+				$obj->exporting = true; //only set this variable during exports
 			}
 		);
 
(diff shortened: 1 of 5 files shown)


fdalcin (changes requested):
Everything LGTM @arzola. 

As we discussed just now, this PR just needs to port template changes related to screen-reader texts from https://github.com/pressbooks/pressbooks-book/pull/857.

fdalcin:
@SteelWagstaff @arzola Web book looks good and the styling is nice while changing themes as well.

While testing exporting PDFs and EPUBs I noticed some styling issues
1. Name seems to not be aligned properly
2. Screen reader text is being rendered when it shouldn't -- website:
3. Links seem to be breaking lines -- not sure if is the intended behaviour
image

4. The contributors back-matter post type seems to not replace links properly
image

arzola:
@fdalcin I sent the fix for the point 4

fdalcin:
@arzola added a friendly commit to remove `role="presentation"` that was causing EPUB validation erros.

arzola:
@fdalcin I also sent a little commit to fix an unclosing section div for exports

fdalcin (approved):
@arzola LGTM now. Tested again viewing the book page and exporting the book. 

Screen reader texts don't appear anymore on exports, no validation errors, and styling is better after removing screen reader texts. Links are breaking to a new line because they are too long, so I think it's fine.
image

Issue #2354, opened 2021-09-01
Contributors problems with XML import routine


SteelWagstaff:
### Description

Problem 1: Our XML import routine doesn't handle chapter with multiple authors associated upon import. Chapters which had multiple authors in the original only have one author post-import.

Problem 2: Our XML import routine doesn't adjust the associated contributor in cases where we have a duplicate contributor when we export/import XML from one book to another. References to the altered contributor are replaced by references to the unaltered contributor which has the same slug as the contributor from the original book. 

### Steps to Reproduce

1. Create a new book
2. Edit the contributor information for your user in that book. 
3. Create one or more additional contributors in the book.
4. Associate your user and these other contributors as authors/editors, etc for the book metadata and as authors in various chapters
5. Export this book as a Pressbooks XML file
6. Create a second new book. Delete the default front/back matter and chapter in this book.
7. Import the Pressbooks XML from the first book

https://user-images.githubusercontent.com/13485451/131748772-7cf28ab2-af3d-485e-beaa-7fc40da2df18.mp4

**What you expected to happen:** the imported book is identical to the target book.

**What did happen:** Multiple new contributors are created, and the contributor for your user is created with a disambiguated slug
![Screenshot from 2021-09-01 14-39-41](https://user-images.githubusercontent.com/13485451/131748896-e2c87bb5-fd5c-46e7-9b87-429757e83148.png)

The book info references the contributor in the 2nd book rather than the expected contributor from the first book in both book info and individual chapters
![Screenshot from 2021-09-01 14-40-17](https://user-images.githubusercontent.com/13485451/131749016-1df71f0a-9109-4f3b-b687-6c9e03f9ae84.png)
![Screenshot from 2021-09-01 14-40-27](https://user-images.githubusercontent.com/13485451/131749013-bfb0b2f3-58da-45f3-b353-59852c394438.png)

Chapters with multiple authors only include one author in the imported book. When the single author is the book creator, the reference points to the wrong contributor (i.e. has the same slug in the target book as it did in the source)
![Screenshot from 2021-09-01 14-40-39](https://user-images.githubusercontent.com/13485451/131749030-f3b70a80-c577-4036-a5f0-8c1de678a4aa.png)

SteelWagstaff:
I'm still seeing problems with the XML import routine when contributors in the target book have the same name and slug as contributors in the book being imported from. See https://integrations.pressbooks.network/xmlimport/ as an example. Prior to import, the book had the following contributors:
[...]

PR #2373, opened 2021-09-17
Pb 2354 xml import fix

arzola:
This PR solves #2354 

I added an extra check to be sure we don't duplicate contributors when import and also will create new slugs and match those slugs with the created ones.

### How to test

1. Checkout this branch
2. Register a Contributor in the target book and create another contributor in the source book with the same slug to prove that contributor will not be overriden and a new one would be created with another slug.
2. Export any book that contains one of more contributors added to chapters and also could be a great idea to use some contributors with suffix and prefix information, profile picture, etc...
3. Chapters should be imported as expected with the right contributors added, also profile pictures should be copied to the target book
4. Try to import again the file and the contributors should not be duplicated

### DIFF
--- a/inc/modules/import/wordpress/class-wxr.php
+++ b/inc/modules/import/wordpress/class-wxr.php
@@ -223,7 +223,7 @@ class Wxr extends Import {
 	public function findExistentTerm( $term ) {
 
 			$args = [
-				'taxonomy' => 'contributor',
+				'taxonomy' => Contributors::TAXONOMY,
 				'hide_empty' => false,
 				'meta_query' => [], // phpcs:ignore
 			];
@@ -312,12 +312,12 @@ class Wxr extends Import {
 
 		// and import them if they don't already exist.
 		foreach ( $terms as $t ) {
-			if ( $t['term_taxonomy'] === 'contributor' ) {
+			if ( $t['term_taxonomy'] === Contributors::TAXONOMY ) {
 				// Find an equal contributor if exist get the slug and avoid dupes
 				$existent_term = $this->findExistentTerm( $t );
 				if ( ! $existent_term ) {
 					$term = $this->insertTerm( $t );
-					$new_term = get_term( $term['term_id'], 'contributor' );
+					$new_term = get_term( $term['term_id'], Contributors::TAXONOMY );
 				}
 				$this->contributorsSlugsToFix[ $t['slug'] ] = $existent_term ? $existent_term->slug : $new_term->slug;
 			}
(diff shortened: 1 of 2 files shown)


inc/modules/import/wordpress/class-wxr.php
```diff
@@ -171,6 +184,73 @@ function setCurrentImportOption( array $upload ) {
...
+	 */
+	public function findExistentTerm( $term ) {
+
+			$args = [
+				'taxonomy' => 'contributor',
+				'hide_empty' => false,
+				'meta_query' => [], // phpcs:ignore
+			];
+
+			$fields_to_compare = [ 'contributor_first_name', 'contributor_last_name', 'contributor_prefix', 'contributor_suffix' ];
```
richard015ar, line 231: Os, a question here: does it means if we find a contributor with let's say same first name and prefix, but the rest is different, do we consider it as a existent term?

inc/modules/import/wordpress/class-wxr.php
```diff
@@ -171,6 +184,73 @@ function setCurrentImportOption( array $upload ) {
...
+
+	/**
+	 * This function search for existent contributors by checking every field to look for exact duplicates.
+	 * @param $term
+	 * @return false|mixed
+	 */
+	public function findExistentTerm( $term ) {
+
+			$args = [
+				'taxonomy' => 'contributor',
```
richard015ar, line 226: Suggestion no required:
Os, in this line and in https://github.com/pressbooks/pressbooks/pull/2373/files#diff-e1f448c60b824b3e2beafb94d5c36f83b0db6e4135db5ea32dcd54a45fe486d4R315 the taxonomy constant could be added like you did in: https://github.com/pressbooks/pressbooks/pull/2373/files#diff-e1f448c60b824b3e2beafb94d5c36f83b0db6e4135db5ea32dcd54a45fe486d4R625

tests/test-modules-import.php
```diff
@@ -96,4 +97,108 @@ public function test_scrapeAndKneadImages() {
...
+	public function test_searchMultipleContributorValues() {
+		$contributors = new \Pressbooks\Contributors();
+		$contributors->insert( 'Leo Schopenhauer', 1 );
+		$contributors->insert( 'Leo Simon', 1 );
+		$contributors->insert( 'Mary User', 1, 'pb_editors' );
+
+		$post_meta = [
+			[
+				'key' => 'pb_authors',
+				'value' => 'Leo',
```
richard015ar, line 177: :1st_place_medal:

inc/modules/import/wordpress/class-wxr.php
arzola: Only If all of those fields are the same we consider that is the same term

richard015ar (approved):
Looks elegant to me! thank you @arzola!

Issue #2374, opened 2021-09-17
About the author(s) section does not appear for front/back matter in exports


SteelWagstaff:
We're currently only displaying the 'About the author(s)' section for chapters in our XHTML/HTMLBook/EPUB/PDF exports. When the theme option is selected, this section should be displayed for front/back matter content which has authors assigned as well.

See https://github.com/pressbooks/pressbooks/blob/c345db61801f5b593486a57eeeea0a70f9a7e952/inc/modules/export/xhtml/class-xhtml11.php#L1548 for example of how it is implemented for chapters in XHTML exports.

We should probably also change chapter_id to post_id here: https://github.com/pressbooks/pressbooks/blob/c345db61801f5b593486a57eeeea0a70f9a7e952/inc/modules/export/namespace.php#L291

SteelWagstaff:
Confirmed fixed on staging environment

PR #2389, opened 2021-09-24
Add about the authors information at the end of back and front matter

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2374

This PR adds about the authors information to PDF, EPUB 2 and 3 and HTMLBook export routines for front and back matters.

### Notes about how to test using different use cases
Enable the `Display about the author` global option (Appearance > Theme options), and put different authors in chapters , back matters and front matters.
Those should be displayed at the end of those.


SteelWagstaff (approved):
Works as expected in my local branch. Approved.

PR #2393, opened 2021-09-25
Blade Template Fixes

SteelWagstaff:
Partial fix for https://github.com/pressbooks/pressbooks/issues/2371. Reverts automatic contributors back matter creation. Fixes class for contributor SVG icons and class for contributor info when profile pic is present in Blade template. Accompanies SCSS fixes in https://github.com/pressbooks/pressbooks-book/pull/865

### DIFF
--- a/inc/class-activation.php
+++ b/inc/class-activation.php
@@ -229,12 +229,6 @@ class Activation {
 				'post_type' => 'back-matter',
 				'menu_order' => 1,
 			],
-			'contributors' => [
-				'post_title' => __( 'Contributors', 'pressbooks' ),
-				'post_name' => 'contributors',
-				'post_type' => 'back-matter',
-				'menu_order' => 2,
-			],
 			// Pages
 			'authors' => [
 				'post_title' => __( 'Authors', 'pressbooks' ),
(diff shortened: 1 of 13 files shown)


fdalcin:
@SteelWagstaff sent a friendly commit that fixes HTML rendered for the contributor bio. 

Taking a look if I can find a fix for extra `<br>`  and oEmbeds too.

templates/posttypes/contributor.blade.php
```diff
@@ -3,82 +3,84 @@
...
 					@endif
-				@endif
-			</div>
-		@endif
+				</div>
+			@endif
+		</div>
 	</div>
-	<div class="contributor__bio">{!! wp_kses( $contributor['contributor_description'], true ) !!}</div>
+	<div class="contributor__bio">{!! apply_filters( 'the_content', $contributor['contributor_description'] ) !!}</div>
```
fdalcin, line 85: Using the same filter used to render front-matter/chapter/back-matter body.

templates/posttypes/contributor.blade.php
```diff
@@ -3,82 +3,84 @@
...
-			<p class="contributor__institution">
-				@if ( !isset( $exporting ) )
-					<span class="screen-reader-text">institution: </span>
-				@endif
-				{{ $contributor['contributor_institution'] }}
-			</p>
-		@endif
-		@if ( $contributor['contributor_user_url'] )
-			<p class="contributor__website">
+		<div class="{{ $contributor['contributor_picture']  ? 'contributor__info__small' : 'contributor__info__large' }}">
```
fdalcin, line 6: Wrapping everything but the profile picture, so we can have a better control of the info block.

inc/class-activation.php
```diff
@@ -229,12 +229,6 @@ private function wpmuActivate() {
 				'post_type' => 'back-matter',
 				'menu_order' => 1,
 			],
-			'contributors' => [
```
SteelWagstaff, line 232: No longer automatically creates content with the contributors back matter type when registering a new book.

inc/cloner/class-cloner.php
```diff
@@ -1107,7 +1107,6 @@ public function removeDefaultBookContent( $contents ) {
 				'main-body',
 				'chapter-1',
 				'appendix',
-				'contributors',
```
SteelWagstaff, line 1110: No longer needed, since contributors back matter isn't going to be default book content

templates/posttypes/contributor.blade.php
arzola: In my local this is causing a infinite loop when one of my contributors doesn't have bio I think that filter is not suited for that because is doing like a recursive call maybe just calling our sanitize_string function allowing safe HTML could be a good option I sent a friendly commit

fdalcin: Could it be a cache problem?

arzola (approved):
LGTM I tested it, we would need to adjust the twitter/urls behavior

Issue #2358, opened 2021-09-04
Include new contributors in cloning routine


SteelWagstaff:
## Desired Outcome

When we clone an existing book, the resulting clone should include all of the contributors from the source book which were a) assigned a role in book info, b) assigned as an author to front/back matter, parts, or chapters. Each of these contributors should include all of the same contributor info they had in the source book (name, slug, first name, bio, links, profile photo, etc.) and should be associated with book info and book content in identical ways. 

## Notes
When we clone a new book, we first have to create a book shell. When we do this, a contributor is automatically created for the person who is cloning the book. It is possible that the cloned book will also have a contributor with the same slug as this new book. If this is the case, the contibutor in the source book should be overwritten and replaced by the contributor from the target book. At the end of the clone there should be no duplicate contributors, in other words. 

### Testing Scenarios
1. Ricardo clones a book which has 4 contributors in its contributor list (Oscar, Steel, Ho Man, and Dalcin). In book info, Oscar is listed as the editor, Steel listed as the author, and Ho Man and Dalcin listed as contributors. At the end of the cloning routine, the newly cloned book has 5 contributors in its contributor list: Oscar, Steel, Ho Man, Dalcin and Ricardo (the person who initiated the clone). In book info, Oscar is listed as the editor, Steel is listed as the author, and Ho Man & Dalcin are listed as the contributor. All of the contributor info for Oscar, Steel, Ho Man, and Dalcin should be cloned in the target book exactly as they appeared in the source book. Ricardo should exist in the contributor list but should not be assigned to a role in book info.
2. Ricardo clones a book which has 2 contributors in its contributor list: Steel and Ricardo. Steel is listed as the editor and Ricardo is listed as the author. During the cloning routine, we notice that the slugs for the two Ricardos are identical, so we overwrite the contributor information for Ricardo in the target book with the information from the source book. At the end of the clone, there are 2 contributors in the contributor list for the cloned book: Steel and Ricardo. Steel is listed as the editor and Ricardo is listed as the author. The contributor info for Steel and Ricardo in the target book is identical to the info from the source book.

## Ideas
In terms of what we need for contributors, we should look at https://github.com/pressbooks/pressbooks/blob/d712117757f36503505e61e85a82fa13acdc980b/inc/cloner/class-cloner.php#L422-L428

[...]

PR #2394, opened 2021-09-25
Add contributors metadata info in section metadata API endpoints

richard015ar:
Related issue: https://github.com/pressbooks/pressbooks/issues/2358
Related change: https://github.com/pressbooks/pressbooks/pull/2369

This PR adds contributors metadata in the API in the sections (Chapters, Back Matters, Front Matters, etc), not only for editors and other types of contributors, but also for authors.

### Testing example
_Note: Do not clone a book from integration until staging contains this version of the API. If you clone a book from staging without this new version, then the contributors information won't be exposed at the metadata section level. Instead, use a local version or use production to test compatibility with latest stable versions._

Here's a video about how to test and check the API endpoints, specially for the cloning process:

https://user-images.githubusercontent.com/13248424/134782513-104c87c7-fffb-4f34-a893-336b0ba9f5d0.mp4

**Important note: Make sure cleaned the wp cache**

### DIFF
--- a/inc/class-book.php
+++ b/inc/class-book.php
@@ -77,6 +77,8 @@ class Book {
 	 * Returns book information in a useful, string only, format. Data is converted to HTML.
 	 *
 	 * @param int $id The book ID.
+	 * @param bool $contributors_as_string Read contributors list as a string.
+	 * @param int $read_contributors_from_cache Read contributors from cache, if book information is stored in wp cache.
 	 *
 	 * @return array
 	 */
--- a/inc/class-book.php
+++ b/inc/class-book.php
@@ -104,7 +104,7 @@ class Book {
 		// -----------------------------------------------------------------------------
 
 		$cache_id = "book-inf-$blog_id";
-		if (  static::useCache() ) {
+		if ( static::useCache() ) {
 			$cached_book_information = wp_cache_get( $cache_id, 'pb' );
 			if ( $cached_book_information ) {
 				if ( ! $read_contributors_from_cache ) {
@@ -124,10 +124,13 @@ class Book {
 		if ( $meta_post ) {
 
 			// Contributors
-			$book_information = array_merge( $book_information, $contributors->getAll(
-				$meta_post->ID,
-				$contributors_as_string,
-				true )
+			$book_information = array_merge(
+				$book_information,
+				$contributors->getAll(
+					$meta_post->ID,
+					$contributors_as_string,
+					true
+				)
 			);
 
 			// Post Meta
--- a/inc/class-book.php
+++ b/inc/class-book.php
@@ -80,25 +80,13 @@ class Book {
 	 *
 	 * @return array
 	 */
-	static function getBookInformation( $id = null, $contributors_as_string = true ) {
+	static function getBookInformation( $id = null, $contributors_as_string = true, $read_contributors_from_cache = true ) {
 
 		if ( ! empty( $id ) && is_int( $id ) ) {
 			$blog_id = $id;
 			switch_to_blog( $blog_id );
 		} else {
 			global $blog_id;
-		}
-
-		// -----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


SteelWagstaff (changes requested):
@ricardopressbooks Could the issue be because we're defining $contributor_type too narrowly (as 'pb_authors' only) in the new insert function? https://github.com/pressbooks/pressbooks/blob/53712aa75577dff5d1da3e646841bb7d026df4db/inc/class-contributors.php#L231

richard015ar:
@SteelWagstaff I found the issue. The cache in [the method that brings the book metadata](https://github.com/pressbooks/pressbooks/blob/975c8b900b211f93ab20b233cbee1b1fba874f89/inc/class-book.php#L107), as we knew once the book information is loaded, it is cached. We read the information multiple times when cloning a book. Sometimes we just need basic information, and sometimes we need more information like contributors metadata. Once the information was read using different methods, included others not used for the cloning routine, which does not requires the contributors metadata, is not possible to read this info, because it was cached.

I added an extra parameter to specify those method we need to read the contributors metadata, even if the rest of the book information was cached. It fixes the problem.

I left a couple of comments in the PR to explain this better.

Regarding the lines you mentioned, those are [default parameters](https://www.php.net/manual/en/functions.arguments.php#functions.arguments.default), those are used when any parameter is sent to the method, which for the cloning routine is not the case.

inc/class-book.php
```diff
@@ -80,7 +80,7 @@ static function isBook() {
 	 *
 	 * @return array
 	 */
-	static function getBookInformation( $id = null, $contributors_as_string = true ) {
+	static function getBookInformation( $id = null, $contributors_as_string = true, $read_contributors_from_cache = true ) {
```
richard015ar, line 85: This extra parameter add the flexibility to read or not contributors from cache. If `contributors_as_string` was true from a previous call in this method, then the book information was cached and the contributors information was stored with basic info: only `name`.
With this change, if we set `$read_contributors_from_cache = false` allows dynamically read the contributors from the database, even if the contributors were read as a string and stored in the wp cache.

inc/class-book.php
```diff
@@ -89,34 +89,49 @@ static function getBookInformation( $id = null, $contributors_as_string = true )
...
-				$book_information[ $key ] = $val;
-			};
+			$book_information = array_merge(
+				$book_information,
+				$contributors->getAll(
+					$meta_post->ID,
+					$contributors_as_string,
+					true
+				)
+			);
```
richard015ar, line 136: It is a simpler way to do what we did previously to store the contributors information in `$book_information`

inc/cloner/class-cloner.php
```diff
@@ -1322,7 +1322,7 @@ protected function cloneMetadata() {
 						$contributor_data['slug'] = sanitize_title_with_dashes( remove_accents( $contributor_data['name'] ), '', 'save' );
 					}
 					$this->contributors->insert( $contributor_data, $metadata_post_id, $key, $this->downloads, 'slug' );
-					if ( $key === 'pb_authors' && isset( $contributor_data['slug'] ) ) {
```
richard015ar, line 1325: This 2nd condition is redundant, it was performed a few lines before.

inc/metadata/namespace.php
```diff
@@ -975,7 +975,7 @@ function add_json_ld_metadata() {
 	if ( $context === 'section' ) {
 		global $post;
 		$section_information = get_section_information( $post->ID );
-		$book_information = Book::getBookInformation( null, false );
+		$book_information = Book::getBookInformation( null, false, false );
```
richard015ar, line 978: Adding additional parameter where we want to make sure we don't want to read contributors as a string stored in the cache if some other method or action did it. It will force to read the contributors from the database instead.

SteelWagstaff (approved):
LGTM

Issue #2368, opened 2021-09-15
Handle Duplicate Contributors During API Import Process


SteelWagstaff:
## Description
When we perform an API import into an existing book, there is the possibility that imported content will have multiple contributors assigned as authors and that these contributors will have the same slugs as contributors which already exist in the target book. We need to handle these in a non-destructive way.

Desired outcome: when we import front/back matter & chapters, we should check slugs for chapter authors against the existing contributor list. If there are duplicates, we should disambiguate (slug-2, slug-3) and update the slug references for the authors associated with imported content to point to the newly created contributors (this can probably share a function with what we're need to do to address the same issue when we perform XML imports: https://github.com/pressbooks/pressbooks/issues/2354).

PR #2395, opened 2021-09-27
Handle contributors during API imports

arzola:
This PR handle #2368 

### How to test

1. Import some chapters with attached authors in a book that has contributors with the same slug but a different Name
2. Review if the imported chapter has attached the right contributors

### DIFF
--- a/inc/cloner/class-cloner.php
+++ b/inc/cloner/class-cloner.php
@@ -231,6 +231,12 @@ class Cloner {
 	protected $knownH5P = [];
 
 	/**
+	 * Flag to indicate the cloner is being used to import content
+	 * @var bool
+	 */
+	public $isImporting = false;
+
+	/**
 	 * Constructor.
 	 *
 	 * @since 4.1.0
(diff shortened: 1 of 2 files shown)


SteelWagstaff:
@arzola Problem noted in testing:
1. When I import a chapter which has an author with the same slug but a different name as a contributor in the target book, this contributor is not imported and the clone book links to the already existing contributor.
2. When I import a chapter which has an author with a different slug but the same name as a contributor in the target book, this contributor is not imported and the clone book links to the already existing contributor.
In both cases, a new contributor should be created (in case 1, the contributor should have a different slug (i.e. name-1), and the imported chapter should link to the newly created contributor.

richard015ar (approved):
Changes LGTM
I tested 2 cases mentioned by Steel, the case mentioned in the description and the cloning process to confirm it works as before.
See:


https://user-images.githubusercontent.com/13248424/135157097-d1554e56-4353-49da-9bb6-1fb9dd1346b2.mp4

https://user-images.githubusercontent.com/13248424/135157087-387a2e44-b285-4b53-9f2b-689fb4c94a82.mp4

https://user-images.githubusercontent.com/13248424/135157082-e0c5f5b2-ce57-45e0-bca0-0ac65776dd9c.mp4

PR #2399, opened 2021-09-27
fix: handle Twitter oEmbeds for EPUB and PDF

greatislander:
See https://github.com/pressbooks/pressbooks/issues/2371.

This PR introduces a conditional to replace the Twitter oEmbed source with a simple link to the tweet or timeline.


greatislander:
Opened in wrong repository.

Issue #2404, opened 2021-09-28
About the author section always starts on a new page in PDF export. Should start on same page as content previous to it.


SteelWagstaff:
![Screenshot from 2021-09-24 20-23-41](https://user-images.githubusercontent.com/13485451/134756268-80ccdd3d-c033-4f87-8868-8214520c06ea.png)

SteelWagstaff:
Tested on integrations and working as expected.

PR #2405, opened 2021-09-28
Add about the authors section inside the post content

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2404

This PR adds the about the authors section inside the post content in the XHTML export routine, to consider the pagination.

### Testing
For test, enable the "Display about the authors" global option and export the a book with chapter, back matter and front matters with authors assigned.


SteelWagstaff (approved):
Tested locally. Works exactly as intended.

Issue #2406, opened 2021-09-28
Post meta associations break when a contributor's slug is edited


SteelWagstaff:
Contributors have slugs. When you associate a contributor as the author of a chapter or assign them a book-level role, we associate the slug with the post meta. If you assign a contributor to be the author of a chapter and then edit their slug, this association is broken. 

We should do one of the following:
1. if possible, create a method which updates all post meta when a contributor's slug is edited (find original slug and replace with new slug)
2. hide/remove the ability to edit contributor slugs from users
3. Make the slug field 'read only' on the edit contributors form

fdalcin:
We are setting the slug input as read only in Edit Contributor form.

PR #2412, opened 2021-09-29
Fix duplicated contributors for URL import routine

richard015ar:
This PR fixes 2 issues:
- When we have a target book with a set of contributors and we we are importing from a source book through the URL and the target book contains one or more of the same contributors slug that the source book, the contributors in the target book would be overwritten.
- If we would import a source book through the URL import routine and the source book contains multiple sections with the same contributor, after importing the target book would contain duplicated contributors.

This change also addresses:
- https://github.com/pressbooks/pressbooks/issues/2406 (we add read only property in the edit contributors input page using jQuery).
- Dependabot PR: https://github.com/pressbooks/pressbooks/pull/2377
- Dependabot PR: https://github.com/pressbooks/pressbooks/pull/2365


inc/class-contributors.php
```diff
@@ -201,7 +201,7 @@ public function getArray( $post_id, $contributor_type, $include_term_meta = fals
 						foreach ( $meta as $name ) {
 							$result = $this->insert( $name );
 							if ( $result !== false ) {
-								$added = $this->link( $result['term_id'], $post_id, $contributor_type );
+								$added = $this->link( $result, $post_id, $contributor_type );
```
richard015ar, line 204: It was a legacy output format for `insert` method

fdalcin (approved):
LGTM @richard015ar, thanks!

PR #2413, opened 2021-09-30
Expose new post metadata for parts

richard015ar:
This PR exposes the `pb_part_invisible` metadata part as `pb_part_invisible_string` to address this issue: https://github.com/pressbooks/pressbooks/pull/2366 but keep compatibility with previous API versions.

### DIFF
--- a/inc/api/endpoints/controller/class-posts.php
+++ b/inc/api/endpoints/controller/class-posts.php
@@ -94,6 +94,11 @@ class Posts extends \WP_REST_Posts_Controller {
 			}
 		}
 
+		if ( $post->post_type === 'part' ) {
+			$response->data['meta']['pb_part_invisible_string'] = is_null( $response->data['meta']['pb_part_invisible'] ) ?
+				'on' : '';
+		}
+
 		return $response;
 	}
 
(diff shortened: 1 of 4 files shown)


arzola (approved):
LGTM I feel we should add more tests to increase the coverage but is good enough for now

greatislander:
FYI, this does appear to cause some issues— it's causing the PB CLI's clone test to fail because the `is_null` check is run against an array key which may not be set here: https://github.com/pressbooks/pressbooks/blob/423a0e24df3f28c8b8c66687628831a551a42a48/inc/api/endpoints/controller/class-posts.php#L97

I think this may be fixable by modifying [line 96](https://github.com/pressbooks/pressbooks/blob/423a0e24df3f28c8b8c66687628831a551a42a48/inc/api/endpoints/controller/class-posts.php#L96) as follows:

```diff
- if ( $post->post_type === 'part' && isset( $response->data['meta'] ) ) {
+ if ( $post->post_type === 'part' && isset( $response->data['meta']['pb_part_invisible'] ) ) {
```

I'll try to open a PR for this shortly.

PR #2416, opened 2021-09-30
Update version to next minor dev

ho-man-chan:


SteelWagstaff (approved):
:+1:

SteelWagstaff (changes requested):
I think we need to bump the version in pressbooks.php, no?

PR #2419, opened 2021-10-02
Check contributors cached format in bookInformation method

richard015ar:
This PR fixes the getBookInformation method in book class to always read from the cache or database in the needed format (string or array). It also checks meta field in overrideResponse method to add the pb_invisible_part_string when needed and fixes broken tests.

### DIFF
--- a/inc/api/endpoints/controller/class-posts.php
+++ b/inc/api/endpoints/controller/class-posts.php
@@ -94,7 +94,7 @@ class Posts extends \WP_REST_Posts_Controller {
 			}
 		}
 
-		if ( $post->post_type === 'part' ) {
+		if ( $post->post_type === 'part' && isset( $response->data['meta'] ) ) {
 			$response->data['meta']['pb_part_invisible_string'] = is_null( $response->data['meta']['pb_part_invisible'] ) ?
 				'on' : '';
 		}
--- a/inc/class-book.php
+++ b/inc/class-book.php
@@ -109,7 +109,18 @@ class Book {
 		if ( static::useCache() ) {
 			$cached_book_information = wp_cache_get( $cache_id, 'pb' );
 			if ( $cached_book_information ) {
-				if ( ! $read_contributors_from_cache ) {
+				$contributors_cached_as_string = true;
+				foreach ( $contributors->valid as $contributor_type ) {
+					if (
+						isset( $cached_book_information[ $contributor_type ] ) &&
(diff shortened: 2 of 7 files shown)


inc/class-book.php
```diff
@@ -109,7 +109,16 @@ static function getBookInformation( $id = null, $contributors_as_string = true,
...
-				if ( ! $read_contributors_from_cache ) {
+				$contributors_cached_as_string = true;
+				foreach ( $contributors->valid as $contributor_type ) {
+					if (
+						isset( $cached_book_information[ $contributor_type ] )
+					) {
+						$contributors_cached_as_string = is_string( $cached_book_information[ $contributor_type ] );
+						break;
+					}
+				}
```
richard015ar, line 120: As contributor types are dynamic and sometimes some of those those could be absents in different books, we need to check in all types. It will return at the first existing contributor type stored in the cache.

inc/api/endpoints/controller/class-posts.php
```diff
@@ -94,7 +94,7 @@ public function overrideResponse( $response, $post, $request ) {
 			}
 		}
 
-		if ( $post->post_type === 'part' ) {
+		if ( $post->post_type === 'part' && isset( $response->data['meta'] ) ) {
```
richard015ar, line 97: Sometimes we don't have the `meta` key in the API response, for example in the `toc` endpoint in the part section `meta` is not exposed.

inc/class-book.php
```diff
@@ -109,7 +109,14 @@ static function getBookInformation( $id = null, $contributors_as_string = true,
...
 			if ( $cached_book_information ) {
-				if ( ! $read_contributors_from_cache ) {
+				$contributors_cached_as_string = true;
+				foreach ( $contributors->valid as $contributor_type ) {
+					if ( isset( $cached_book_information[ $contributor_type ] ) ) {
+						$contributors_cached_as_string = is_string( $cached_book_information[ $contributor_type ] );
+						break;
+					}
+				}
+				if ( ! $read_contributors_from_cache || $contributors_cached_as_string !== $contributors_as_string ) {
```
richard015ar, line 119: We need to make sure the format we are requesting as a parameter is the same format we stored in the cache. If those are different, we'll read from the database anyway to return the requested format.

SteelWagstaff (approved):
:+1:

PR #2462, opened 2021-10-26

ho-man-chan:


omshivaprakash:
@SteelWagstaff @ho-man-chan I contributed to most of the projects for Kannada (kn) in transifex. Could you please import this language too?

Issue #2470, opened 2021-10-28
[Hackathon] - Fix copyright, pre/ul styling and EPUB validation issues


richard015ar:
- [x] message prince in copyright
- [x] Validation for EPUB (Talk to Steel)
- [x] `<pre>` and `<ul>` elements are printed different in PDF

richard015ar:
Copyright is rendered as expected, only difference is the text
`This book was produced with Pressbooks (https://pressbooks.com) and
rendered with Prince. `
which is enabled when certain global variables are present, see: https://github.com/pressbooks/pressbooks/blob/38b69cb3b9ec488dd2ea0202e35c7f929ec1519e/inc/modules/export/prince/class-pdf.php#L277

I used `dev` branch locally to make sure it is displayed in the same way than before.
For integrations probably we don't have that variable enabled.

richard015ar:
Validation issues:
I cloned this book: https://integrations.pressbooks.network/jctestbook/

Exporting this book on integration using the current dev branch, has validation issues.
Using dev branch on my local environment I see the validation issues are:
```
2021/11/01 19:26:32 [error] 1000#1000: *571 FastCGI sent in stderr: "PHP message: Pressbooks\Modules\Export\Epub\Epub
Array
(
    [time] => Mon Nov  1 19:26:32 2021
    [user] => admin
    [site_url] => https://pressbooks.test/jctestclonedintegration
    [blog_id] => 245
    [theme] => McLuhan
)
FATAL(RSC-016): /srv/www/pressbooks.test/current/web/app/uploads/sites/245/pressbooks/exports/jctestint-1635794782.epub/EPUB/chapter-001-headings.xhtml(19,8880): Fatal Error while parsing file: The entity "nbsp" was referenced, but not declared.
ERROR(RSC-005): /srv/www/pressbooks.test/current/web/app/uploads/sites/245/pressbooks/exports/jctestint-1635794782.epub/EPUB/chapter-001-headings.xhtml(-1,-1): Error while parsing file: The entity "nbsp" was referenced, but not declared.
ERROR(RSC-005): /srv/www/pressbooks.test/current/web/app/uploads/sites/245/pressbooks/exports/jctestint-1635794782.epub/EPUB/chapter-002-paragraph-formatting.xhtml(17,1579): Error while parsing file: value of attribute "src" is invalid; must be a URI
```
Same errors when using the exportHackathon-friday branch. So I assumed we don't introduced any additional issues in the validation process.

[...]

Issue #2448, opened 2021-10-21
[Hackathon] - Fix tests for Exports Hackathon Branch & update Miro Diagram


SteelWagstaff:
- [x] Write new tests and get old tests working for the Exports Hackathon Branch. (@richard015ar)
- [x] Create basic diagram in Miro with the structure of the export routine. (@arzola)

**Import note: the second item was added after the estimation**

PR #2474, opened 2021-10-29
Export hackathon tests

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2448

This PR adds unit tests for the new blade templates end export helpers class.

**Note: We have an validation issue for EPUB export routine, which will be addressed in: https://github.com/pressbooks/pressbooks/issues/2470. So, a few tests related to export fail, not because the implementation but because validation issues.**


inc/modules/export/xhtml/class-xhtml11.php
```diff
@@ -1517,7 +1517,7 @@ protected function echoBackMatterGenerator( $book_contents, $metadata ) : \Gener
 			] );
 
 			echo $this->blade->render( 'export/generic-post-type', $data );
-			
+
```
richard015ar, line 1520: It was a linting auto-fix. For some reason PHPCS complains about this.

templates/export/generic-post-type.blade.php
```diff
@@ -16,9 +16,9 @@
 		{!! $append_post_content !!}
 	@endif
 	@if( isset( $endnotes ) )
-		{{ $endnotes }}
+		{!! $endnotes !!}
```
richard015ar, line 19: We can have footnotes and endnotes with paragraph, images, and other html elements.

tests/test-modules-export-export.php
```diff
@@ -448,7 +448,8 @@ public function test_sanityCheckXhtmlWithoutBuckram() {
 		$this->assertContains( '<p><em>Ka kite ano!</em></p>', $xhtml_content );
 		$this->assertContains( 'https://github.com/pressbooks/pressbooks', $xhtml_content );
 		// Heading elements should be in a "bad" place.
-		$this->assertContains( '</h2></div><div class="ugc chapter-ugc"><h2 class="chapter-subtitle">Or, A Chapter to Test</h2>', $xhtml_content );
```
richard015ar, line 451: We check structure in blade template testing.

fdalcin (changes requested):
Really like your approach to the new tests @richard015ar.

However, I'm noticing a couple of things: 
I think you've accidentally deleted `tests/data/log.csv`. I left a couple of comments for the tests that fail. 

We have a tests that fails for EPUB validation that because it's rendering the slug in a bad way. We can ignore this failing test for now because this is fixed in another branch and should be fine once both are merged.

fdalcin (approved):

fdalcin (approved):
LGMT, thanks @richard015ar

Issue #2471, opened 2021-10-28
[Hackathon] - Export parts, chapter title


richard015ar:
- [x] Parts are not being exported properly. Export a book on integration with: empty parts, invisible/visible parts, with/without chapters and compare with the current implementation.
- [x] Chapter titles (search by `chapter-title` class) should be an h2 instread a h1 in `chapter.blade.php` template. In the near future we would need to create dynamic headers.

PR #2478, opened 2021-11-01
fix: improve parts and chapters conditional

fdalcin:
Fix for #2471 

This PR aims to fix parts not being rendered properly when they had empty content + chapters or content and no chapters. It also changes the book title heading to `h2` instead of `h1` for now, fixing the title breaking on exports.

The PR also fixes some EPUB issues where front-matters and back-matters were not being properly referenced in the book, causing validation issues and internal links not working as expected.

### How to test

1. Clone a book from integrations
2. Make sure settings for the original one and the clone one are the same
3. Export the original one form integrations
4. Export the clone one and compare them

### DIFF
--- a/inc/modules/export/class-exporthelpers.php
+++ b/inc/modules/export/class-exporthelpers.php
@@ -28,6 +28,7 @@ trait ExportHelpers {
 		$needs_tidy_html = $options['needs_tidy_html'] ?? false;
 		$endnotes = $options['endnotes'] ?? false;
 		$footnotes = $options['footnotes'] ?? false;
+		$slug_as_href = $options['slug_as_href'] ?? false;
 
 		$data = [
 			'id' => $post_data['ID'],
@@ -36,7 +37,7 @@ trait ExportHelpers {
 		$method = studly_case( $post_type_identifier );
 		$taxonomy_method = "get{$method}Type";
 		$data['subclass'] = $this->taxonomy->{$taxonomy_method}( $post_data['ID'] );
-		$data['slug'] = "{$data['post_type_class']}-{$post_data['post_name']}";
+		$data['slug'] = $slug_as_href ? $post_data['post_name'] : "{$data['post_type_class']}-{$post_data['post_name']}";
 		$data['title'] = ( get_post_meta( $post_data['ID'], 'pb_show_title', true ) ? $post_data['post_title'] : '<span class="display-none">' . $post_data['post_title'] . '</span>' ); // Preserve auto-indexing in Prince using hidden span
 		$data['content'] = $post_data['post_content'];
 		$data['append_post_content'] = apply_filters( "pb_append_{$post_type_identifier}_content", '', $post_data['ID'] );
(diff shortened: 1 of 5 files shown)


inc/modules/export/class-exporthelpers.php
```diff
@@ -28,15 +28,16 @@ public function mapBookDataAndContent( $post_data, $metadata, $post_number, $opt
...
 		$data = [
 			'id' => $post_data['ID'],
 		];
-		$data['post_type_class'] = str_replace( '_','-', $post_type_identifier ); // This class is used to map with the SCSS class in buckram Ex: front-matter
+		$data['post_type_class'] = str_replace( '_', '-', $post_type_identifier ); // This class is used to map with the SCSS class in buckram Ex: front-matter
 		$method = studly_case( $post_type_identifier );
 		$taxonomy_method = "get{$method}Type";
 		$data['subclass'] = $this->taxonomy->{$taxonomy_method}( $post_data['ID'] );
-		$data['slug'] = "{$data['post_type_class']}-{$post_data['post_name']}";
+		$data['slug'] = $slug_as_href ? $post_data['post_name'] : "{$data['post_type_class']}-{$post_data['post_name']}";
```
fdalcin, line 40: Checks to see how to handle `slugs` since we use them for referencing internal links in EPUBs

inc/modules/export/epub/class-epub.php
```diff
@@ -1582,6 +1582,7 @@ protected function renderFrontMatterGenerator( $book_contents, $metadata ) : \Ge
 			$data = $this->mapBookDataAndContent( $front_matter, $metadata, $i, [
 				'type' => 'front_matter',
 				'needs_tidy_html' => true,
+				'slug_as_href' => true, // we want the slugs to be proper anchors in the EPUB export
```
fdalcin, line 1585: New optional parameter to render slugs properly while exporting EPUBs

tests/test-templates-export.php
```diff
@@ -85,7 +85,7 @@ public function test_chapter() {
 			$chapter_rendered
 		);
 		$this->assertContains( '<p class="chapter-number">1</p>', $chapter_rendered );
-		$this->assertContains( "<h1 class=\"chapter-title\">$title</h1>", $chapter_rendered );
+		$this->assertContains( "<h2 class=\"chapter-title\">$title</h2>", $chapter_rendered );
```
richard015ar, line 88: @fdalcin this is a friendly commit to adapt the unit tests to the new changes since the tests were not present at the moment of this PR was created

richard015ar (approved):
LGTM! thank you @fdalcin!
My tests were:
- I cloned a book from integrations
- I added locally (and on integrations) part without content and chapters
- I added a part with content and without chapters
- I added a part with chapters but without content
- I made sure both contains same relevant settings
- I exported to PDF & EPUB both books in each environment 
- I compared both exported files and both looks as expected (differences found are unrelated to this change).

Issue #2401, opened 2021-09-28
Contributor links in the about the author section have an extra `<br>` before the link content in EPUB exports


SteelWagstaff:
![Screenshot from 2021-09-24 20-21-11](https://user-images.githubusercontent.com/13485451/134756211-e2313764-37f7-4733-866a-a708fbbd7562.png)
![Screenshot from 2021-09-24 20-17-02](https://user-images.githubusercontent.com/13485451/134756218-4540c751-b51f-442d-a259-bcc42261ebff.png)

SteelWagstaff:
To solve this, avoid the WP autop function, like we do for back matter contributors. Ask Os for details if needed.

PR #2487, opened 2021-11-04
Fix breaklines for contributors template

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2401

This change fixes the breaklines displayed in the epub exported in the "About the authors" section at the end of the chapter.
The reason of this is the [wpautop filter](https://codex.wordpress.org/Plugin_API/Filter_Reference). Basically each new breakline in the template is translated as a `<br />` html tag in the blade template.
We could disabled the wpautop filter by `remove_filter('the_content', 'wpautop');`, but instead increment impact risks I just remove the breaklines in the template, which keeps the template still the same.

### For testing
- Choose a book with contributors.
- Make sure those contains Website URL, Twitter and other fields to be displayed in the EPUB.
- Turn on the "Display about the author" theme option to display the info at the end of each chapter. 
- Export the book to EPUB.
- Make sure the about the author information does not contain extra breaklines.
- Inspect the EPUB exported to make sure any extra `<br />` were added.

### DIFF
--- a/inc/modules/export/epub/class-epub.php
+++ b/inc/modules/export/epub/class-epub.php
@@ -595,7 +595,6 @@ class Epub extends ExportGenerator {
 		}
 		$content = apply_filters( 'the_export_content', $content );
 		$content = str_ireplace( [ '<b></b>', '<i></i>', '<strong></strong>', '<em></em>' ], '', $content );
-		$content = $this->fixAnnoyingCharacters( $content );
 		$content = $this->tidy( $content );
 		return $content;
 	}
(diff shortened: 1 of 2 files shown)


inc/modules/export/epub/class-epub.php
```diff
@@ -654,27 +653,6 @@ public function validateGenerator() : \Generator {
...
-
-		// EPUB specific
-
-		// Adobe Digital Editions has problems with exotic dashes, that is to say if this were 1999...
-		// TODO: Some users want this, others do not want this, make up your mind...
-		// $html = str_replace( array( '–', '&#8211;', '—', '&#8212;', '‑' ), '-', $html ); @codingStandardsIgnoreLine
-
-		return $html;
-	}
-
```
richard015ar, line 677: This method is unused and unnecessary.

Issue #2460, opened 2021-10-26
Replace loadTemplate with blade templates


arzola:
As part of the epub/html modernization we want to replace loadTemplate with blade templates to have the HTML templating unified

We are using the following approach

```
<?php
\Pressbooks\Utility\put_contents(
			$this->tmpDir . "/EPUB/$filename",
			$this->loadTemplate( $this->dir . '/templates/epub201/html.php', $vars )
		);
```

We want to replace by the following

```
<?php
\Pressbooks\Utility\put_contents(
			$this->tmpDir . "/EPUB/$filename",
			$this->blade->render( 'export/html', $vars )
		);
```

richard015ar:
Something we agreed in the dev team: it is a nice to have but not mandatory to achieve the sprint goals.

PR #2488, opened 2021-11-04
Replace load templates with blade template rendering

fdalcin:
Solves #2460 

This PR replaces all `loadTemplate` calls by using blade template rendering.
It also tries to clean up the code and add some better type hinting to the `class-epub`.

It mainly affects EPUB exports.

### How to test

Make sure that all EPUB files are generate as before. The ones that used `html` and the `toc`, `container`, `opf` and `ibooks`


inc/modules/export/class-exporthelpers.php
```diff
@@ -49,14 +48,15 @@ public function mapBookDataAndContent( $post_data, $metadata, $post_number, $opt
...
-		$data['title'] = ( get_post_meta( $post_data['ID'], 'pb_show_title', true ) ? $post_data['post_title'] : '<span class="display-none">' . $post_data['post_title'] . '</span>' ); // Preserve auto-indexing in Prince using hidden span
+		$data['title'] = get_post_meta( $post_data['ID'], 'pb_show_title', true ) ? $post_data['post_title'] : '<span class="display-none">' . $post_data['post_title'] . '</span>'; // Preserve auto-indexing in Prince using hidden span
 		$data['content'] = $post_data['post_content'];
 		$data['append_post_content'] = apply_filters( "pb_append_{$post_type_identifier}_content", '', $post_data['ID'] );
 		$data['short_title'] = trim( get_post_meta( $post_data['ID'], 'pb_short_title', true ) );
 		$section_license = $this->doSectionLevelLicense( $metadata, $post_data['ID'] );
 
 		if ( $needs_tidy_html ) {
-			$data['content'] = $this->kneadHtml( $data['content'], $post_type_identifier, $post_type_identifier );
+			$data['content'] = $this->kneadHtml( $data['content'], $post_type_identifier, $post_number );
```
fdalcin, line 58: We were passing the wrong parameter here

inc/modules/export/epub/class-epub.php
```diff
@@ -66,6 +66,27 @@ class Epub extends ExportGenerator {
...
+	 * @var string
+	 */
+	protected $epubDir;
+
+	/**
+	 * Assets directory.
+	 *
+	 * @var string
+	 */
+	protected $assetsDir;
```
fdalcin, line 88: New variables to simplify and avoid repetition

inc/modules/export/epub/class-epub.php
```diff
@@ -527,20 +546,18 @@ public function convertGenerator(): \Generator {
...
-
 		// We need to change global $id for shortcodes, the_content, ...
 		global $id;
 		$old_id = $id;
 
 		// Do root level structures first.
 		foreach ( $book_contents as $type => $struct ) {
 
-			if ( preg_match( '/^__/', $type ) ) {
+			if ( 0 === strpos( $type, '__' ) ) {
```
fdalcin, line 556: Regular expressions are kind of hard to read, once we move to PHP 8 this could be replaced by `str_starts_with`

inc/modules/export/epub/class-epub.php
```diff
@@ -844,30 +856,32 @@ protected function fetchAndSaveUniqueMedia( $url, $fullpath ) {
...
 	/**
 	 * Clean up content processed by HTML5 Parser, change it back into XHTML
 	 *
-	 * @param $html
+	 * @param string $html
 	 *
 	 * @return string
 	 */
-	protected function html5ToXhtml( $html ) {
+	protected function html5ToXhtml( string $html ): string {
```
fdalcin, line 870: I think that this method could be removed since we want to have a valid HTML5. Might be worth to check this in the future

inc/modules/export/epub/class-epub.php
```diff
@@ -2711,29 +2681,46 @@ static function hasDependencies() {
...
-			if ( file_exists( $new_path ) ) {
-				$path = $new_path;
-			}
+		if ( ! is_array( $data ) ) {
+			\Pressbooks\Utility\put_contents( $path, $data );
+
+			return;
 		}
 
-		return parent::loadTemplate( $path, $vars );
```
fdalcin, line 2736: The parent `loadTemplate` method still exists, I believe that it could be removed as well. It seems to be used only for EPUBs

inc/modules/export/epub/class-epub.php
arzola: I also swapped those because the regex that we are performing is less performant for a simple string lookup

inc/modules/export/epub/class-epub.php
```diff
@@ -585,7 +600,7 @@ protected function preProcessBookContents( $book_contents ) {
 	 *
 	 * @return string
 	 */
-	protected function preProcessPostContent( $content, $type = '', $post_id = 0 ) {
+	protected function preProcessPostContent( string $content, string $type = '', int $post_id = 0 ): string {
```
arzola, line 603: @fdalcin I'm not sure if we are ready to use typed properties and return values because those I remember are only PHP 7.4 and currently we are still supporting PHP 7.3

inc/modules/export/epub/class-epub.php
```diff
@@ -775,8 +788,7 @@ protected function tidy( $html ) {
 	 *
 	 * @return string|array
 	 */
-	protected function fetchAndSaveUniqueMedia( $url, $fullpath ) {
-
+	protected function fetchAndSaveUniqueMedia( string $url, string $fullpath ) {
```
richard015ar, line 791: @fdalcin in order to follow the proposed changes. Should we specify the type declaration for this method as well?

richard015ar (approved):
Thank you for the nice template improvement and the cleaning on the code. It looks good to me!

For testing I exported to epub multiple books cloned from integrations and compare the books exported to make sure both looks the same even when the configuration is changed in both places.
I also run the export tests to make sure everything works as before.

Issue #2490, opened 2021-11-05
[Hackathon - Export] - Pending tasks


richard015ar:
Pending tasks for new export routine (PDF  - EPUB):
- [x] Review and merge `exportHackathon-friday` branch to dev. Make sure tests, codecov and lint works.
- [x] Check [the section in the head (title section) generation](https://github.com/pressbooks/pressbooks/blob/exportHackathon-friday/inc/modules/export/class-exporthelpers.php#L53-L55) for EPUB. It seems to be used to identify the back matters during the import routine.
- [x] Review EPUB import routine to make sure it works as before even with the structure changes we applied.
- [x] Use exploratory testing with multiple books and compare those on integrations to make sure it is working as expected. Ask to @SteelWagstaff to help us to test there.

richard015ar:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pressbooks-Product-Backlog-60ae7d807f6e370010b3d5be/issues/pressbooks/pressbooks/2490?planning-poker) with ZenHub @arzola @fdalcin

richard015ar:
## EPUB import routine
### Testing notes
The experimental environment contains the latest `exportHackathon-friday` branch at the moment of these notes.
- I exported to EPUB a book on integrations: https://integrations.pressbooks.network/jctestbook/
- I imported that book on experimental env: https://pressbooks.test/importfromintegration
That worked as expected.

- I exported a book to EPUB on experimental env: https://pressbooks.test/jctestclonedintegration
- I imported that file on integrations: https://integrations.pressbooks.network/importfromlocal
That worked as expected.

### The files structure changed, but why the import routine still recognized the new structure?
The `META-INF` folder is still the same name in previous a new versions. Inside that folder we export a `container.xml` file which contains the full path for the `book.opf` file. 
Example of a experimental env exported book:
```xml
<container xmlns="urn:oasis:names:tc:opendocument:xmlns:container" version="1.0">
  <rootfiles>
    <rootfile full-path="EPUB/book.opf" media-type="application/oebps-package+xml"/>
  </rootfiles>
</container>
```
And for integrations:
```xml
<container xmlns="urn:oasis:names:tc:opendocument:xmlns:container" version="1.0">
  <rootfiles>
    <rootfile full-path="book.opf" media-type="application/oebps-package+xml"/>
  </rootfiles>
</container>
```
[...]

PR #2494, opened 2021-11-08
Display title in EPUB

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2490 

This change avoid an empty title in the EPUB exported file in the `<head>` section for back matters, when `pb_show_title = false`.

Note: the `<span class="display-none">` wrap is for PDF and it is already specified in multiple places when needed, like: https://github.com/pressbooks/pressbooks/blob/dev/inc/modules/export/xhtml/class-xhtml11.php#L1335

### Testing
- Select a book a in the organize section disable `Show title` column for one or more back matters available.
- Export the book to EPUB
- Inspect the EPUB back matter exported and make sure the title is present in `<head>` section, but it is not in the content.
- Import the book exporter (Tools > Import > EPUB) and make sure the back matter title is displayed properly in the first step of the import routine.
- Make sure the `show title` settings also works for front matters and chapters.
- Combine: import / export from/to integrations to make sure there are compatibility regarding to the show title configuration.

### DIFF
--- a/inc/image/namespace.php
+++ b/inc/image/namespace.php
@@ -130,7 +130,7 @@ function is_valid_image( $file, $filename, $is_stream = false ) {
 		$file = $tmpfile_meta['uri'];
 	}
 
-	$type = @exif_imagetype( $file ); // @codingStandardsIgnoreLine
+	$type = IMAGETYPE_JPEG; // @codingStandardsIgnoreLine
 	if ( IMAGETYPE_JPEG === $type || IMAGETYPE_GIF === $type || IMAGETYPE_PNG === $type ) {
 		return true;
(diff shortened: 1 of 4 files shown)


fdalcin (approved):
LGTM, thanks @richard015ar.

Tested setting show title to false on from front-matter, chapter and back-matter and it works as expected. 
**The `<head>` contains the proper title, while the title itself is not rendered.**

Then, I selected to import the EPUB file to a new book and the titles are being displayed correctly in the page to select the content to be imported. Also after importing, the book title is correctly imported.

Testing importing the new EPUB version to integration also worked as expected.

PR #2504, opened 2021-11-11
chore: remove phpunit dependency

fdalcin:
This PR removes `phpunit/phpunit` direct dependency and rely only on `yoast/phpunit-polyfills`.


arzola (approved):
LGTM thanks Dalcin

PR #2508, opened 2021-11-15
feature: add anchors for interactive elements

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2287

This PR unifies the export (PDF & EPUB) templates for:
- oembed
- h5p
- video
- audios
interactive contents.

It also adds anchor links to the mentioned elements to be able to specify those in the templates for export.

### Testing
- Pick a book and add interactive contents (one or more of the mentioned above)
- Export it to PDF and EPUB and make sure the file exported contain the expected links with anchors. _Note: for EPUB 3 audios and another interactive contents are available for interaction._


inc/interactive/class-content.php
```diff
@@ -73,7 +83,7 @@ static public function hooks( Content $obj ) {
 		// @see https://codex.wordpress.org/Embeds/
 		add_action( 'init', [ $obj, 'registerEmbedHandlers' ] );
 		add_filter( 'oembed_providers', [ $obj, 'addExtraOembedProviders' ] );
-		add_filter( 'oembed_result', [ $obj, 'adjustOembeds' ], 10, 3 );
+		add_filter( 'embed_oembed_html', [ $obj, 'adjustOembeds' ], 10, 3 );
```
richard015ar, line 86: From: https://wordpress.stackexchange.com/questions/353313/combine-embed-oembed-html-and-oembed-result
- embed_oembed_html filters the cached oEmbed HTML.
- oembed_result filters the HTML returned by the oEmbed provider.
So oembed_result is when the html is first retrieved, before caching. embed_oembed_html is when the cache has been fetched but before rendering.

inc/interactive/class-content.php
```diff
@@ -263,16 +275,20 @@ public function replaceOembed( $return, $data, $url ) {
...
 		$url = wp_get_shortlink( $id );
+		if ( isset( $this->iframes[ $id ] ) ) {
+			$this->iframes[ $id ] ++;
+		} else {
+			$this->iframes[ $id ] = 1;
+		}
+		$anchor_id = 'oembed-' . $this->iframes[ $id ];
 
 		$html = $this->blade->render(
-			'interactive.oembed', [
```
richard015ar, line 271: Provider displaying and thumbnail for the template was removed by request

inc/interactive/class-content.php
```diff
@@ -309,17 +323,20 @@ public function replaceInteractiveTags( $html ) {
...
 			// Replace
 			$elements = $dom->getElementsByTagName( $tag );
 			$element_number = $elements->length;
 			for ( $i = $elements->length; --$i >= 0; ) {
 				$element = $elements->item( $i );
 				$template = $this->blade->render(
-					"interactive.{$tag}", [
-						'title' => $this->getTitle( $id ),
-						'url' => wp_get_shortlink( $id ),
+					'interactive.media', [
```
richard015ar, line 335: **Note for when we want to group interactive content for export purposes:** 
It generates extra breaklines.
A method I tested and worked for deleting this:
```php
	/**
	 * Remove breaklines after interactive contents elements rendered.
	 *
	 * @param $dom
	 * @return mixed
	 */
	public function removeBreaklinesBetweenInteractiveContents( $dom ) {
		$nodes = $dom->getElementsByTagName( 'div' );
		for ( $i = 0; $i < $nodes->length; $i++ ) {
			$div_element = $nodes->item( $i );
			if ( strpos( $div_element->getAttribute( 'class' ), 'interactive-content' ) !== false ) {
				$interactive_element = $div_element;
				$next_element = $interactive_element->nextSibling;
				if (
					! is_null( $next_element ) &&
					! is_null( $next_element->nextSibling ) &&
					$next_element->nextSibling->nodeName === 'br'
				) {
					$next_element->nextSibling->parentNode->removeChild( $next_element->nextSibling );

				}
			}
		}
		return $dom;
	}
```

Process the whole content at the beginning would be a better solution.

templates/interactive/media.blade.php
```diff
@@ -0,0 +1,11 @@
+<div class="textbox interactive-content interactive-content--{{ $tag }}">
+	<span class="interactive-content__icon"></span>
+	<p>
+		{{ __( 'One or more interactive elements has been excluded from this version of the text. You can view them online here: ', 'pressbooks' ) }}
+		@if( isset( $id ) )
```
richard015ar, line 5: This variable is not present for excluded oembeded `iframe` tags.

inc/interactive/class-phet.php
```diff
@@ -47,17 +77,32 @@ public function registerEmbedHandlerForExport() {
...
+		$embed = $this->blade->render(
+			'interactive.media', [
+				'title' => get_the_title( $id ),
+				'url' => wp_get_shortlink( $id ),
+				'tag' => 'iframe',
+				'id' => 'iframe-phet-' . $this->iframes_counter[ $id ],
+			]
 		);
+		apply_filters( 'embed_' . self::EMBED_ID, $embed, $matches, $attr, $url, $rawattr );
+		return $embed;
```
fdalcin, line 105: Could we do this directly?
```php
return apply_filters( 'embed_' . self::EMBED_ID, $embed, $matches, $attr, $url, $rawattr );
```

fdalcin (changes requested):
Changes LGTM @richard015ar, thanks!

While testing I noticed a weird issue that only happens with PDF exports.

When exporting for PDF, if the URL is too long, the link url will break lines as expected. However, `oembed` links don't work when this happens (see the video below).
https://user-images.githubusercontent.com/1761690/142223481-c79104eb-8cd2-43d2-8160-17598dff2508.mov

inc/interactive/class-phet.php
```diff
@@ -30,14 +32,42 @@ public function registerEmbedHandlerForWeb() {
...
+	public function applyEmbedFilterForWeb( $matches, $attr, $url, $rawattr ) {
+		global $id; // This is the Post ID, [@see WP_Query::setup_postdata, ...]
+		$this->setIframeCounterByPost( $id );
+		$embed = sprintf(
+			'<iframe id="iframe-phet-%2$s" src="https://phet.colorado.edu/sims/html/%1$s" width="800" height="600" scrolling="no" allowfullscreen></iframe>',
+			esc_attr( $matches[1] ),
+			$this->iframes_counter[ $id ]
 		);
+		apply_filters( 'embed_' . self::EMBED_ID, $embed, $matches, $attr, $url, $rawattr );
+		return $embed;
```
fdalcin, line 57: Same thing as above

inc/interactive/class-phet.php
richard015ar: I added this return to be able to test the method, so far we don't need any return on this method but it will simplify our testing

fdalcin (approved):

fdalcin:
LGTM, thanks @richard015ar! 

As we discussed earlier, this issue happens already with long URLs -- [see](https://www.princexml.com/forum/topic/4147/when-a-link-wraps-across-pages-it-breaks). We will work on that as a separate issue.

PR #2525, opened 2021-11-24
ci: pull updated translation files on workflow dispatch

greatislander:
This is a dry run of the GitHub Action for updating translation files from Transifex (PO, MO). It will attach the updated files to the action as an artifact. It includes instructions on how to modify the workflow to automatically commit the changes to `dev`.

To run the action, follow the instructions here: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow


SteelWagstaff (approved):
:+1:

Issue #2467, opened 2021-10-27
Avoid empty fields in PB API


richard015ar:
### Description

After the new contributor formats changes in the API we began exposing empty arrays for each contributor type (reviewers, authors, editors, illustrators, etc) that is not present in the database. Here's an example of a book with authors and contributors, but without illustrators, reviewers, and others:
![Screenshot from 2021-10-27 11-04-45.png](https://images.zenhubusercontent.com/60a7e600d2c5ea0adb753a68/32e67e17-c257-43e1-b31d-0eaa17cb0a84)

When we don't have a specific contributor type, we should avoid exposing the keys.

### Steps to Reproduce

1. Select a book and go to Book Info. 
2. Add 1 or more authors and 1 or more contributor types (like illustrators, reviewers, editors, etc)
3. Left 1 or more contributor types empty
4. Go to `https://YOUR_NETWORK_URL/wp-json/pressbooks/v2/books` and search the selected book.
5. The book must expose empty arrays for those contributor types you left empty in Book Info section.

**Expected behavior:** Key with empty array values should not be exposed in the API.

**Actual behavior:** Empty arrays are exposed in the API for those contributor types that are not filled in Book Info.

**Note**: once the empty arrays are not exposed anymore we would need to revert https://github.com/pressbooks/book-directory-fetcher/pull/171.

PR #2526, opened 2021-11-24
fix: do not expose empty contributors keys in PB API

richard015ar:
Related issue: https://github.com/pressbooks/pressbooks/issues/2467

This change avoids to expose empty contributors keys based on contributors types (illustrators, editors, reviewers, etc) in the API.

### Test
For testing you can pick a public book and go to Book Info, and then add some contributors like editors, and illustrators, and leave some of them empties, like reviewers or contributors.
Then, visit: https://<YOUR_LOCAL_URL_ENV>/wp-json/pressbooks/v2/books
The JSON exposed search for the book you picked, the JSON object must display only the keys where the contributors types are stored.


arzola (commented):
I just reviewed it it looks good and it works as expected

I'm just wondering if it's fine that  the /wp-json/pressbooks/v2/metadata endpoint still shows the empty contributors

arzola (approved):
LGTM I think the metadata endpoint it could be fixed in another ticket

PR #2551, opened 2021-12-10
feat: Add method to sync network manager ids

fdalcin:
Fixes pressbooks/pressbooks-network-analytics#166. Accompanies pressbooks/pressbooks-network-analytics#168

This PR adds a new `updateNetworkManagers` to the user data collector which is responsible to sync network manager ids.

#### How to test

Follow instructions at pressbooks/pressbooks-network-analytics#168

### DIFF
--- a/inc/admin/networkmanagers/namespace.php
+++ b/inc/admin/networkmanagers/namespace.php
@@ -91,11 +91,8 @@ function update_admin_status() {
 
 		if ( is_array( $restricted ) && ! empty( $restricted ) ) {
 			update_site_option( 'pressbooks_network_managers', $restricted );
-			// This meta option will only be used by the BI, but we want to keep that in sync.
-			update_site_option( 'pressbooks_network_managers_ids', implode( ',', $restricted ) );
 		} else {
 			delete_site_option( 'pressbooks_network_managers' );
-			delete_site_option( 'pressbooks_network_managers_ids' );
 		}
 		// Reset the cheap cache after updating the option
 		_restricted_users( true );
--- a/inc/datacollector/class-user.php
+++ b/inc/datacollector/class-user.php
@@ -6,6 +6,8 @@
 
 namespace Pressbooks\DataCollector;
 
+use function Pressbooks\Admin\NetworkManagers\_restricted_users;
+
 class User {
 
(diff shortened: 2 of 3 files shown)


inc/admin/networkmanagers/namespace.php
```diff
@@ -91,11 +91,8 @@ function update_admin_status() {
 
 		if ( is_array( $restricted ) && ! empty( $restricted ) ) {
 			update_site_option( 'pressbooks_network_managers', $restricted );
-			// This meta option will only be used by the BI, but we want to keep that in sync.
-			update_site_option( 'pressbooks_network_managers_ids', implode( ',', $restricted ) );
```
fdalcin, line 95: I opted for removing the update directly here and rely only on the cron to update the value

inc/datacollector/class-user.php
```diff
@@ -95,6 +97,23 @@ public function updateAllUsersMetadata(): \Generator {
 		}
 	}
 
+	/**
+	 * Updates the network manager's site meta data
+	 *
+	 * @return void
+	 */
+	public function updateNetworkManagers(): void {
```
richard015ar, line 105: wondering if it should be in pressbooks network analytics plugin, since the network admins management is mostly there, additionally this method seems to be used only there. What do you think?

fdalcin: Could be an option, yes. I opted to add it here to keep it close to the other data collectors.

richard015ar: We are adding a new namespace with @arzola `\PressbooksNetworkAnalytics\DataCollector` for the new table and collection actions. Probably a new method or class in that namespace would be more centralized. We can keep it here in the meantime, move to a pressbooks network analytics plugin (probably a temporary namespace method) or wait until the new namespace is created in near future PRs. I would choose option 2, but it is up to you, for me any of those options are ok. What do you think guys? @arzola  @fdalcin ?

fdalcin: I'm ok with moving to network analytics once we have the new namespace. Keep in mind that user and book data collectors will continue to live here though.

arzola (approved):
LGTM

PR #2560, opened 2021-12-19
feat: conditionally render custom feed

fdalcin:
Closes pressbooks/private#633

This PR conditionally renders the custom feed. It will only render if `pressbooks-plugins-config` is not active or if is active and the feed options differs from the default ones.

TK

### DIFF
--- a/inc/admin/dashboard/namespace.php
+++ b/inc/admin/dashboard/namespace.php
@@ -23,6 +23,28 @@ function get_rss_defaults() {
 	];
 }
 
+function should_display_custom_feed() {
+	$default_options = get_rss_defaults();
+
+	$options = array_map(
+		'stripslashes_deep', get_site_option(
+			'pressbooks_dashboard_feed', $default_options
+		)
+	);
+
+	if ( ! $options['display_feed'] ) {
+		return false;
+	}
+
+	$plugins_config_is_active = is_plugin_active( 'pressbooks-plugins-config/pressbooks-plugins-config.php' );
+
+	if ( ! $plugins_config_is_active ) {
+		return true;
+	}
+
+	return trim( $options['url'], '/' ) !== trim( $default_options['url'], '/' );
+}
+
 /**
  *  Remove unwanted network Dashboard widgets, add our news feed.
  */
@@ -42,7 +64,8 @@ function replace_network_dashboard_widgets() {
 			'pressbooks_dashboard_feed', get_rss_defaults()
 		)
 	);
-	if ( ! empty( $options['display_feed'] ) ) {
+
+	if ( should_display_custom_feed() ) {
 		add_meta_box( 'pb_dashboard_widget_blog', $options['title'], __NAMESPACE__ . '\display_pressbooks_blog', 'dashboard-network', 'side', 'low' );
 	}
 }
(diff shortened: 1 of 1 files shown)


inc/admin/dashboard/namespace.php
```diff
@@ -23,6 +23,28 @@ function get_rss_defaults() {
...
+		return false;
+	}
+
+	$plugins_config_is_active = is_plugin_active( 'pressbooks-plugins-config/pressbooks-plugins-config.php' );
+
+	if ( ! $plugins_config_is_active ) {
+		return true;
+	}
+
+	return trim( $options['url'], '/' ) !== trim( $default_options['url'], '/' );
```
fdalcin, line 45: This could be improved to check against a list of URLs instead

inc/admin/dashboard/namespace.php
```diff
@@ -206,12 +229,12 @@ function pending_invitations_callback() {
...
-            <a class='button button-primary' href='" . home_url( '/newbloguser/' . $metadata['key'] ) . "'>" . __( 'Accept', 'pressbooks' ) . '</a>
-        </div>
-        <hr/>
-        ';
+		<div>
+			<p>$message</p>
+			<a class='button button-primary' href='" . home_url( '/newbloguser/' . $metadata['key'] ) . "'>" . __( 'Accept', 'pressbooks' ) . '</a>
+		</div>
+		<hr/>
+		';
```
fdalcin, line 237: Linting fix

SteelWagstaff (dismissed):

SteelWagstaff:
@fdalcin i think we should check for both https://pressbooks.com/feed and https://pressbooks.com/blog/feed, since both of those return the same RSS feed at present

inc/admin/dashboard/namespace.php
```diff
@@ -23,6 +23,30 @@ function get_rss_defaults() {
...
+
+	if ( ! $plugins_config_is_active ) {
+		return true;
+	}
+
+	$urls_to_skip = [
+		'https://pressbooks.com/feed',
+		'https://pressbooks.com/blog/feed',
+		'https://pressbooks.community/c/pressbooks-product-updates/19.rss',
+	];
```
fdalcin, line 45: Opted to check both URLs + the Pressbooks Product Update as well

fdalcin:
> @fdalcin i think we should check for both https://pressbooks.com/feed and https://pressbooks.com/blog/feed, since both of those return the same RSS feed at present

@SteelWagstaff I added a check for both and also checked against the Product Update one since we don't want to have duplicates

inc/admin/dashboard/namespace.php
```diff
@@ -23,6 +23,30 @@ function get_rss_defaults() {
...
+/**
+ * @param array $options
+ * @return bool
+ */
+function should_display_custom_feed( array $options ): bool {
+	if ( ! $options['display_feed'] ) {
+		return false;
+	}
+
+	$plugins_config_is_active = is_plugin_active( 'pressbooks-plugins-config/pressbooks-plugins-config.php' );
```
arzola, line 35: @fdalcin I think pressbooks core should not know that pressbooks-plugin-config exists I suggest to `hookify` this instead doing that check probably you could create an action hook in pb-config and check for the existance in pressbooks core

**pressbooks-plugin-config**
```
add_filter(
				'should_display_custom_feed', function() {
					return true;
				}
			);
```

**pressbooks**
```
if(has_filter( 'should_display_custom_feed' )) {

}
```

What do you think?

arzola (changes requested):

inc/admin/dashboard/namespace.php
fdalcin: I like the idea! Sent changes here and in pressbooks/pressbooks-plugins-config#16. Let me know what you think and if you have any suggestion

arzola (approved):
LGTM excellent job

PR #2587, opened 2022-01-12
feat: Add 'Institution' field to book info

### LINKED ISSUE #2407: Add 'Institution' field to book info
Part of https://github.com/pressbooks/ideas/issues/404

We should make the following changes in Pressbooks
- [x] Create a new Institution list in our symbionts folder (see ideas below -- but we can start with JSON file from Ontario)
- [x] Create a new book info field called ‘Institution’ which will allow book authors to select one or more of the institutions available in their network’s ‘Institutions’ list. Provide text to allow users to suggest a new institution
- [x] Add institution field to API and include in cloning routine 

## Notes
For the first item, we would prefer to have institutions function as a controlled vocabulary/taxonomy. I think the most ideal option would be to create/use a prebuilt controlled vocabulary/list of institutions (which would then function like Thema subject headings), though I don't know whether we could create/find such a list or whether we'd want to maintain this. One option might be https://univ.cc/? Alternatively, we could allow institutions to be be created/maintained at the network level only by super admins/network managers. I don't understand how/whether a network level taxonomy can be shared with books/sites, nor do I have a clear idea what would/should happen when a book was cloned. At its simplest, network managers could simply input several institutions, one per line, the way we allow them to maintain an iframe Allowlist:
Screenshot from 2022-01-05 19-18-45 
We could also simply let this be a free-entry field in book info, and treat each line as a separate entity? 

This is currently [blocked in Asana](https://app.asana.com/0/1199100224616909/1200482439883299) by "Revamp Aldine theme." See https://github.com/pressbooks/pressbooks-aldine/issues/242

fdalcin:
Closes #2407 

This PR adds a new optional multi select `Institutions` field to the book info page. It also exposes the new field to the `wp-json/pressbooks/v2/metadata` endpoint so the cloner is able clone the new information.

#### How to test

1. Make sure the new 'Institution(s)' field gets displayed in the Book Info section
2. Make sure you're able to save institutions
3. Make sure values are exposed properly (books with no institutions should not expose the information)
4. Make sure cloning routine works without issues and the institutions are cloned properly


inc/class-book.php
```diff
@@ -151,16 +150,21 @@ static function getBookInformation( $id = null, $contributors_as_string = true,
 			$expected_the_content = [ 'pb_custom_copyright', 'pb_about_unlimited' ];
 			$expected_url = [ 'pb_cover_image' ];
 			foreach ( get_post_meta( $meta_post->ID ) as $key => $val ) {
-
 				// Skip anything not prefixed with pb_
-				if ( ! preg_match( '/^pb_/', $key ) ) {
+				if ( 0 !== strpos( $key, 'pb_' ) ) {
```
fdalcin, line 154: This is a small improvement for readability.

inc/class-book.php
```diff
@@ -151,16 +150,21 @@ static function getBookInformation( $id = null, $contributors_as_string = true,
...
 				if ( $contributors->isValid( $key ) || $contributors->isDeprecated( $key ) ) {
 					continue;
 				}
 
+				if ( $key === 'pb_institutions' ) {
+					$book_information[ $key ] = $val;
+
+					continue;
+				}
+
```
fdalcin, line 167: We want to expose the value as an array instead of processing into a comma separated string

inc/cloner/class-cloner.php
```diff
@@ -1343,6 +1343,10 @@ protected function cloneMetadata() {
 				}
 			} elseif ( $key === 'pb_title' ) {
 				update_post_meta( $metadata_post_id, $key, $this->targetBookTitle );
+			} elseif ( $key === 'pb_institutions' ) {
+				foreach ( $value as $institution ) {
+					add_post_meta( $metadata_post_id, $key, $institution );
+				}
```
fdalcin, line 1349: Instead of checking for the key, we could maybe do something like this:
```php
 elseif ( is_array( $key ) ) {
    foreach ( $value as $institution ) {
        add_post_meta( $metadata_post_id, $key, $institution );
    }
}
```

inc/metadata/namespace.php
```diff
@@ -1197,3 +1218,20 @@ function check_thema_lang_file( $post ) {
...
+	$items = json_decode(
+		\Pressbooks\Utility\get_contents( PB_PLUGIN_DIR . 'symbionts/institutions/institutions.json' ), true
+	);
+
+	return array_reduce(
+		$items, static function ( $institutions, $institution ) {
+				return $institutions + [ $institution['name'] => $institution['name'] ];
+		}, []
+	);
+}
```
fdalcin, line 1237: New method to load the institutions json file

tests/test-metadata.php
```diff
@@ -105,19 +105,44 @@ public function test_has_expanded_metadata() {
 	 * @group metadata
 	 */
 	public function test_book_information_to_schema() {
```
fdalcin, line 107: I updated the old test to create an actual book with some metadata instead of passing a simple array

richard015ar: I think this is a better test in a bigger context. 
One thing to consider here is we assumed that `getBookInformation` method works as expected, which is outside the scope for the `book_information_to_schema` method. In some way the `book_information_to_schema` method/unit of code that we want to test is now dependent of the `getBookInformation` method/unit. So, now we have a higher level of unit test.
Checking the code, so far we always use `getBookInformation` and `book_information_to_schema` together when the 2nd is present, so for me this tests is still good.
One thing to consider is, if we in some point use the `book_information_to_schema` without `getBookInformation` or, if we use `getBookInformation` and process in some way the data before use `book_information_to_schema`, we would like to update this probably.

inc/cloner/class-cloner.php
richard015ar, line 1349: Dalcin, here a proposal. In order to avoid:
- repetition of https://github.com/pressbooks/pressbooks/pull/2587/files#diff-6917f7de9f5467b4f4fb047e37161b8a207f4430ddffaaf6086f8698d196f469R1341-R1343
- an extra ifelse non-tested

we could add the `pb_institutions` to `$metadata_array_values` in https://github.com/pressbooks/pressbooks/pull/2587/files#diff-6917f7de9f5467b4f4fb047e37161b8a207f4430ddffaaf6086f8698d196f469R1325, and
here: https://github.com/pressbooks/pressbooks/pull/2587/files#diff-6917f7de9f5467b4f4fb047e37161b8a207f4430ddffaaf6086f8698d196f469R1341-R1343
 we could add the extra condition like this:
```php
$values = $key !== 'pb_institutions' ? explode( ', ', $value ) : $value;
foreach ( $values as $v ) {
	add_post_meta( $metadata_post_id, $key, $v );
}
```

richard015ar:
@fdalcin thank you for the changes and the improvements in the code where you work/change. I appreciate you improve all the code around that you inspect and work.

I left a comment for a little improvement. I tested locally cloning books with insitutions and without instituions, and I also review the metadata edpoints.
For more exploratory testing I also deployed the change to experimental environment and using the current dev branch without institution field, I cloned a book from experimental with institutions to test the compatibility, and that worked good.

tests/test-metadata.php
richard015ar: btw this is a side note, I am not suggesting any changes on this.

inc/cloner/class-cloner.php
fdalcin: For now I think it makes sense to do this way. If we add extra information to the institution we'll likely need to change this.

richard015ar (approved):
Changes looks good to me! btw, great tests! thank you Dalcin.

Issue #2570, opened 2022-01-05
Bug cloning a book that contains an SVG image


tw77:
### Description

Using the cloning tool to clone a book (in-network) that contains an SVG image in its Media library, as seen with [this book](https://pressbooks.online.ucf.edu/strategies/), fails and shows this error:

<img width="942" alt="Screen Shot 2022-01-05 at 7 59 13 AM" src="https://user-images.githubusercontent.com/44182629/148248504-daa6dbf8-f4db-4bf4-b9f2-ee595739e673.png">

The SVG image in this book: https://pressbooks.online.ucf.edu/app/uploads/sites/483/2021/10/wc_icon_discussion.svg

Sentry link: https://sentry.io/organizations/pressbooks/issues/2483642948/?project=3954026&query=is%3Aunresolved

Relevant JSON data: https://pressbooks.online.ucf.edu/strategies/wp-json/wp/v2/media?page=4

### Steps to Reproduce

1. Navigate to Clone a Book on this network.
2. Set https://pressbooks.online.ucf.edu/strategies/ as the Source Book URL.
3. Attempt to clone.

**Expected behavior:** A clone of the book is created.

**Actual behavior:** Nothing happens and this EventStream Connection Error is shown.

SteelWagstaff:
To fix, try improving conditional here: https://github.com/pressbooks/pressbooks/blob/51dce1a0387545602e5d5c1bda461261a993bff4/inc/cloner/class-cloner.php#L756

add check to make sure item mime type is not 'image/svg+xml'
`if ( $item['media_type'] === 'image' && $item['mime_type'] !== 'image/svg+xml' )`

tw77:
Tested fix on Integrations and Textopress networks. I'm able to upload an SVG image to a book, insert the image in a chapter, save, and then clone the book with the cloning tool. The newly created book includes the SVG image.

Confirmed working with internally cloned books on Integrations and Textopress as well as a book cloned from Integrations to Textopress.

PR #2596, opened 2022-01-18
Improve media handling

fdalcin:
Issue #2570 

This PR checks for SVG images to handle them without the size option.

### How to test

1. Create a book with SVG images (they should be part of the book content, otherwise they'll not be cloned)
2. Clone the book and make sure the event stream error does not happen
3. Make sure SVG images are cloned and are present in the book content


SteelWagstaff (approved):
LGTM.

Issue #2576, opened 2022-01-06
Instruct the cloning routine to apply the source book’s theme and theme options and import any custom styles if the theme versions are identical


SteelWagstaff:
Instruct the cloning routine to apply the source book’s theme and theme options and import any custom styles if the theme versions are identical (maybe as part of postProcess?) https://github.com/pressbooks/pressbooks/blob/51dce1a0387545602e5d5c1bda461261a993bff4/inc/cloner/class-cloner.php#L495
* If successful, display the following message: "Cloning succeeded! Cloned ... and x glossary terms to BOOK TITLE. The source book's theme, theme settings, and custom styles were successfully applied." (bolded text is new):
* If the target book’s theme is not available on the source book’s network, apply the network’s default theme to the newly cloned book and display the following message: "Cloning succeeded! Cloned ... and x glossary terms to BOOK TITLE. The source book's theme, 'THEME NAME (version)', was not available on this network and could not be applied. Contact your network manager with questions about theme availability." (bolded text is new)
Cloning success message is generated here: https://github.com/pressbooks/pressbooks/blob/51dce1a0387545602e5d5c1bda461261a993bff4/inc/class-eventstreams.php#L200-L212

tw77:
Tests so far:

Cloning within Integrations works, bringing over themes, theme options and custom styles. 

Cloning from Integrations to Textopress (where the source theme is available) also works, bringing over theme, theme options and custom styles.

Cloning from Textopress to Integrations (where the source theme is not available) gets the expected result. We see the message about the theme not being available. The book clones without bringing over the theme, theme options or custom styles.

I see that the University network doesn’t have this new cloning feature yet, so when I clone a book from Integrations to University, the cloned book is created but the theme, theme options and custom styles are not applied.

I also notice that cloning a book _from_ University to Integrations doesn't result in the theme, theme options or custom styles being applied. This may be expected, but just to confirm, both networks -- the network with the source book, and the network where the clone is created -- appear to need to have this new routine set up in order for a clone to bring over the theme, theme options and custom styles from the source book.

tw77:
Confirmed the above.

When cloning a book with an older version of a theme to a network with a newer version of the theme, this is handled as an unavailable theme:

<img width="1074" alt="Screen Shot 2022-01-26 at 12 11 11 PM" src="https://user-images.githubusercontent.com/44182629/151240946-74406df6-4da1-4f33-b434-3a2513890ebc.png">

[...]

PR #2600, opened 2022-01-20
Pb 2576 Switch theme. Clone theme options and custom styles.

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2576

This PR reads [the theme and styles endpoints](https://github.com/pressbooks/pressbooks/pull/2597) to switch the theme of the target book during the cloning routine and clone the theme options and custom styles.

The test for cloning needs a refactor, since we use an external sample book for test the cloning routine.
Developers agreed on avoid unit tests for this feature this time.

### Testing case sample

https://user-images.githubusercontent.com/13248424/150425736-018d12de-a2b9-4db9-98e8-ff0e8a8faf17.mp4

Try also to clone books with previous API versions (without theme / styles options), and using experimental branch you could try to clone books with different theme versions.


inc/api/endpoints/controller/class-styles.php
```diff
@@ -12,7 +12,7 @@ class Styles extends \WP_REST_Controller {
 	/**
 	 * @var
 	 */
-	protected $rent_base;
+	protected $rest_base;
```
richard015ar, line 15: thanks @connerbw for reporting this!

fdalcin (approved):
Thank you @richard015ar, LGTM.

Here are my notes:

### Same network

Book is cloned with the right theme and with all theme options and custom styles properly applied

### Different network

#### Same theme available and active

Book is cloned with the right theme and with all theme options and custom styles properly applied.

#### Same theme available but not active

Book is cloned with the right theme and with all theme options and custom styles properly applied. However if the user changes the theme, it'll not be possible to set the original theme back since it's not network activated.

#### Same them not available

Book is clone with the default theme and no theme options and custom styles are copied.

Issue #2575, opened 2022-01-06
add information about the active theme (name & version), theme options, and Custom Stylesheets (web, ebook, PDF) to API


SteelWagstaff:
Create endpoints with information about the current theme & version, all existing theme options, and book custom stylesheets so that they are accessible for the cloning routine.

Custom stylesheets are created as custom post types of the type 'custom-style'. See https://github.com/pressbooks/pressbooks/blob/51dce1a0387545602e5d5c1bda461261a993bff4/inc/class-styles.php#L108-L178

SteelWagstaff:
Tested on integrations:

Two new API endpoints have been added to all public books:
<BOOK URL>/wp-json/pressbooks/v2/styles exposes custom styles for epub, web and prince format.
<BOOK URL>/wp-json/pressbooks/v2/theme exposes basic theme info and theme options (global, ebook, pdf and web).

For private books, the endpoints returns 'rest_forbidden', as expected.

The only question I have is why the two endpoints both feature a `@context": "http://schema.org"` statement at the top. None of the metadata on these pages fits an actual schema.org schema, so I don't think this is actually accurate. See the schema.org schema for `Book`, for example: https://schema.org/Book

richard015ar:
@SteelWagstaff I think those were a reference I created when I tried the 1 single metadata endpoint instead 2 different ones we decided later.
I removed the references, are you willing to review it in: https://github.com/pressbooks/pressbooks/pull/2606 ?

SteelWagstaff:
Tested and working as expected. Ready for release.

PR #2606, opened 2022-01-25
fix: remove schema.org reference for styles and theme API endpoints

richard015ar:
Reference: https://github.com/pressbooks/pressbooks/issues/2575#issuecomment-1020616411

This PR removes any references about schema.org for styles & theme API endpoints, since those are not a schema.org types.


SteelWagstaff (approved):
:+1:

PR #2608, opened 2022-01-25
feat: add default theme option to network settings

richard015ar:
Issue: https://github.com/pressbooks/pressbooks-network-analytics/issues/185

This change adds the Default theme option to Network Settings page.

### Test case sample
- Go to Network Admin > Settings > Network Settings section
- Make sure a new Theme Settings / Default theme option is defined in the form. A select with available themes must be rendered.
- Change the Default Theme and save it
- Create a new book and make sure the theme applied by default is the same you defined in the previous step.

### DIFF
--- a/inc/admin/network/class-networksettings.php
+++ b/inc/admin/network/class-networksettings.php
@@ -61,11 +61,12 @@ class NetworkSettings {
 		echo '<tr><th scope="row">' . __( 'Default Theme', 'pressbooks' ) . '</th><td>';
 		$options = '';
 		$themes = $GLOBALS['pressbooks']->allowedBookThemes( \WP_Theme::get_allowed_on_network() );
+		$default_theme = get_site_option( $this->customOptions['default_theme'], self::getDefaultTheme() );
 		foreach ( $themes as $theme => $_ ) {
 			$options .= sprintf(
 				'<option value="%1$s"%2$s>%3$s</option>',
 				$theme,
-				selected( get_site_option( $this->customOptions['default_theme'], self::DEFAULT_THEME ), $theme, false ),
+				selected( $default_theme, $theme, false ),
 				wp_get_theme( $theme )->get( 'Name' )
 			);
 		}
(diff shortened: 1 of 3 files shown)


inc/admin/network/class-networksettings.php
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * @var NetworkSettings
+	 */
+	private static $instance = null;
+
+	public function __construct() {
+		$this->customOptions = [
+			'default_theme' => self::DEFAULT_THEME_OPTION,
+		];
```
richard015ar, line 31: I created this property thinking in possible new fields we want to add

inc/admin/network/class-networksettings.php
```diff
@@ -0,0 +1,95 @@
...
+// @phpcs:disable Pressbooks.Security.NonceVerification.Missing
+// @phpcs:disable Pressbooks.Security.ValidatedSanitizedInput.MissingUnslash
+
+namespace Pressbooks\Admin\Network;
+
+class NetworkSettings {
+
+	const DEFAULT_THEME_OPTION = 'pressbooks_default_book_theme';
+
+	const DEFAULT_THEME = 'pressbooks-malala';
```
richard015ar, line 16: We will use these constants when adding a hook in pressbooks config plugin to create the default theme option by default when the network is created.

inc/admin/network/class-networksettings.php
```diff
@@ -0,0 +1,95 @@
...
+		add_filter( 'wpmu_options', [ $obj, 'renderCustomOptions' ] );
+		add_action( 'update_wpmu_options', [ $obj, 'saveNetworkSettings' ] );
+	}
+
+	/**
+	 * Render custom network settings options
+	 *
+	 * @return void
+	 */
+	public function renderCustomOptions() : void {
```
richard015ar, line 58: For more options to render and save we would iterate over the `$this->customOptions` array, instead pick the `'default_theme'` option one.

inc/admin/network/class-networksettings.php
SteelWagstaff: Should it be `pressbooks-book` rather than Malala, since the first is open source?

richard015ar: yes, you're right. I created a method to get the default theme (see https://github.com/pressbooks/pressbooks/pull/2608/files#diff-ab3df14e8786c533f34c4228942654997df95ceed2493b79d54dcd883945e8b6R102). It basically returns pressbooks-malala is available, otherwise it returns pressbook-book

arzola (approved):
Working flawlessly LGTM

PR #2609, opened 2022-01-26
Update PHPUnit to 9.5 and improve CI pipeline

SteelWagstaff:
This PR bumps the required version of PHPUnit 9.5 and updates the CI pipeline accordingly. The new text matrix includes WP 5.9 (released today) and Prince 14.2, which is what we're using on our production servers, I believe.

We should decide how to handle the PHPUnit and Yoast/phpunit-polyfills dev dependencies. Because we'll only be supporting WP 5.9 or greater following our next release, we only need one of these -- both are redundant.


SteelWagstaff:
Closing in favor of #2612

PR #2612, opened 2022-01-26
chore: update tests to be compatible with PHP 7.3, 7.4 and WP 5.9

arzola:
This PR fix the test suite behaviour to be fully compatible with PHP 7.3 and PHP 7.4 using the latest PHPUnit version (9) I integrated  #2609 into this


tests/test-image.php
```diff
@@ -88,7 +88,7 @@ public function test_fudge_factor() {
 		$before = (int) ini_get( 'memory_limit' );
 		$format = 'png';
 		$file = __DIR__ . '/data/pb.png';
-		@\Pressbooks\Image\fudge_factor( $format, $file );
+		@\Pressbooks\Image\fudge_factor( $format, $file, 100 ); //adjust fudge to png during tests to avoid false positives
```
arzola, line 91: It seems the garbage collector is doing a great job saving memory so that test was not working because it was enough memory to process the image so I adjusted the resize factor during tests to require more memory and increase it accordingly

tests/test-metadata.php
```diff
@@ -1,8 +1,10 @@
 <?php
 
+use DMS\PHPUnitExtensions\ArraySubset\ArraySubsetAsserts;
+
 class MetadataTest extends \WP_UnitTestCase {
 
-	use utilsTrait;
+	use utilsTrait, ArraySubsetAsserts;
```
arzola, line 7: This assertion was removed in PHPUnit 8 I needed to add an external package to use it `https://packagist.org/packages/dms/phpunit-arraysubset-asserts`

tests/test-modules-export-export.php
```diff
@@ -397,7 +397,7 @@ public function test_sanityChecks( $module, $prerequisite ) {
 			}
 			if ( strpos( $format, '\HTMLBook\HTMLBook' ) !== false ) {
 				// TODO: HTMLBook is too strict we don't pass the validation
-			} elseif ( $runtime->isPHPDBG() && strpos( $format, '\Epub\Epub' ) !== false ) {
```
arzola, line 400: PHPUnit > 8 does not require this check, this code is always running in testing thread

tests/test-modules-export-export.php
```diff
@@ -590,7 +590,7 @@ public function test_getContributorsForSectionXHTML() {
 	 */
 	public function test_HTMLBookConstructor() {
 		$html_book = new Pressbooks\Modules\Export\HTMLBook\HTMLBook( [ 'endnotes' => true ] );
-		$this->assertContains( 'endnotes', $_GET );
+		$this->assertArrayHasKey( 'endnotes', $_GET );
```
arzola, line 593: The assertion to check if an array contains an element changed

tests/test-modules-import.php
```diff
@@ -160,7 +160,11 @@ public function test_wxrInsertAndFindTerm() {
 
 		// Clean attachments after test
 		$date = date( 'Y/m' );
-		array_map( 'unlink', array_filter( (array) glob( "/tmp/wordpress/wp-content/uploads/{$date}/*" ) ) );
```
arzola, line 163: Sometimes pressbooks generate not only files but directories, so we need to check if the deleted files are files because unlink does not remove directories and PHP 7.4 throws a warning

tests/test-privacy.php
```diff
@@ -75,6 +75,12 @@ public function test_addPrivacyPolicyContent() {
 	 */
 	public function test_namespace() {
 		$last_updated_before = get_blog_details()->last_updated;
+		// Doing it right
```
arzola, line 78: Every-time we try to setup the wp privacy we need to be sure is being used in the admin area

tests/utils-trait.php
```diff
@@ -256,4 +256,20 @@ public function _allowIframes( $allowed, $context ) {
 
 		return $allowed;
 	}
+
+	/**
+	 * This behaviour is working as expected on the front end, but during tests we need to set this custom values because they are hidden in the UI.
+	 * It was working before because PHP 7.3 was hiding the warning for access to an inexistent array key.
+	 * In the past if you try to read $options['key'] even $option was not an array no warning was triggered.
+	 */
+	public function _setPdfOptionsForTesting() {
```
arzola, line 265: Setting default values for PDF exports during tests (before PHP 7.4) no warnings were triggered but new versions are more strict, those values does not affect the behaviour just needs to be present for some conditionals

fdalcin (approved):
Amazing work @arzola, LGTM.

Issue #1284, opened 2018-07-19
Include Theme, Theme Options, and Custom Styles in Cloning Routine


greatislander:
### Feature Description

The Pressbooks REST API is a content API, and the cloning tool handles content alone by design. However, there are use cases in which it would make sense to allow theme information (both theme selection and theme options) to be exported and imported so that these settings can be migrated between books or networks.

### Feature Use Case

A book is drafted in one place and published in another. The author wishes to ensure that the published book has the same theme options and theme that were used on the draft book.

### Other Notes

Not all book themes will be installed on all Pressbooks networks, so we will need to gracefully fall back to a default theme/theme options (and advise users of this) if a specific theme is not available.

TODO:
- [x] #2575
- [ ] #2576

PR #2617, opened 2022-01-27
fix: improve notice message

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/1284

This change adds an space for the new notice message when a theme is not available during cloning.

For testing, clone a book with an invalid theme (not available in the target network, or with different theme version).


fdalcin (approved):
LGTM

PR #2619, opened 2022-01-28
fix: clean up RSS cache when saving RSS settings

fdalcin:
Partial fix for pressbooks/pressbooks-network-analytics#180. Accompanies pressbooks/pressbooks-network-analytics#188.

This PR instructs the app to clear the RSS settings cache whenever the user updates the RSS settings.

#### How to test

1. Go to the Network Dashboard page and update the feed title and url.
1. Without clearing the cache go to the dashboard page and validate the new feed is displayed properly.


SteelWagstaff (approved):
:+1:

PR #2622, opened 2022-02-01
fix: custom RSS feed not being displayed

fdalcin:
Fixes pressbooks/pressbooks-network-analytics#180

This PR allows users to see the default RSS feed regardless of their role.


SteelWagstaff (dismissed):
:+1:

SteelWagstaff (approved):
:+1:

Issue #2572, opened 2021-11-18
Make "Import" a first level menu item in book dashboard


pb-amy:
### Feature Description

As an author, I'd want to more quickly locate the 'import' option from my book dashboard so that I can bring content into my book shell from other sources. 

### Feature Use Case

Import is a widely-used feature but it is currently nested under "Tools" in the book admin menu. We should make the import page a top level menu item, like 'Export'. The link to import in the tools menu should be removed and a link to the 'Import' tool should be and placed just under 'export' in the book dashboard menu, like this:

![Screenshot from 2022-01-05 21-04-55](https://user-images.githubusercontent.com/13485451/148331234-83dc63b2-ffdb-46ec-b76e-6b5f7248a4ad.png)

The link should continue to point to the same location, i.e. `BOOKURL/wp-admin/tools.php?page=pb_import`

### Other Notes

In the screenshot above, I used the same dashicon as is currently used for the export menu item, but with `transform: scaleX(-1)` applied to the CSS

SteelWagstaff:
Import menu does appear at top level as expected. However, when selected, several of the menu icons for other menu items appear to be flipped vertically and display 'upside-down'. See screenshot:
![Screenshot from 2022-02-09 08-33-52](https://user-images.githubusercontent.com/13485451/153246133-bdd43d1e-b928-48fb-9ef1-0203d364eec4.png)

SteelWagstaff:
Tested. Works as expected on integrations and textopress.

PR #2623, opened 2022-02-01
feat: make "Import" a first level menu item

fdalcin:
Issue #2572 

This PR moves the previous "Tools > Import" to a first level menu item "Import". 

The functionality should behave the same way as before, however as a side effect of this change the Import URL changed from `BOOK_URL/wp-admin/tools.php?page=pb_import` to `BOOK_URL/wp-admin/admin.php?page=pb_import`.

### How to test

1. Create a book and navigate to its dashboard
1. Make sure you see a new "Import" item just below "Export"
1. Make sure you no longer see an "Import" item under "Tools" menu
1. Import content to make sure the functionality behaves as expected


assets/src/styles/pressbooks.scss
```diff
@@ -8,6 +8,12 @@
   transform: translateZ(0);
 }
 
+.toplevel_page_pb_import {
+  .wp-menu-image {
+    transform: rotate(180deg);
+  }
+}
+
```
fdalcin, line 16: Right now I'm using the same icon from the "Export" menu, so I'm rotating it to avoid confusion. We could also pick a new icon from here https://developer.wordpress.org/resource/dashicons/#editor-ul

arzola (approved):
LGTM

PR #2628, opened 2022-02-02
feat: simplified menu for network managers

arzola:
This PR tries to simplify the network manager menus as described in https://github.com/pressbooks/private/issues/674. 

- [x] Simplify the view for network managers so that they only see two links in their top-level nav bar: Network Admin and My Books. Remove the 'network root site dashboard' (highlighted in screenshot below) from view entirely.
Screenshot from 2021-11-05 09-50-55
- [x] Add a new 'Network Home' menu to the network admin dashboard (with the Home dashicon icon) which includes the following list items:
    - Customize Appearance: `wp-admin/customize.php?return=%2Fwp%2Fwp-admin%2Fnav-menus.php`
    - Pages: `wp-admin/edit.php?post_type=page`
    - Media Library: `wp-admin/upload.php`
    - Site Traffic: `wp-admin/index.php?page=koko-analytics` (if Koko Analytics is activated for the network)
    - Redirection: `wp-admin/tools.php?page=redirection.php` (if Redirection plugin is activated for the root site)
- [x] Remove the 'Tools' menu from the network manager view 
- [x] Remove 'Network options' as a child menu item from 'Settings' (Settings itself already links to the same URL for network managers). 

Resulting view for network managers should look like this (if they have an SSO or LTI plugin activated, otherwise 'Integrations' should not appear):
Screenshot from 2021-11-05 09-57-44

### How to test

1. Login as network admin or super admin
2. You should see only two menu options in the main wordpress bar as in the picture above
3. You should see the new Network Home menu with the desired options
4. Every-time you login as network admin or super admin you should be redirected to the network admin dashboard


matthias-noetzli:
This seems like a fairly restrictive change considering that the Network Root Site is fairly independent from Pressbooks. Hiding away the ability to manage Plugins (which might have their own menus too), Settings, or even just a blog seems very inconvenient.

If there is a big need to simplify the menus, the visible options should be configurable to allow for different use cases (which likely would still be a hastle depending on the number of plugins, custom post types, etc.).

SteelWagstaff:
Thanks @matthias-noetzli. We agree (and ran into other limitations). We're going to close this PR and potentially rethink the approach.

PR #2634, opened 2022-02-09
fix: update rotate rule so it does not apply to every icon

fdalcin:
Issue #2572 

This PR is a complement of #2623 and fixes the CSS rule so not all icons are rotated when Import is selected.

### How to test

1. Go to a book admin page
2. Click on the import menu
3. Make sure icons are not rotated/flipped


SteelWagstaff (approved):
LGTM locally

PR #2638, opened 2022-02-11
feat: improve institution feature

fdalcin:
Issue pressbooks/private#782. Accompanies pressbooks/pressbooks-aldine#316 and pressbooks/pressbooks-book#970

This PR adds more institutions to the `institutions.json` file. It moves the institutions field to a separate meta box and changes the field to display institutions grouped by country/region. It also exposes additional information to the API and changes the cloner so it's able to properly clone institutions.

The field should display to all networks, regardless  of `pressbooks-ecampus-ontario` plugin is active or not.

#### How to test

1. Make sure the new 'Institution(s)' field gets displayed in the Book Info section
2. Make sure you're able to save institutions
3. Make sure values are exposed properly (books with no institutions should not expose the information)
4. Make sure cloning routine works without issues and the institutions are cloned properly


hooks-admin.php
```diff
@@ -176,6 +176,7 @@
 	add_action( 'save_post', [ '\Pressbooks\Book', 'consolidatePost' ], 10, 2 );
 	add_action( 'save_post_metadata', '\Pressbooks\Admin\Metaboxes\upload_cover_image', 10, 2 );
 	add_action( 'wp_insert_post', '\Pressbooks\Admin\Metaboxes\add_required_data', 10, 2 );
+	add_action( 'save_post_metadata', '\Pressbooks\Admin\Metaboxes\save_institutions_metadata', 10, 2 );
```
fdalcin, line 179: This hook is responsible for saving the institutions since they are now a custom input.

SteelWagstaff (dismissed):
This looks good to me. I'd like more thorough review from someone more familiar with PHP and unit tests, however.

richard015ar (approved):
### Testing notes
- I have a book without institutions and the institutions field were not displayed.
- I added institutions, some of those with `: & +` and other characters and those were exposed in the API as expected.
- The cloning routine worked as expected, cloning the institutions when present.

The code is great! thanks @fdalcin

PR #2650, opened 2022-02-15
chore: added php 7.3-7.4 compatibility checker

arzola:
This PR adds an extra check to ensure our codebase doesn't have any deprecated or non compatible function within our supported PHP versions

https://github.com/pressbooks/private/issues/815


SteelWagstaff (approved):
:+1:

Issue #2556, opened 2021-12-14
Book Data Collector does not collect all export formats


fdalcin:
Pressbooks currently stores which formats were exported by [saving](https://github.com/pressbooks/pressbooks/blob/dev/inc/datacollector/class-book.php#L291-L299) `pb_exports_by_format` value into `wp_blogmeta` table. Currently this does not cover all export formats available as none of the Common Cartridge with LTI links are detected.

As a developer I would like to add support for detecting and storing all export formats while running the data collector.

For reference, [here](https://github.com/pressbooks/pressbooks/blob/dev/inc/modules/export/namespace.php#L137) where file types are declared, and [here](https://github.com/pressbooks/pressbooks/blob/dev/inc/modules/export/namespace.php#L181) where we retrieve the name that should be stored into the database.

PR #2683, opened 2022-03-07
feat: appends missing exported formats tracking

arzola:
See #2556 

This PR appends missing exported formats

### How to test

1. Export a book using the Common Cartridge with LTI links options
2. Run the data collector `wp eval-file bin/sync/books.php`
3. Check if `wp_blogmeta` `pb_exports_by_formats` key has the expected meta values 

Example

```
EPUB,Digital PDF,Print-pdf,HTMLBook,XHTML,Pressbooks XML,WordPress XML,OpenDocument,Common Cartridge (Web Links),Common Cartridge with LTI links (1.1),Common Cartridge with LTI links (1.2),Common Cartridge with LTI links (1.3)
```

### DIFF
--- a/inc/datacollector/class-book.php
+++ b/inc/datacollector/class-book.php
@@ -292,7 +292,7 @@ class Book {
 		$exports_by_format = '';
 		$latest_exports = \Pressbooks\Utility\latest_exports();
 		foreach ( $latest_exports as $filetype => $filename ) {
-			$filetype = str_replace( '_', '-', $filetype );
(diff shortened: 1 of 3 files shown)


fdalcin (approved):
LGTM, thanks @arzola!

### Test notes

I opened a book and exported all available export formats
image

I ran the cron to sync book metadata and all export formats were properly saved
image

Then I deleted a couple of export files and re-ran the cron, and the export formats are updated properly as well
image

@arzola @SteelWagstaff one thing that I noticed is that all `Common Cartridge with LTI links` have the "Format" displayed as `Imscc` in the table. Should we rename those to the proper name instead? If so, should we create a new ticket or use this PR instead?

Issue #2686, opened 2022-03-08
Fix automated .pot translation update bot


arzola:
The new GitHub action which automatically updates our `.pot` files when relevant changes have been made is stripping some of the dynamic translations and comments. See https://github.com/pressbooks/pressbooks/commit/e341b3a92838f061bc8e77c1a2c690d98b1e84b5 for an example.

## Notes
the command we run manually is `wp pb make-pot . languages/<plugin>.pot --exclude="vendor,symbionts" --headers='{"Report-Msgid-Bugs-To":"https://github.com/pressbooks/<plugin>/"}'` or `wp pb make-pot . languages/<theme>.pot --exclude="vendor" --headers='{"Report-Msgid-Bugs-To":"https://github.com/pressbooks/<theme>/"}'`

The current action runs `wp pb make-pot . languages/pressbooks.pot --domain=pressbooks --slug=pressbooks --package-name="Pressbooks" --headers="{\"Report-Msgid-Bugs-To\":\"https://github.com/pressbooks/pressbooks/issues\"}"`

The action output includes several PHP Warnings: `strpos() expects parameter 1 to be string, object given in /home/runner/.wp-cli/packages/vendor/wp-cli/i18n-command/src/MakePotCommand.php on line 683`

There is only one use of strpos() in MakePotCommand.php: https://github.com/wp-cli/i18n-command/blob/fdc6e1548fa5146dc32718257c495962430dfeb1/src/MakePotCommand.php#L728-L741. The line reference differs from latest branch, which makes me think we're using the version of wp-cli _before_ this commit: https://github.com/wp-cli/i18n-command/commit/72f57f0c7461f8b84d8abd1829248f03649dd60e

Will try updating `wp package install wp-cli/i18n-command:2.2.8` to see what difference that might make

SteelWagstaff:
Recent changes are working better, but now exclude blade template references. See https://github.com/pressbooks/pressbooks/commit/dfc68dc8f4e32af75ad57b8e46bcc765ed547446

![Screenshot from 2022-03-10 15-03-29](https://user-images.githubusercontent.com/13485451/157769788-31a4e7b3-209e-4c79-bc7a-a30681ee55db.png) and

SteelWagstaff:
@arzola shared this in internal Slack, which is worth keeping an eye on? https://github.com/wp-cli/i18n-command/pull/304. We may want to resolve https://github.com/pressbooks/pressbooks/issues/1734 as well at some point soon

[...]

PR #2691, opened 2022-03-10
Fix translation action

SteelWagstaff:
This PR attempts to fix the update-pot action by removing the specific version constraint for `wp-cli/i18n-command`. It also temporarily changes the action so that it will run on a manual trigger & produce an artifact rather than committing directly, until we can confirm that the change works as expected.

Partial fix for https://github.com/pressbooks/pressbooks/issues/2686

### DIFF
--- a/.github/workflows/update-pot.yml
+++ b/.github/workflows/update-pot.yml
@@ -18,11 +18,11 @@ jobs:
     - name: Setup PHP with tools
       uses: shivammathur/setup-php@v2
       with:
-        php-version: '7.3'
+        php-version: '7.4'
         tools: composer, wp-cli
     - name: Install dependencies
       run: |
-        wp package install wp-cli/i18n-command:2.2.8
+        wp package install wp-cli/i18n-command
         cd /home/runner/.wp-cli/packages/
         composer config repositories.wp-cli '{"type": "composer","url": "https://wp-cli.org/package-index/","canonical": false}'
         cd ./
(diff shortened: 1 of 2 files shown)


.github/workflows/update-pot.yml
```diff
@@ -18,20 +18,25 @@ jobs:
     - name: Setup PHP with tools
       uses: shivammathur/setup-php@v2
       with:
-        php-version: '7.3'
+        php-version: '7.4'
         tools: composer, wp-cli
     - name: Install dependencies
       run: |
-        wp package install wp-cli/i18n-command:2.2.8
+        wp package install wp-cli/i18n-command
```
SteelWagstaff, line 25: i think this might have been source of our problem?

.github/workflows/update-pot.yml
```diff
@@ -18,20 +21,25 @@ jobs:
...
         cd /home/runner/.wp-cli/packages/
         composer config repositories.wp-cli '{"type": "composer","url": "https://wp-cli.org/package-index/","canonical": false}'
         cd ./
         wp package install pressbooks/pb-cli:2.1.0
         composer require jenssegers/blade:1.1.0
     - name: Update POT file
       run: wp pb make-pot . languages/pressbooks.pot --domain=pressbooks --slug=pressbooks --package-name="Pressbooks" --headers="{\"Report-Msgid-Bugs-To\":\"https://github.com/pressbooks/pressbooks/issues\"}"
-    - name: Commit updated POT file
-      uses: stefanzweifel/git-auto-commit-action@v4.13.1
+      # Remove the next four lines and uncomment the last five lines once the process has been confirmed to work as desired.
```
SteelWagstaff, line 36: creates artifact instead of committing directly -- this way we can see effect of action (make sure comments don't get deleted)

arzola (approved):

PR #2724, opened 2022-04-04
Add Cherokee fonts and bold variants for Adlam and Canadian Aboriginal

SteelWagstaff:
Partial fix for https://github.com/pressbooks/pressbooks-book/issues/960. Requires corresponding change in https://github.com/pressbooks/pressbooks-book/pull/1001. Also includes build assets after merging several dependabot PRs.

To test:
1. Checkout this branch and `pbb-960-cherokee` from pressbooks-book
2. Open a book and insert characters from the Cherokee unicode block: https://www.unicode.org/charts/PDF/U13A0.pdf or https://en.wikipedia.org/wiki/Cherokee_syllabary#Unicode
3. Click Appearance -> Theme Options, select 'Cherokee' from the 'Language & Script Support' Option, and click 'Save Changes'
4. Make sure that the Cherokee regular & bold typefaces are successfully downloaded to your `/uploads/assets/fonts/` folder
5. Produce EPUB and PDF exports of the chapter with the Cherokee glyphs. Make sure that they're rendered as the correct characters instead of tofu blocks or ? marks


arzola (approved):
LGTM works as expected

Issue #2704, opened 2022-03-16
Remember me checkbox on sign in page lacks focus indicator


SteelWagstaff:
The focus indicator for the 'Remember Me' checkbox has been removed/disabled on our sign in page. It is present on the default WordPress page. We should restore it.

https://images.zenhubusercontent.com/9209369/4b0ce3c7-78a4-423e-86a3-0008695dff11/focus_indicator.mp4

![Screenshot from 2022-03-16 07-20-39.png](https://images.zenhubusercontent.com/6012398f80b7cb0f8d1385bc/0f3cd6a5-2c7d-4567-903a-d99b54fa6569)

Here's the javascript for the login page: https://github.com/pressbooks/pressbooks/blob/dev/assets/src/scripts/login.js

Here's the CSS for the login page: https://github.com/pressbooks/pressbooks/blob/dev/assets/src/styles/login.scss

https://github.com/pressbooks/pressbooks/blob/3eefe8a9e1cf0974241800acf72c6e6a6da65be7/assets/src/styles/login.scss#L53-L62

PR #2727, opened 2022-04-05
Pb 2704 focus indicator

SteelWagstaff:
Fix for https://github.com/pressbooks/pressbooks/issues/2704.


arzola (changes requested):
@SteelWagstaff I think we are not inheriting the right box-shadow value `box-shadow: 0 0 0 3px var(--primary),` that shadow is only applied when the input is active meaning an action occurred, current value is browser's default 1px and it doesn't looks good as the others inputs

SteelWagstaff:
@arzola you've identified the problem I had in testing this locally. The SCSS compiler appears not to be able to handle things like `box-shadow: 0 0 0 3px rgb(var(--primary), .25)` for example, which is what we'd ideally want to do. I would have hoped that the default rgb color method could have handled this: https://sass-lang.com/documentation/modules#rgb or perhaps we could have done something with sass:color: https://sass-lang.com/documentation/modules/color, but I think maybe that this variable is being provided AFTER the compiler is run, which breaks it? The reason I think that may be the case is that var(--primary) is controlled in the Aldine configuration settings (it can be changed after compile), but the SCSS has to be built/compiled into CSS earlier. I'm not sure, but have asked @fdalcin to take a look into this briefly as well.

fdalcin:
To keep as reference https://medium.com/techhive-io/how-to-use-css-variables-with-sass-mixins-671e1f6067b3

fdalcin (approved):
LGTM

Here's the test:
https://user-images.githubusercontent.com/1761690/162225004-e2351ff9-6cce-404a-8221-0061dee1c53e.mov

Issue #2429, opened 2021-10-11
Remove `dx` from DOI URLs


SteelWagstaff:
An open source user has informed us that CrossRef's guidance for DOI URLs indicate that they should no longer include `dx` as part of the subdomain. See https://www.crossref.org/display-guidelines/ and https://www.doi.org/factsheets/DOIProxy.html. We should remove dx from the URLs we automatically create for books with DOIs by editing the DOI resolver URL everywhere it occurs. See https://github.com/search?q=org%3Apressbooks+dx.doi.org&type=code 

See https://github.com/pressbooks/pressbooks/pull/1429 for initial implementation.

SteelWagstaff:
Tested and working as expected on integrations. Ready for release.

PR #2728, opened 2022-04-05
Replace DOI resolver URL

SteelWagstaff:
Partial fix for https://github.com/pressbooks/pressbooks/issues/2429. Accompanies https://github.com/pressbooks/pressbooks-book/pull/1003

To test:
1. Checkout this branch and `pb-2429-doi-fix` in pressbooks-book 
2. Enter a DOI for a book in book info (for example `10.1109/5.771073`) and save changes
3. Make sure that the resolver URL uses `https://doi.org` instead of `https://dx.doi.org` when rendered on the book homepage and in file exports
Screenshot from 2022-04-05 14-40-06

### DIFF
--- a/inc/metadata/namespace.php
+++ b/inc/metadata/namespace.php
@@ -371,11 +371,11 @@ function book_information_to_schema( array $book_information, bool $network_excl
 			'value' => $book_information['pb_book_doi'],
 		];
 		/**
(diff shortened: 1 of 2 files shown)


fdalcin (approved):
LGTM, thank you @SteelWagstaff!

Issue #2344, opened 2021-08-31
Inconsistent access to users' my catalog page for network managers


SteelWagstaff:
Every Pressbooks user can create their own 'my catalog' page (e.g. https://kpu.pressbooks.pub/catalog/openkpu). When logged in, users will see an 'Admin' link in top right of page:
![Screenshot from 2021-08-31 15-43-44](https://user-images.githubusercontent.com/13485451/131585215-df218761-c4d3-4424-b074-c7353c6dd215.png)

Super admins can visit the admin interface for users which are not themselves, but network managers cannot. Network managers can visit the admin interface for users which are not themselves if they alter the URL to include a book, however. 

Network managers should never see the 'Sorry, you're not allowed to access this page' for the admin pages for any users' catalog:
![Screenshot from 2021-09-01 07-37-45.png](https://images.zenhubusercontent.com/6012398f80b7cb0f8d1385bc/108ddfc2-9a5b-4439-92c4-eab2f23db4ff)

Examples:
Super admins can visit: https://integrations.pressbooks.network/wp/wp-admin/index.php?page=pb_catalog&user_id=1 but network managers can’t (when they’re not user_id=1). Network managers can view https://integrations.pressbooks.network/opentextbookswithstudents/wp-admin/index.php?page=pb_catalog&user_id=1, however. 

Reported by KPU: https://pressbooks.zendesk.com/agent/tickets/6789

SteelWagstaff:
Desired change: only run the check in lines 18-20 if the user !is_super_admin() or !is_restricted() [never show the can't do that message to super admins or network managers]

SteelWagstaff:
Tested and works as expected on integrations. @tw77 if you'd like to do additional testing, you're welcome to, but I'm satisfied.

PR #2730, opened 2022-04-06
fix: access to my catalog page

fdalcin:
Issue #2344 

This PR is an attempt to fix the inconsistencies in the "My Catalog" page access for network managers.


SteelWagstaff (approved):
LGTM. Regular user sees 'admin' link on their own catalog. Clicking it takes them to their own catalog. They do not see an admin link when visiting other users' catalog pages. When they try to visit another users admin page directly (at network root or in a book), they see a 'You do not have permission to do that' message
Screenshot from 2022-04-06 10-13-08

Network managers and super admins now see the 'admin' link for all users catalogues and can visit them at root/book level without a permissions error.

PR #2737, opened 2022-04-10
feat: add built assets to Dependabot PRs

greatislander:
This PR adds a new GitHub Actions workflow to facilitate testing and merging Dependabot pull requests which modify front end (CSS or JavaScript) dependencies.

## Details

Currently, when Dependabot issues a PR to update a front end dependency (CSS or JavaScript), the developer must test the PR's changes using the following steps:

1. Check out the PR branch.
2. Run `npm ci` to install the updated dependency.
3. Run `npm run build` to compile Sass and transpile JavaScript.
4. Test the changes.
5. Once the changes have been tested and the PR merged, the developer must then run the build script and commit the built assets to the `dev` branch.

This PR removes steps 2, 3, and 5 using the following approach. When a Dependabot PR is issued which modifies `package.json` and/or `package-lock.json` files, a workflow is triggered which:

1. Checks out the Dependabot PR branch.
2. Runs `npm ci` to install the updated dependency.
3. Runs `npm run build` to compile Sass and transpile JavaScript.
4. If the built assets have changed, issues a PR _against the Dependabot branch_.
6. Automatically merges the PR with built assets _into the Dependabot branch_.

This means that Dependabot PRs will now include updated CSS and JavaScript build artifacts, which means:

- [x] Developers can test Dependabot changes immediately without needing to install and build assets.
- [x] Merging a Dependabot PR also updates build artifacts.

## Other notes

- Any changes to the Dependabot PR branch (e.g. `@dependabot rebase`) will trigger a rebuild and automerge of the assets.
- This workflow can be added without modification to other repositories with front-end dependencies. It might even be a good candidate for an organization-level workflow as it does not require per-project customization (only that the primary branch is `dev`).

### DIFF
--- a/.github/workflows/autobuild-dependabot-assets.yml
+++ b/.github/workflows/autobuild-dependabot-assets.yml
@@ -17,11 +17,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ github.head_ref }}
-      - name: Cache Node modules
-        uses: actions/cache@v2
-        with:
-          path: node_modules
-          key: ${{ runner.OS }}-build-${{ hashFiles('**/package-lock.json') }}
       - name: Install dependencies
         run: npm ci
       - name: Build assets
--- a/.github/workflows/autobuild-dependabot-assets.yml
+++ b/.github/workflows/autobuild-dependabot-assets.yml
@@ -1,0 +1,46 @@
+name: Add built assets to Dependabot PRs
+
+on:
+  pull_request_target:
+    branches:
+      - dev
+    paths:
+      - package.json
+      - package-lock.json
+
+jobs:
+  autobuild:
+    if: startsWith(github.head_ref, 'autobuild') == false && startsWith(github.head_ref, 'dependabot') == true && github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Dependabot branch
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+      - name: Cache Node modules
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-build-${{ hashFiles('**/package-lock.json') }}
+      - name: Install dependencies
+        run: npm ci
+      - name: Build assets
+        run: npm run build
+      - name: Create pull request with built assets
+        id: cpr
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: "chore(build): build assets"
+          title: "chore(build): build assets"
+          body: This is an auto-generated PR to build assets changed by Dependabot.
+          labels: dependencies, automerge
+          branch: "autobuild/${{ github.head_ref }}"
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Merge pull request with built assets into Dependabot branch
+        if: ${{ steps.cpr.outputs.pull-request-number }}
+        uses: "pascalgn/automerge-action@v0.15.2"
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          MERGE_LABELS: automerge
+          MERGE_METHOD: squash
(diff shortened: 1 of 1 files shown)


SteelWagstaff (commented):
This looks great to me, generally. Thanks, Ned.

.github/workflows/autobuild-dependabot-assets.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Create pull request with built assets
+        id: cpr
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: "chore(build): build assets"
+          title: "chore(build): build assets"
+          body: This is an auto-generated PR to build assets changed by Dependabot.
+          labels: dependencies, automerge
+          branch: "autobuild/${{ github.head_ref }}"
+          token: ${{ secrets.GITHUB_TOKEN }}
```
SteelWagstaff, line 33: @greatislander Will we need to define/add a GITHUB_TOKEN secret as an organizational or repository secret for this to work? If so, we already have an organizationl secret defined that is used for similar things. I believe it's called `PAT_FOR_GITHUB_ACTIONS`. Could/should we change token reference to match that name?

SteelWagstaff (approved):
:+1:

SteelWagstaff:
@greatislander just ran rebase on a js dependabot PR in pressbooks/pressbooks. Files were updated via force push to PR branch as expected. The only thing I'm confused about is when/how the tests are triggered. Are tests  run on the PR _after_ the JS updates are committed or before and how can I know this is the case? Here's what I think is happening:

1. the rebase action opens a PR against the original dependabot PR branch and triggers tests with these changes. These tests either pass/fail and I can see the results in https://github.com/pressbooks/pressbooks/pull/2738/checks?sha=9a0d97f7bc138cf4787b837860c97055168be68e
2. the PR with build assets is automatically approved and committed/force pushed to the dependabot PR branch. Tests are NOT rerun against this PR branch itself following the commit: https://github.com/pressbooks/pressbooks/pull/2738/checks?sha=bf2ed5e5b2424112aece0f78ac3ed38e09134e2e

The most important thing for me is that I can look at a dependabot PR and have confidence that the build assets were updated successfully and that the test suite was triggered AFTEr the build assets were updated.

greatislander:
Tests were not re-run, you're right. That's because:

> When a pull request is merged by this action, the merge will not trigger other GitHub workflows. Similarly, when another GitHub workflow creates a pull request, this action will not be triggered. This is because [an action in a workflow run can't trigger a new workflow run](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows). However, the [workflow_run](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_run) event is triggered as expected.

Source: https://github.com/pascalgn/automerge-action#limitations

I took a quick look at the [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) trigger and I believe the following change to https://github.com/pressbooks/pressbooks/blob/dev/.github/workflows/lint-and-test.yml would allow the merge into the Dependabot branch to trigger the checks as expected:

```diff
name: Lint and test

on:
  push:
    branches: [ dev, master ]
    tags:
      - '*.*.*'
  pull_request:
    branches: [ dev ]
+ workflow_run:
+   workflows: ["Add built assets to Dependabot PRs"]
+   types:
+     - completed
```

The docs for [pascalgn/automerge-action](https://github.com/pascalgn/automerge-action) also mention using a personal access token instead of `GITHUB_TOKEN` to "work around the limitation". However the link to the relevant GitHub documentation is broken so I'm not sure if that still works. If you were to try that I believe it would go here: https://github.com/pressbooks/pressbooks/blob/afa1ac92641e80e71ecd48b2fbcab40b08dcae41/.github/workflows/autobuild-dependabot-assets.yml#L38

### Conclusion

I think one important thing to note here is that _all your tests will pass or fail regardless of whether or not the front-end assets are built_. Unless you add full front-end automation testing or browser snapshots or something, the only GitHub Actions that run that are impacted by NPM dependency updates are the linters, and these run on uncompiled Sass and JS, so you'll already know whether they pass or fail before the assets are built.

PR #2748, opened 2022-04-15
Replace intervention

SteelWagstaff:
This PR moves some of the functionality previously delegated to soberwp/intervention in pressbooks-book and aldine into core Pressbooks. It accompanies https://github.com/pressbooks/pressbooks-book/pull/1010 and https://github.com/pressbooks/pressbooks-aldine/pull/326

It also separates our CI pipeline into two separate steps: one which builds and lints js/css assets and another which runs standards and tests on php files. This will make checks faster, since we won't be repeatedly running the build & lint steps in each of our test matrices.

### DIFF
--- a/.github/workflows/build-and-lint.yml
+++ b/.github/workflows/build-and-lint.yml
@@ -13,15 +13,15 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node }}
-      - name: Cache node modules
+      - name: Cache Node modules
         uses: actions/cache@v2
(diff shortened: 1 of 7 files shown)


greatislander (approved):

greatislander (changes requested):
I'd add some tests to ensure coverage doesn't slip.

SteelWagstaff:
@greatislander Added new tests and split CI pipeline into two separate steps as recommended.

greatislander (approved):

PR #2767, opened 2022-04-28
feat: support pressbooks-vip features

fdalcin:
Issues pressbooks/private#881, pressbooks/private#877, pressbooks/private#878, and pressbooks/private#879

TK


SteelWagstaff (approved):
Looks good to me!

PR #2790, opened 2022-05-17
Only display 3 most recent posts in dashboard RSS feeds

SteelWagstaff:
Only display 3 most recent posts in dashboard RSS feeds (previously displayed 5 posts). Accompanies https://github.com/pressbooks/pressbooks-plugins-config/pull/28

May need to clear cache to see changes locally.


fdalcin (approved):
LGTM

Issue #2646, opened 2022-02-15
Create a new ‘Support Resources’ block on the user dashboard


pb-amy:
### Description

As a user of Pressbooks, it would be ideal to see all of the support resources available to me in one single place, preferably in a visually appealing and organized way on the Pressbooks dashboard upon logging in.

<img width="1397" alt="Screen Shot 2022-02-15 at 2 14 26 PM" src="https://user-images.githubusercontent.com/61845107/154132589-fe09cfc3-6d0d-4a97-af95-b360169215ca.png">

Need to know: where do we want this box to live? What should to box's title to be (copy)? Links, design.

SteelWagstaff:
I'd like to start with a simpler widget that matches our existing dashboard design. How's this for copy:
`Our [Pressbooks User Guide](https://guide.pressbooks.com/) is a one-stop resource which describes how to do nearly everything with Pressbooks. Start with our [5 step guide to making a book](https://guide.pressbooks.com/front-matter/the-simple-5-step-guide-to-making-a-book-with-pressbooks/) or explore the table of content to find instructions for whatever you're trying to do.

Our [YouTube channel](https://www.youtube.com/c/Pressbooks/videos) features several training videos that cover a wide variety of common publishing tasks.

We regularly host [live training webinars](https://pressbooks.com/webinars/) which cover useful topics for beginners and advanced users alike.`

User dashboard will now look like this:
![Screenshot from 2022-05-17 12-39-47](https://user-images.githubusercontent.com/13485451/168896281-b6f63978-a3ba-46ba-a269-2f9e40481539.png)

SteelWagstaff:
See https://docs.google.com/document/d/1nd2jcxRTBi_J87glwOCfC6lyuiKrCkfUJ1nOfSLDRl8/edit#heading=h.om5y2pg57ssp for discussion

SteelWagstaff:
Tested and displays as expected.

PR #2796, opened 2022-05-19
Improve dashboard display

SteelWagstaff:
Minor fixes for create and clone a book buttons on user dashboard. Fix for https://github.com/pressbooks/pressbooks/issues/2646

To test:
Visit user dashboard and check that the create a book and clone a book links appear next to each other and display properly at all screen widths:
Screenshot from 2022-05-19 16-06-27


fdalcin (approved):
LGTM, buttons appear next to each other on all screen widths as expected

PR #2810, opened 2022-05-28
ci: use wp-cli i18n command instead of pb-cli

greatislander:
Partial resolution to https://github.com/pressbooks/private/issues/910. In addition to replacing the `wp pb make-pot` command with `wp i18n make-pot`, this PR also replaces the `update-translations.yml` workflow's use of gettext with the `wp i18n make-mo` command, which does the same thing.

I've commented out the step which autocommits changes. Instead, when these workflows run if they result in any changed files, they will open a PR. Devs can merge if it looks good. In future, you may wish to switch back to autocommit.

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -19,9 +19,7 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: '7.4'
-        tools: composer, wp-cli
-    - name: Install dependencies
-      run: wp package install wp-cli/i18n-command
+        tools: composer, wp-cli/wp-cli-bundle
     - name: Generate MO files
       run: wp i18n make-mo languages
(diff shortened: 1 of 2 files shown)


.github/workflows/update-translations.yml
```diff
@@ -15,12 +15,23 @@ jobs:
       run: tx pull --all --force --minimum-perc=25
       env:
         TX_TOKEN: ${{ secrets.TX_TOKEN }}
-    - name: Install xgettext
-      run: sudo apt-get install -y gettext
+    - name: Setup PHP with tools
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '7.3'
```
SteelWagstaff, line 21:
```suggestion
        php-version: '7.4'
```

SteelWagstaff (approved):
:+1:

PR #2814, opened 2022-06-01
fix: expose and clone private glossary terms

richard015ar:
Issue: https://github.com/pressbooks/private/issues/912

This PR exposes private glossary terms to the API and allows them to be cloned. It also avoid the `no post` string when the glossary term was moved to trash.

### Testing case
- Clone a book with private and public glossary terms. Make sure there are references to public and private glossary terms within multiple chapters.
- The target book must contain all the private and public books with the correct references to the glossary terms.
- Move to trash any glossary term referenced in 1 or more chapters.
- Make sure there's not a `no post` string in the web version of the chapter.


SteelWagstaff (approved):
LGTM :+1:

fdalcin (approved):
LGTM

PR #2849, opened 2022-06-27
fix: handle unset array key in parts API

greatislander:
See: https://github.com/pressbooks/pressbooks/pull/2413#issuecomment-1166846929


greatislander:
Not ready yet. I was working off an older version of dev but the new cloned changes undid the fix.

greatislander:
Ready for review.

greatislander:
No longer needed based on changes in #2897.

PR #2858, opened 2022-06-29
feat: add api items permissions filter and source clonable check filter

richard015ar:
Issue: https://github.com/pressbooks/pressbooks-vip/issues/177

This PR adds 2 new hooks:
-  `pb_set_api_items_permission`: which through a boolean defines if items exposed in each API endpoint can be displayed despite other permissions checking. If it returns false or the filter is not defined, then the normal permission checking flow will occur. 
- `pb_set_source_clonable`: which through a boolean defines if a book is clonable despite the licenses or the `blog_public` metadata. If the filter was not defined, the normal checking flow will occur. 

These changes also adds a new Custom Post Type API extension for back-matter, front-matter, chapter and glossary posts type. It will add a custom permission checking, instead using the default WP API behavior. 

### Testing
For testing, use the these changes in pressbooks-vip: https://github.com/pressbooks/pressbooks-vip/pull/189
And try to read private or licensed books through the API with a regular user or logged our. All the endpoints used by cloning routine must be exposed. Example: `<BOOK URL>/wp-json/h5p/v1/all` for books with H5P activities or `<BOOK_URL>/wp-json/pressbooks/v2/glossary` if the book contains glossary terms.

### DIFF
--- a/inc/cloner/class-cloner.php
+++ b/inc/cloner/class-cloner.php
@@ -996,7 +996,11 @@ class Cloner {
 	 *
 	 * @return bool Whether or not the book is public and licensed for cloning (or true if the current user is a network administrator and the book is in the current network).
 	 */
-	public function isSourceCloneable( $metadata_license ) {
+	public function isSourceCloneable( $metadata_license ): bool {
+		if ( has_filter( 'pb_set_source_clonable' ) ) {
+			return apply_filters( 'pb_set_source_clonable', [] );
+		}
+
 		$restrictive_licenses = [
 			'https://creativecommons.org/licenses/by-nd/4.0/',
 			'https://creativecommons.org/licenses/by-nc-nd/4.0/',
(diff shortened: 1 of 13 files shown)


inc/cloner/class-cloner.php
```diff
@@ -996,7 +996,11 @@ public function getBookGlossary( $url ) {
 	 *
 	 * @return bool Whether or not the book is public and licensed for cloning (or true if the current user is a network administrator and the book is in the current network).
 	 */
-	public function isSourceCloneable( $metadata_license ) {
+	public function isSourceCloneable( $metadata_license ): bool {
+		if ( has_filter( 'pb_set_source_clonable' ) ) {
```
richard015ar, line 1000: I must check if the filter is also true. Fix it.

fdalcin (approved):
LGTM @richard015ar 

Was able to expose private/deactivated books after requesting to clone them as expected. Cloned books from integrations/textopress to make sure we didn't break the cloning routine somehow.

PR #2870, opened 2022-07-06
fix: improve hook handling for clonable hook

richard015ar:
Improving hook handling. See code changes for detail.

Note: It was discussed in a meeting with Dalcin and Oscar.


fdalcin (approved):
LGTM

Issue #2877, opened 2022-07-13
Cloning routine no longer includes front and back matter content


SteelWagstaff:
When cloning books since 5.36.0 was released, the titles are cloned for front and back matter but the content itself is not included. The 'show title' setting is also ignored and set to off upon clone, and show in web/show in exports is turned on, no matter what the setting was in the source book.

We also noticed that the auto-migration cloning routine did not include private front and back matter at all. If a part included only chapters where 'show in web' was unselected, none of the chapters were included in the clone at all. Compare  https://sound.pressbooks.com/ with https://pressbooks.com/sound for one example.

In cases where private content is cloned (when a super admin clones inside a network, for example), the cloned content should have the same title and publishing conditions (show in web, show in exports, show title) as the original content.

PR #2878, opened 2022-07-13
Pb 2877 fix cloning issue

richard015ar:
Issue: #2877 

These changes were discussed and tested locally in a team meeting.


fdalcin (commented):
LGTM!

Tests are failing because the source book is not exposing the content because of the issue we're trying to fix.

fdalcin (approved):

PR #2879, opened 2022-07-14
fix: keep private content private after cloning

fdalcin:
Issue: #2877 

This PR attempts to keep all private content as private after the clone routine completes. This is important because front-matters, chapters, or back-matters that are private in the source book should remain private after cloned.

### Current behaviour

At the moment we clone them as public and add the "Private: {CHAPTER NAME}" as the title.

### New behaviour

Private content will no longer have the  preceding "Private: " before the title, and should remain as private chapters on the target book.

### How to test

1. Create a book with private front-matter, chapters, and back-matters
1. Make the book clonable
1. Clone the book and check the output
1. All book content should be cloned with the right content, and private chapters should not be marked as show in web
1. It's important to check if other parts of the book are cloned as well -- glossary, custom styles, theme, etc.

### DIFF
--- a/inc/api/endpoints/controller/class-customtype.php
+++ b/inc/api/endpoints/controller/class-customtype.php
@@ -65,6 +65,11 @@ class CustomType extends \WP_REST_Posts_Controller {
 				],
 			],
 		];
+		$schema['properties']['status'] = [
+			'description' => __( 'The status for the post.' ),
+			'type'        => 'string',
+			'context' => [ 'view', 'edit', 'embed' ],
+		];
 		$schema['properties']['meta'] = $this->meta->get_field_schema();
 		return $schema;
 	}
--- a/inc/cloner/class-cloner.php
+++ b/inc/cloner/class-cloner.php
@@ -1552,7 +1552,7 @@ class Cloner {
 
 		// Private and public glossaries can be cloned
 		if ( $post_type !== 'glossary' ) {
-			$section['status'] = 'publish';
+			$section['status'] = $section['status'] ?? 'publish';
 		}
 
 		// Download media (images, videos, `select * from wp_posts where post_type = 'attachment'` ... )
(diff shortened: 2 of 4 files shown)


inc/api/endpoints/controller/class-customtype.php
```diff
@@ -65,6 +65,11 @@ public function get_item_schema(): array {
 				],
 			],
 		];
+		$schema['properties']['status'] = [
+			'description' => __( 'The status for the post.' ),
+			'type'        => 'string',
+			'context' => [ 'view', 'edit', 'embed' ],
+		];
```
fdalcin, line 72: Expose the `status` of the content using the API -- this allows us to know if the content is `publish` or `draft` for example.

inc/cloner/class-cloner.php
```diff
@@ -1552,7 +1552,7 @@ protected function cloneSection( $section_id, $post_type, $parent_id = null ) {
 
 		// Private and public glossaries can be cloned
 		if ( $post_type !== 'glossary' ) {
-			$section['status'] = 'publish';
+			$section['status'] = $section['status'] ?? 'publish';
```
fdalcin, line 1555: Use the `status` of the section here and fallback to `publish` if not present.

inc/cloner/class-cloner.php
```diff
@@ -1561,7 +1561,7 @@ protected function cloneSection( $section_id, $post_type, $parent_id = null ) {
 		$content = $this->retrieveH5P( $content );
 
 		// Set title and content
-		$section['title'] = $section['title']['rendered'];
+		$section['title'] = $section['title']['raw'] ?? $section['title']['rendered'];
```
fdalcin, line 1564: Use the `raw` title instead of the `rendered` when available -- this removes the preceding "Private: " on the title of the section.

arzola (approved):
LGTM works as expected

PR #2898, opened 2022-07-29
fix: resolve isset check for invisible parts

greatislander:
Applies changes from #2849 to new PR.


fdalcin (approved):
LGTM, thanks @greatislander

PR #2905, opened 2022-08-08
fix: update _openTextbook method to add blogmeta keys

fdalcin:
Part of pressbooks/pressbooks-vip#249

This PR updates the `_openTextbook` method so it also updates the `blogmeta` table to reflect the changes to the book visibility and license.


richard015ar (approved):
LGTM

Issue #1863, opened 2019-11-27
Chapters that require a password are cloned without content


mcgratay:
### Description

There are three privacy statuses for chapters inside of a public book:
- Shown in Web
- Shown in Exports 
- Password protected

Only chapters which are selected to show in web can be cloned as part of a public, openly licensed book. However, **because password protected chapters require the "Show in web" setting to be enabled, these chapters are also able to be cloned.**

The cloning routine copies the password protected chapters into the new book, but not the content, since that content is hidden. 

Example: https://privacytest.textopress.com

### Steps to Reproduce

1. Create a book with chapter that's shown in web and add filler text to the body
2. Set the chapter to Require a password and save all changes
3. Go to **Book Info** and apply a CC license that allows derivatives
4. Set the book's privacy setting to "Public" on the Organize page
5. Clone that book, then view the password protected chapter

**Expected behavior:** Only chapters that are accessible can be cloned

**Actual behavior:** Password-protected chapters are cloned without any content

SteelWagstaff:
This recently caused an issue for our pressbooks.com -> pressbooks.pub migration tool. We'll need to find a solution so that others aren't adversely impacted.

SteelWagstaff:
To test: check the migration request routine to make sure that password protected content is now cloned.

tw77:
In my initial test I migrated a book from Textopress where all pages were public ("Show in web" checked) but password-protected. The content in each page was cloned successfully including password-protected front matter, chapters in two parts and back matter.

In the resulting book, each page was no longer password-protected, but was now set to private instead ("Show in web" unchecked). I assume this is desired behaviour. The password-protection and specific passwords aren't retained in the migrated book, but the routine ensures that pages that were password-protected are hidden by default. Noting this here just to make sure.

Next up to test:

- migration of a book with pages that are set to private to begin with and that are also password-protected
- in-network clone in Integrations2 (password-protected content shouldn't be cloned here)
- clone from one network to another (password-protected content shouldn't be cloned here)

tw77:
When I clone https://dev.pressbooks.pub/testpwprotected/wp-admin/ to Integrations2 after setting all pages (front matter, chapters in two parts and back matter) to password-protected, I find that, as expected, none of the content of the pages is brought over in this regular clone.

[...]

PR #2906, opened 2022-08-09
fix: expose `raw` key for post titles in API

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/1863

This PR exposes the raw key for post titles during the API responses to avoid the `the_title` hook for the `\PressbooksBook\Filters\add_private_to_title` callback to avoid to prepend `Private: ` to titles when exposing to the TOC API endpoint in the `rendered` key, which will avoid to add it to the target books during the cloning routine. It improves the solution proposed in https://github.com/pressbooks/pressbooks/pull/2903.

### Testing
For testing, you could try to clone books as an admin (to access to private content) with privates front matters, back matters or chapters. 
The target books must respect the same post status than the target book and it must not prepend `Private: ` string to the titles for private posts.

### DIFF
--- a/inc/api/endpoints/controller/class-posts.php
+++ b/inc/api/endpoints/controller/class-posts.php
@@ -102,6 +102,14 @@ class Posts extends \WP_REST_Posts_Controller {
 				'on' : '';
 		}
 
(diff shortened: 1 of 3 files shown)


fdalcin (approved):
LGTM, thanks @richard015ar

PR #2908, opened 2022-08-13
chore(docs): update links to user and developer guides

greatislander:
I noticed that the links to documentation are all broken since docs.pressbooks.org was merged into pressbooks.org. I've updated them all.


SteelWagstaff (approved):
Thanks Ned!

PR #2921, opened 2022-09-01
Replace error_log to trigger_error for Sentry logging

ho-man-chan:
Ticket: https://github.com/pressbooks/private/issues/336

Features:

Add Sentry logging using trigger_error instead of error_log
Test: Errors using the function trigger_error will be sent to Sentry.


richard015ar (approved):
LGTM. Thanks @ho-man-chan

PR #2932, opened 2022-09-13
fix: clean up metadata if key is empty

fdalcin:
This PR attempts to fix an issue while collecting book metadata.

If one adds a metadata for a given field (e.g. Institutions, subjects, editors, publisher) and then later removes that value, the `wp_blogmeta` table would keep old values. This PR aims to clean up old metadata when running the data collector.

__How to test__

1. Add one or more values to institutions, subjects, or editors
2. Check that `wp_blogmeta` has the correct values for those fields -- __pb_institutions__, __pb_subjects__, __pb_subjects_code__, __pb_subjects_string__, __pb_editors__, __pb_publishers__
3. Remove one or more values from any of those fields
4. Check that `wp_blogmeta` has the correct values
5. Empty any of those fields
6. Check that `wp_blogmeta` has the correct values -- empty values should be removed or be set as `null`

### DIFF
--- a/inc/datacollector/class-book.php
+++ b/inc/datacollector/class-book.php
@@ -201,7 +201,6 @@ class Book {
 	 * @param int $book_id
 	 */
 	public function copyBookMetaIntoSiteTable( $book_id ) {
-
 		// TODO:
 		//  Override \Pressbooks\L10n\get_book_language() so that all info collected appears in Admin language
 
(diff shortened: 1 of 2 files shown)


inc/datacollector/class-book.php
```diff
@@ -260,20 +263,25 @@ public function copyBookMetaIntoSiteTable( $book_id ) {
...
+		// clean up subject codes and strings before adding
+		delete_site_meta( $book_id, self::SUBJECTS_CODES );
+		delete_site_meta( $book_id, self::SUBJECTS_STRINGS );
+
 		if ( $subject_list ) {
-			delete_site_meta( $book_id, self::SUBJECTS_CODES );
-			delete_site_meta( $book_id, self::SUBJECTS_STRINGS );
 			$subjects = explode( ', ', $subject_list );
 			foreach ( $subjects as $subject ) {
 				add_site_meta( $book_id, self::SUBJECTS_CODES, $subject );
```
fdalcin, line 287: Shouldn't we add the line 268 and 270 in this foreach as well?

PR #2936, opened 2022-09-19
Pbpiv 1004 network managers sync

richard015ar:
Related issue: https://github.com/pressbooks/private/issues/1004

This change use a more reliable way to get the restricted users. It is a result of a debug session with part of the Tech Team.


fdalcin (approved):
LGTM

PR #2940, opened 2022-09-21
fix: decode generic post type title for xtml and epub

richard015ar:
Issue: https://github.com/pressbooks/private/issues/895

This change decode the post title string for generic post type exported in PDF and EPUB.

### Test case sample
- Add html tags to one or more of front matters or back matters titles in any book.
- Make sure "Show title" are true in the Organize page for these posts
- Export the book to PDF and EPUB
- Titles in TOC and in the post itself must not contain the html tags explicitly but its renderization instead.

### DIFF
--- a/inc/utility/namespace.php
+++ b/inc/utility/namespace.php
@@ -700,7 +700,7 @@ function format_bytes( $bytes, $precision = 2 ) {
  */
 function debug_error_log( $message, $message_type = null ) {
 	if ( defined( 'WP_TESTS_MULTISITE' ) === false && WP_DEBUG ) {
-		\trigger_error( $message, $message_type ); // @codingStandardsIgnoreLine
+		\trigger_error( $message, $message_type ?? E_USER_NOTICE ); // @codingStandardsIgnoreLine
 	}
 }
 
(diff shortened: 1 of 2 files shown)


inc/utility/namespace.php
```diff
@@ -700,7 +700,7 @@ function format_bytes( $bytes, $precision = 2 ) {
  */
 function debug_error_log( $message, $message_type = null ) {
 	if ( defined( 'WP_TESTS_MULTISITE' ) === false && WP_DEBUG ) {
-		\trigger_error( $message, $message_type ); // @codingStandardsIgnoreLine
+		\trigger_error( $message, $message_type ?? E_USER_NOTICE ); // @codingStandardsIgnoreLine
```
richard015ar, line 703: Debug error logs showed `PHP Warning:  Invalid error type specified in /srv/www/pressbooks.test/current/web/app/plugins/pressbooks/inc/utility/namespace.php on line 703` when trying to see export validation issues. Looks like we are not specifying any error type and `trigger_error` only accept integers as a error type, but not null.

fdalcin (approved):
LGTM

Tested locally and works as expected

PR #2953, opened 2022-09-26
feat: add CTA banner setting

richard015ar:
Issue: https://github.com/pressbooks/pressbooks-network-analytics/issues/254

This PR adds a new Network Setting to display or hide the CTA Banner.
It depends on: https://github.com/pressbooks/pressbooks-book/pull/1112 Make sure those changes are in place.

### Testing case
- Go to Network level > Settings > Network Settings > CTA Banner settings
- By default the checkbox must be true.
- Set it to false (uncheck it) and check any book home page (or another book page). The CTA Banner must no be present.
- Set it to true, and the banner in any book home page must be present.

### DIFF
--- a/inc/admin/network/class-networksettings.php
+++ b/inc/admin/network/class-networksettings.php
@@ -15,6 +15,8 @@ class NetworkSettings {
 
 	const DEFAULT_THEME = 'pressbooks-malala';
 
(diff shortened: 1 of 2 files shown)


SteelWagstaff (approved):
Works as expected when testing locally.

Issue #2439, opened 2021-10-12
Two elements do not include correct label in name


SteelWagstaff:
Two labels for UI components do not contain the same text as is presented visually:
- [ ] The accessible name for the GL button within the editor is Insert Glossary Term. This may prevent voice control software users from being able to activate the button. 
![Screenshot from 2021-10-12 16-11-47](https://user-images.githubusercontent.com/13485451/137041144-1294199a-020c-4957-bfc4-23f872517237.png)
We should replace the term text for GL, LaTeX, and Footnote buttons with dashicons. Steel to provide preferred dashicon: https://developer.wordpress.org/resource/dashicons/
- [ ] the accessible name for the Choose File button on the Import page is “Import Source”.
![Screenshot from 2021-10-12 16-12-15](https://user-images.githubusercontent.com/13485451/137041138-1a0aadb2-229e-4213-976c-1886840fdf51.png)

SteelWagstaff:
Tested and works as expected on integrations. Ready for release.

SteelWagstaff:
Found a bug with import routine from HTML source. The source URL input field remains `disabled` even after the import from URL radio button is selected:
![Screenshot from 2022-10-17 09-58-33](https://user-images.githubusercontent.com/13485451/196238360-9117e706-fadd-42f2-b82a-2f6d784bb9b4.png)

I think it's because we removed this javascript 
![Screenshot from 2022-10-17 10-00-45](https://user-images.githubusercontent.com/13485451/196238783-33600407-417f-4b4f-89de-62af54812037.png)

but left `disabled` in place here: https://github.com/pressbooks/pressbooks/blob/3710c37a3e09d0731291efbe304158d4acc5a801/templates/admin/import.php#L206

greatislander:
That `disabled` attribute just needs to be removed. It's hidden rather than disabled so no longer necessary.

fdalcin:
Checking it now

SteelWagstaff:
Fixed and included in Pressbooks 6.2.1 release

PR #2959, opened 2022-09-30
fix: improve labelling of file input on import page (#2439)

greatislander:
This PR improves the import source selection by properly labelling the file upload, partially resolving #2439. Now instead of showing both file input and URL input and disabling the one which isn't selected, the appropriate option will now be shown based on the radio value. I've implemented this change using AlpineJS.


greatislander:
One note, @SteelWagstaff et al— I don't know if progressively adopting Alpine for admin JS is of interest to the team. If not I can switch this back to vanilla JS as it's very minimal.

SteelWagstaff:
@greatislander yeah -- that's the part I'm not sure about myself. Let me talk with the team on Monday?

fdalcin:
@greatislander @SteelWagstaff I’m ok with it, will require some work though.

greatislander:
> @greatislander @SteelWagstaff I’m ok with it, will require some work though.

My thinking is it's the kind of transition that could happen gradually, so whenever an interactive part of the admin UI that uses JavaScript gets updated, it could be transitioned to Alpine if there's bandwidth for that. I took a queue from your use of Alpine in the new Network Catalog here.

richard015ar:
Thanks, @fdalcin for ping the team about this. @greatislander implementing Aldine is a great idea; let's start with it. Thank you!

PR #2963, opened 2022-10-01
fix: improve organize table accessibility and table mobile layouts (resolve #1784, resolve #2438)

### LINKED ISSUE #1784: Admin Organize Tables Lack Accessible Design
There are many a11y barriers in the current Organize page. 

## Specific problems which should be addressed
* Row headers on the Organize page are not identified as headers even though they define the entire row. This means that when screen readers read out the contents of one of the checkbox cells they simply read it as a checkbox and do not associate it with the related section. Row headers should have role=”rowheader”  
Screenshot from 2021-10-12 15-52-01
* The links ("Edit", "Trash", "View", "Move Up", "Move Down") in the row header cells can only be viewed on hover. While these elements still receive keyboard focus, they are not visible when using a keyboard. Note that similar functionality on other pages, such as the Contributors page, functions correctly and the options are revealed when users tab through the page. 
* The buttons in the ```<tfoot>``` of each table add an additional empty row to every table, which is confusing. These cells should not have role=”columnheader”. Ideally, the Add links should exist outside of the table programmatically, while visually they can remain in the table to clarify what section content will be added to. Alternatively, they can be regular table rows with a single cell containing the Add button, which spans all table columns. 
* The content of the first cell of several of the tables (Column 1 header, "Main Body" in the chapters table) does not fit with the other content of the table and is confusing. If something is going to go in the table header it should be a column header. This content would ideally be moved out of the table entirely.
* The 'select/unselect all' table headers functionality is not well communicated to users. They function like buttons, but are communicated as links. We should more clearly indicate the user's ability to show/hide all elements in a column through a separate link/button (select all/deselect all which are always present to the user).
* Labels for checkboxes do not provide enough specificity about the checkbox. The label connects the checkbox with the column header but not the row header. 

**Expected behavior:** 
* Row headers should be clearly labeled as row headers. 
* Links that appear in the row headers should be moved to menu buttons in the row header cell. This way they will not be associated with every cell in the row.
* Data that does not belong to the tabular data of the table should be moved out of the table. This inlcudes the "Add X" buttons and the "Main Body" links in the chapters table. 

## TODO
- [ ] turn chapter names into row headers with roles & any other needed aria attributes
- [ ] insert permanent menu buttons for links edit | trash | view buttons (the ones which now only appear on hover)
- [ ] move 'add front matter' etc buttons out of table  
- [ ] Add permanent select all/deselect all buttons to the show in web / show in exports / show title columns
- [ ] replace part name with static column header called 'Chapter Title'
- [ ] move part name and actions outside of table + place 'add chapter' button next to them [no longer in table foot]
- [ ] Include information about the type of content (for example Dedication, Glossary, Demo Chapter, etc.) in the label. The following checkbox is communicated as “show in web”, instead it should be “Demo Chapter, Show in Web”. 
Screenshot from 2021-10-12 15-34-19

## Organize page
https://github.com/pressbooks/pressbooks/blob/dev/inc/admin/organize/namespace.php

### LINKED ISSUE #2438: Improve display of admin menus for users on narrow mobile devices
When viewport is 320 pixels wide some content and functionality is cut off. 
- [x] Organization options and checkboxes on the Organize page
Screenshot from 2021-10-12 16-02-13
- [x] table header on the Contributors page. 
Screenshot from 2021-10-12 16-02-35
- [ ] some inputs on the Edit Book Information page require horizontal scrolling
Screenshot from 2021-10-12 16-02-44
- [ ] content on the Theme Options page is cut off. 
Screenshot from 2021-10-12 16-03-05
- [x] content on the Contact Us page is cut off
Screenshot from 2021-10-12 16-02-55

greatislander:
This PR addresses the first two items in #2438:

- Organization options and checkboxes on the Organize page
- Table header on the Contributors page

It also converts the Organize page to a blade template and addresses the main issues identified in #1784.


richard015ar (changes requested):
Thanks, @greatislander ! The Contributors header looks much nicer now, and the actions on the Organize page also look good.

Nevertheless, the Organize page's checkboxes still seem to be broken: https://watch.screencastify.com/v/rN45hV2guG4YKBeZFwwy.

I don't know if the original intention was to exclude the fix checkboxes or if it is a style issue added here.
Happy to merge if we reduce the scope of the PR.

greatislander:
@richard015ar I missed that. I'll fix those as well and update the PR.

greatislander:
@richard015ar Ready for another review. After discussing with @SteelWagstaff I also converted the Organize page to a blade template and addressed the main issues identified in https://github.com/pressbooks/pressbooks/issues/1784. I expect you will want to redesign and rebuild this page altogether at some point but having a blade template to work with should expedite that process.

SteelWagstaff:
Thanks @greatislander -- can you run `composer fix` & commit changes to address the standards nags here: https://github.com/pressbooks/pressbooks/actions/runs/3751065665/jobs/6371588244?

SteelWagstaff (changes requested):
Love the improvements @greatislander! Two requests:
1. Please move the `Add Front Matter`, `Add Chapter`, `Add Part` buttons from below the table to just to the right of the content type heading element. Perhaps with a container div with flex applied? Something a bit more like this screenshot?
Screenshot from 2022-12-22 09-55-57
2. Please remove the footer column from the tables

SteelWagstaff:
@greatislander I pulled your changes into a branch, made some blade template and CSS tweaks, and resolved the .pot conflict. There's a new PR from that branch here: https://github.com/pressbooks/pressbooks/pull/3101. You willing to review that, or to make similar commits to this branch?

SteelWagstaff (approved):

Issue #2636, opened 2022-02-08
Fix code to be PHP 8 compatible


arzola:
Four errors in current CI tests matrix when running against PHP8: 

Example:
```
1) SamesiteTest::test_login
PHPUnit\Framework\Exception: PHP Deprecated:  Required parameter $object_type follows optional parameter $group_slug in /home/runner/work/pressbooks-lti-provider-1p3/pressbooks/symbionts/custom-metadata/custom_metadata.php on line 772
PHP Warning:  The magic method Pressbooks\Book::__wakeup() must have public visibility in /home/runner/work/pressbooks-lti-provider-1p3/pressbooks/inc/class-book.php on line 62
```

- [x] Make sure that pressbooks/pressbooks passes test suite for PHP 8 and can run without issues on a server using PHP 8

connerbw:
The fix is probably removing the `= ''` part. Cursory glance I don't see how `is_registered_field` can work with less than all 3 params.

beckej13820:
I didn't know that PHP 8 wasn't supported in Pressbooks.

There is another error with PHP 8 compatibility. The import of documents often (but not always) has an event stream error with the progress slidebar.

connerbw:
I upgraded a legacy projet with Rector recently. (becayuse my wife's blog still runs on one of my old projects):

This was the config I used:

```php
return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->paths([
        __DIR__ . '/sux0r2',
    ]);

    // Directories to skip, rules I don't agree with...
    $rectorConfig->skip([
        __DIR__ . '/sux0r2/includes/symbionts',
        __DIR__ . '/sux0r2/data',
        __DIR__ . '/sux0r2/temporary',
        \Rector\Php80\Rector\FunctionLike\UnionTypesRector::class,
        \Rector\Php81\Rector\FunctionLike\IntersectionTypesRector::class,
        \Rector\RemovingStatic\Rector\ClassMethod\LocallyCalledStaticMethodToNonStaticRector::class,
        \Rector\Transform\Rector\StaticCall\StaticCallToNewRector::class,
    ]);

    // define sets of rules
    $rectorConfig->sets([
        LevelSetList::UP_TO_PHP_81
    ]);
};
```

More info about Rector: https://getrector.org/

PR #2975, opened 2022-10-12
chore: php 8 support

fdalcin:
Issue #2636
Part of pressbooks/private#961

This PR updates the code to support php 8

### DIFF
--- a/.github/workflows/standards-and-tests.yml
+++ b/.github/workflows/standards-and-tests.yml
@@ -15,25 +15,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [8.0]
+        php: [7.4, 8.0]
         os: [ubuntu-20.04]
         epubcheck: [4.2.6]
         prince: [14.3-1]
-        wordpress: [6.0.2]
-        experimental: [false]
+        wordpress: [6.0.2, latest]
         include:
-          - php: 7.4
-            os: ubuntu-20.04
-            wordpress: latest
-            epubcheck: 4.2.6
-            prince: 14.3-1
-            experimental: true
-          - php: 8.0
-            os: ubuntu-20.04
-            wordpress: latest
-            epubcheck: 4.2.6
-            prince: 14.3-1
-            experimental: true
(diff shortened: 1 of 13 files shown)


.github/workflows/standards-and-tests.yml
```diff
@@ -15,25 +15,17 @@ jobs:
...
-            os: ubuntu-20.04
-            wordpress: latest
-            epubcheck: 4.2.6
-            prince: 14.3-1
-            experimental: true
+          - experimental: true
+          - experimental: false
+            php: 8.0
+            wordpress: 6.0.2
+
```
fdalcin, line 28: I've updated the pipeline a bit so now it runs the following matrix

|. PHP | WP | experimental |
| :-: | :-: | :-: |
| 7.4 | 6.0.2 | true |
| 7.4 | latest | true |
| 8.0 | 6.0.2 | false |
| 8.0 | latest | true |

This results on 4 different jobs that will run tests, however only the PHP 8.0 + WP 6.0.2 will run the CS and Code coverage. In the future when we drop PHP 7.4 support we simply remove 7.4 from the test matrix and the pipeline should be ok.

richard015ar (approved):
I tested it locally and works without notice, errors or warnings in the debug and error logs.
Changes LGTM.
However, more manual testing is recommended.

arzola (changes requested):
LGTM

Just we need to keep an eye on the `libxml_disable_entity_loader` removal because according to the docs on versions lower than PHP 8 using the following flag could lead to a security thread

image

https://php.watch/versions/8.0/libxml_disable_entity_loader-deprecation

So I suggest to add one of the following conditions with a TODO tag to remove when we get rid of PHP 7.4

```
if (\PHP_VERSION_ID < 80000) {
      libxml_disable_entity_loader(true);
 }
//or
if (\LIBXML_VERSION < 20900) {
      libxml_disable_entity_loader(true);
 }
```

fdalcin:
Thanks @arzola, I've addressed your suggestion

PR #2977, opened 2022-10-13
fix: unescape title and subtitle

fdalcin:
Issue pressbooks/private#1020

This PR attempts to fix apostrophe not being properly rendered on PDFs.

### How to test

1. Checkout to this branch
2. Update the book title or subtitle to add apostrophes
3. Export a new PDF version of the book
4. Apostrophes should be rendered properly.

>**Note**
This should not break EPUBs but would be good to test them as well.
Also, file names are being generated as the following right now `Jc039s-test-book-1665686454.pdf` instead of `Jcs-test-book-1665686454.pdf`. I don't remember them having the `039` for past releases.

### DIFF
--- a/templates/export/half-title.blade.php
+++ b/templates/export/half-title.blade.php
@@ -1,1 +1,1 @@
-<div id="half-title-page"><h1 class="title">{{ $title }}</h1></div>
+<div id="half-title-page"><h1 class="title">{!! $title !!}</h1></div>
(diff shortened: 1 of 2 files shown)


richard015ar (approved):
LGTM. I tested using different books with special chars in the title, and it works.

SteelWagstaff:
Works as expected on integrations. Ready for release

Issue #2985, opened 2022-10-24
Create health check API endpoints


SteelWagstaff:
Create health check API endpoints to help with infrastructure monitoring. Os proposes making a new plugin managed by composer for this. Could be a small Laravel app.

Do a set of tests: returns either an OK or an error message. 

Check DB connection, redis/cache connection, file system space remaining, availability of main site API

fdalcin:
Created a draft PR so we can discuss about cache approach and other improvements.

fdalcin:
PR adds a new endpoint to the network API - `{NETWORK_URL}/wp-json/pressbooks/v2/health-check?_token={PB_HEALTH_CHECK_TOKEN}`.

**Deploy**
@cmurtagh @ho-man-chan when deploying this to a network, it's necessary to create a new  `PB_HEALTH_CHECK_TOKEN` environment variable with a random string that will be used to authorize the endpoint to be used

**Test**

- When testing it's important to validate the new endpoint fails if the `_token` parameter is missing or is not the same stored in the `.env` file.
- If a valid request is sent, it's important to check if the checks respond as expected (you can disable the cache for a while to see a check failure for example)
- It's important to validate `{NETWORK_URL}/wp-json/wp/v2/users` endpoint returns a 404 response and does not expose user data anymore
- Performing a complete test for clone/import/export routines would also be desired since this touches API and could cause unexpected impact.

fdalcin:
**Health check endpoint**

- [x] Fails with no token parameter
<img width="1789" alt="image" src="https://user-images.githubusercontent.com/1761690/202199013-8a8d7bc4-4cab-43b3-afab-567cfe1e7c15.png">

- [x] Fails with empty or invalid token parameter
<img width="1789" alt="image" src="https://user-images.githubusercontent.com/1761690/202199201-43d6aff8-1331-4514-a79a-c6cb5d10affe.png">
<img width="1789" alt="image" src="https://user-images.githubusercontent.com/1761690/202199308-cd161f9b-7963-401d-a32e-0ba2f70ef26a.png">

- [x] Respond properly when using the correct token
<img width="1789" alt="image" src="https://user-images.githubusercontent.com/1761690/202199511-52cd7620-1240-40cc-91be-ac2c6426aa8f.png">

- [x] Respond properly if something is wrong
<img width="1789" alt="image" src="https://user-images.githubusercontent.com/1761690/202199727-07eb71e9-bb80-4e84-b1fa-d12f7b233432.png">

[...]

PR #3020, opened 2022-11-10
feat: Health check endpoint

fdalcin:
Issue #2985 

This PR aims to add a health check endpoint to help with infrastructure monitoring. It adds a `/wp-json/pressbooks/v2/health-check` endpoint that tests against a couple of checks.

It will return a **200** status code with the following output if every check passes:

image

If one or more checks fail the status code will be **500** and the output would be something like below:

image

This PR also removes all **GET** requests associated with `wp-json/wp/v2/users`. The response should now be a **404** with the following output:

image

**How to test**

1. Add `PB_HEALTH_CHECK_TOKEN` key to your **.env** file with a random string
2. Validate `https://pressbooks.test/wp-json/pressbooks/v2/health-check` API call fails if `_token` param is missing or is different from the one in your **.env** file.
3. Validate `https://pressbooks.test/wp-json/pressbooks/v2/health-check?_token=YOUR-TOKEN-HERE` responds with a **200** status code and all checks **Ok**
4. Make sure a **500** status code is returned if at least one of the checks is not working properly
5. Make sure `wp-json/wp/v2/users` endpoint returns a 404 and cloning and exports work as expected.

### DIFF
--- a/inc/api/endpoints/controller/class-healthcheck.php
+++ b/inc/api/endpoints/controller/class-healthcheck.php
@@ -3,9 +3,9 @@
 namespace Pressbooks\Api\Endpoints\Controller;
 
 use Pressbooks\Health\Check;
-use Pressbooks\Health\checks\CacheCheck;
-use Pressbooks\Health\checks\DatabaseCheck;
-use Pressbooks\Health\checks\FilesystemCheck;
+use Pressbooks\Health\Checks\CacheCheck;
+use Pressbooks\Health\Checks\DatabaseCheck;
+use Pressbooks\Health\Checks\FilesystemCheck;
 use Pressbooks\Health\Checks\ObjectCacheProCheck;
 use Pressbooks\Health\Result;
 use WP_REST_Controller;
--- a/inc/api/endpoints/controller/class-healthcheck.php
+++ b/inc/api/endpoints/controller/class-healthcheck.php
@@ -9,6 +9,7 @@ use Pressbooks\Health\checks\FilesystemCheck;
 use Pressbooks\Health\Checks\ObjectCacheProCheck;
 use Pressbooks\Health\Result;
 use WP_REST_Controller;
+use WP_REST_Request;
 use WP_REST_Response;
 use WP_REST_Server;
 
@@ -22,7 +23,7 @@ class HealthCheck extends WP_REST_Controller {
 		register_rest_route($this->namespace, $this->rest_base, [
 			'methods' => WP_REST_Server::READABLE,
 			'callback' => [ $this, 'healthCheck' ],
-			'permission_callback' => '__return_true',
+			'permission_callback' => [ $this, 'authorize' ],
 		]);
 	}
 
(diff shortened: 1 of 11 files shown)


fdalcin:
@pressbooks/developers I'd appreciate insights about these checks, specially the cache one. How to handle checking cache is in place and working as expected?

cmurtagh:
Let me know if you want to have a quick chat about this, and we can go over it in Zoom.

arzola (approved):
LGTM and works as expected BTW great modern PHP style @fdalcin 😎

inc/api/endpoints/controller/class-healthcheck.php
```diff
@@ -0,0 +1,70 @@
...
+	public function authorize( WP_REST_Request $request ): bool {
+		$value = $request->get_param( '_token' );
+		$expected = env( 'PB_HEALTH_CHECK_TOKEN' );
+
+		if ( $value !== $expected ) {
+			return false;
+		}
+
+		return true;
+	}
```
richard015ar, line 69: I think we could replace it by
```php
return $request->get_param( '_token' ) !== env( 'PB_HEALTH_CHECK_TOKEN' )
```
What do you think @fdalcin ?

inc/api/namespace.php
```diff
@@ -142,6 +142,36 @@ function init_root() {
...
+						continue;
+					}
+
+					if ( ! isset( $handler['methods'] ) ) {
+						continue;
+					}
+
+					if ( $handler['methods'] !== 'GET' ) {
+						continue;
+					}
```
richard015ar, line 166: We could unify this in a single condition I think, kind of `! is_array( $handler ) | ! isset( $handler['methods'] ) | $handler['methods'] !== 'GET'`

inc/health/checks/class-filesystemcheck.php
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	protected function canConnectToFilesystem(): bool {
+		global $wp_filesystem;
+
+		if ( ! WP_Filesystem() ) {
+			return false;
+		}
+
+		return $wp_filesystem->connect();
```
richard015ar, line 42: Cara, does it sounds better? 
```php
global $wp_filesystem;
return WP_Filesystem() ? $wp_filesystem->connect() : false;
```
I think this reduction still sounds clear and simpler to me.

inc/api/endpoints/controller/class-healthcheck.php
fdalcin: I did this way because I was performing extra checks before, I guess now it makes sense to simplify things

inc/api/namespace.php
fdalcin: I don't really like a single if statement because it's harder to read in my opinion. Happy to change if you have a strong preference for it.

inc/health/checks/class-filesystemcheck.php
fdalcin: Yep, will do

inc/api/namespace.php
richard015ar: Got it! both options work for me. I'll resolve this thread.

arzola: I prefer too early return / continue than chaining conditions in this case of scenarios

richard015ar (approved):
LGTM, Thanks cara!

Issue #2359, opened 2021-09-06
Some localized strings don't update until book info is saved a second time


techfiz:
### Description

When a book has more than one 'author' assigned via book info, the authors' names are displayed on the book home page with the word 'and' between the second to last and last authors' names. When a new language is set in book info, this string isn't updated until book info is saved twice. Appears to be a caching issue.

### Steps to Reproduce

1. Add two author or more authors to a book via book info. 
2. View the book home page and observe that the word 'and' has been inserted between the authors' names
3. Change the book language in book info to a language other than English (like French, or Spanish)
4. View the book home page and observe that the word 'and' has not been localized as expected (_et_ or _y_)
5. Save book info a second time
6. Observe that the word has been localized
7. Change book language back to English
8. Observe that the string has not been changed to English
9. Save book info a second time
10. Observer that the word has changed back to English

**Expected behavior:** 
The word 'and' between authors' names should be localized like other metadata on the book home page upon save of book info

**Actual behavior:** 
The word 'and' remains in English until book info has been saved a second time.

techfiz:
@SteelWagstaff Add two authors to get and between the author names on title page. This "and" between names appear in english and there is no option to change that to respective script like chapter and part in admin.

SteelWagstaff:
We generate the book author text here: https://github.com/pressbooks/pressbooks-book/blob/8013a8c5672f4a5673fba26d40b55cc797d00ba9/partials/content-cover-book-header.php#L17

SteelWagstaff:
@techfiz can you help us investigate this by telling us what language you're trying to use for your target book, providing the URL for a sample book where you're experiencing the error? We'd also love to see the contents of the 'diagnostics' page from your network, as well as a list of any 'site languages' that you've installed on your server. This will help us reproduce your issue and fix it more quickly.

SteelWagstaff:
Notes: The language does appear to be updated to use the localized string, but with a delay or lag. If a book is in English and I change the language to French, nothing happens to the author string. If I then change to Spanish, the 'and' in the author string appears in French. If I then change to Arabic, the 'and' appears in Spanish, and so forth. Not sure why this is happening, but it is a bug and should be fixed in a future release.

richard015ar:
Seems to be a cache problem. If we save again in the Book Info section the cache is refreshed and the `and` is translated properly.
[...]

PR #3043, opened 2022-11-28
fix: specify domain translation for contributors imploding method

richard015ar:
This PR intends to fix: https://github.com/pressbooks/pressbooks/issues/2359. Please read it for context.

We prioritize a logged in user language selection (locale) over the book's language when displaying a book from its home page. This change applies this prioritization only when the user is in the admin interface.

### Testing
- Log in using an admin or regular user.
- Go to Edit profile > Language and pick `English (United States)` for example, or any other, except the `Site Default` option.
- Pick a book and go to the Edit Info section
- Add two or more authors to the book, pick a language different from the language you picked up in your profile. For example, following the test case, you could choose `Spanish`. Make sure you add a license.
- Save the Book Info.
- Make the book public in Organize > Public option.
- Go to the book home page. You should see the list of authors with a `and` between the last two authors. It must be in the language of the book. License and other metadata must be in the Book Language, also.
- Log out and revisit the book home page.
- All the metadata, including `and` and license description, must be in the selected book language.


richard015ar:
Running translation github action on this branch, required for this PR.

inc/l10n/namespace.php
```diff
@@ -326,7 +326,7 @@ function wplang_codes() {
  */
 function get_locale() {
 	// If the user has set a locale, use it
-	if ( function_exists( 'wp_get_current_user' ) ) {
+	if ( function_exists( 'wp_get_current_user' ) && is_admin() ) {
```
SteelWagstaff, line 329: @greatislander You willing to help review this PR? @richard015ar discovered that this is probably the change we need, rather than adding the strings to pressbook-book, etc. My understanding is that this will only load the user's locale on WordPress admin pages -- otherwise it will use the global local (set at the level of the site). Helps us make cleaner distinction between front end and back end localization and avoids strings being translated unneccessarily in the front end when a user who has set their language to something other than a book's language views home page while logged in.

Issue #3017, opened 2022-11-09
Build better 'new user dashboard'


SteelWagstaff:
We want to replace the new user dashboard that a user sees the first time they log into Pressbooks (i.e. https://NETWORKURL/wp/wp-admin/) with something that's more inviting and makes their options clearer to them.

See https://www.figma.com/file/DDZ7YuklYsX9AAyVc5caUv/User-Dashboard?node-id=0%3A1 (edited dashboard 4.0) for wireframe prototype. 

We should also ensure that a user is taken to this dashboard after logging in to the root site using the sign in link (rather than simply being taken back to the root site as a logged in user).

richard015ar:
Questions:
- Are there some Mobile/table/others compatibility plans?
- Are there some plans to change/adapt the right-side and top menu? Curious about the discrepancy between these sections and the central part of the dashboard.
- Finally, could network managers customize/add new sections in the future? I don't know if that's something our clients would want.

SteelWagstaff:
Answers:
1. The two columns should collapse into a single column on mobile devices
2. No plans to change top and left-side menus right now
3. No, network managers cannot customize or add new sections at present.
Definition of done is to have something releasable.

fdalcin:
Opened a draft PR with the work in progress, see #3048

tw77:
The new user dashboard works well. But when I login to Integrations using the sign in link on the root site, I'm taken back to the root site as a logged in user rather than being brought to this dashboard.

fdalcin:
Thanks for noticing that @tw77.

@pressbooks/tech we could discuss tomorrow how to solve this behaviour when signing in from the root site

SteelWagstaff:
I think we need to remove get_permaline() from this line and it's fixed

fdalcin:
> I think we need to remove get_permalink() from this line and it's fixed: https://github.com/pressbooks/pressbooks-aldine/blob/3f1e0e0494fc5465227a346131dbd0202aeb5a86/inc/helpers/namespace.php#L357

Wouldn’t this affect the signing in from book url as well?

SteelWagstaff:
It shouldn't. That's controlled in pressbooks-book, I believe: https://github.com/pressbooks/pressbooks-book/blob/6ef05505e52db1c133efac1828dc4334fc1ec811/inc/helpers/namespace.php#L234

tw77:
This is now working as hoped for on Integrations. I'm brought to the dashboard when I sign in from the root site.

[...]

PR #3061, opened 2022-12-08
fix: remove old dashboard references

fdalcin:
#3017 

This PR removes references to the old way of rendering the dashboard + rss feed options


SteelWagstaff (approved):
LGTM

PR #3066, opened 2022-12-09
Improve DOM and styles for clone a book page

SteelWagstaff:
This PR improves the DOM and styles for the clone a book page. In particular, if fixes:
- [x] improve styles for book cards
- [x] don't display directory search section when env variables aren't present
- [x] make sure all input have accessible labels
- [x] improve text when displaying search results (template should say 20 of ### results shown)
- [x] make sure progress bar text is hidden if clone fails

It also make a few small tweaks to the book/user dashboards.


templates/admin/cloner/page.blade.php
```diff
@@ -1,37 +1,35 @@
...
+			</div>
+			<div class="image">
+				<a class="pb-directory-logo" href="https://pressbooks.directory" target="_blank"><img src="https://pressbooks.directory/assets/logo-pressbooks-directory.svg" alt="Pressbooks Directory"/></a>
+			</div>
 		</div>
-		<label class="pb-label" for="searchbox">{{ __( 'Search', 'pressbooks' ) }}</label> <span id="stats" aria-live="polite"></span>
+		<label class="pb-label screen-reader-text" for="searchbox">{{ __( 'Search', 'pressbooks' ) }}</label> <span id="stats" aria-live="polite"></span>
 		<div id="searchbox" class="pb-directory-search"></div>
 		<div id="book-cards" class="book-cards"></div>
 	</div>
```
fdalcin, line 35: I think we should hide this div entirely if the network does not have Algolia API set up

inc/utility/namespace.php
```diff
@@ -1606,7 +1606,5 @@ function delete_options_cached() : void {
  * @return bool
  */
 function are_algolia_search_env_var_presents(): bool {
```
fdalcin, line 1608: Could we rename this method to something like `is_algolia_search_enabled`?

fdalcin (approved):
LGTM

PR #3090, opened 2022-12-15
Bump DocRaptor pipeline version

SteelWagstaff:
Use the latest DocRaptor pipeline in our PDF and cover generator export routines.


inc/covergenerator/class-generator.php
```diff
@@ -459,6 +459,7 @@ public function generateWithDocraptor( $pdf_profile, $document_content, $output_
 						$done = true;
 						break;
 					case 'failed':
+						// TODO: are these done in the wrong order? $done will never be set to true here, I don't think
```
SteelWagstaff, line 462: @fdalcin Can you look at this? Why are are attempting to set $done _after_ running wp_die? Won't that mean it never is changed?

Issue #3092, opened 2022-12-15
Create & Adapt a Book blocks displayed on user dashboard when network settings don't allow users to create & clone books


tw77:
### Description

It's possible for networks to not allow registered users to create or clone books:
<img width="501" alt="Screenshot 2022-12-15 at 1 50 33 PM" src="https://user-images.githubusercontent.com/44182629/207974689-345376de-4523-4f0b-bc3d-76988ece98e4.png">

When this setting is unchecked, the new user dashboard still shows the Create a Book and Adapt a Book options.

For open source users, this setting will be called `Allow new registrations`:
![Screenshot from 2022-12-15 14-43-41](https://user-images.githubusercontent.com/13485451/207982824-2577c543-8c86-43a1-8f0c-f70c5a281eb4.png)

In this case, blocks should not be shown if either of the first two settings are selected:

### Steps to Reproduce

1. In Network Admin > Network Settings > Book & User Registration, uncheck "Allow registered users to create and clone new books"
2. Save changes.
3. Log out.
4. Sign in to the network as a regular user.

**Expected behavior:**  When the global network setting 'Allow registered users to create and clone new books' is unchecked, regular users do not see the `Create a Book` and `Adapt a Book` blocks on their user dashboard. These blocks can still appear for network managers/super admins. 

**Actual behavior:** We're brought to the user dashboard and the Create a Book and Adapt a Book blocks appear, even though those options aren't present under 'My Books' in the top menu. 
<img width="1418" alt="Screenshot 2022-12-15 at 1 54 48 PM" src="https://user-images.githubusercontent.com/44182629/207975337-6b19939d-9c18-4cf9-99c8-2b3a9bb8f633.png">

When a user clicks the `Create a Book` button, they are brought to the registration page where they see the message: 'You are logged in already. No need to register again!' Clicking the `Adapt a Book` button brings them to the cloning page where they are able to clone books normally.

tw77:
The Create a Book button brings them to the registration page where they see the message: 'You are logged in already. No need to register again!' The Adapt a Book button brings them to the cloning page.

tw77:
Tested fix on Integrations and it is working.

PR #3094, opened 2022-12-19
fix: remove create & clone blocks when network setting is off

fdalcin:
Issue #3092 

This PR aims to remove Create & Adapt a Book blocks when the network setting is off.

**How to test**

1. Uncheck "Allow registered users to create and clone new books" in Network Admin > Network Settings > Book & User Registration
2. Save changes.
3. Visit the user dashboard `wp/wp-admin/index.php?page=pb_home_page` and make sure Create & Adapt a Book blocks are not displayed.

>**Note**
If you're logged it as a super admin you will still be able to clone books, so you should be able to see the Clone a Book block.

### DIFF
--- a/inc/admin/dashboard/class-newuserdashboard.php
+++ b/inc/admin/dashboard/class-newuserdashboard.php
@@ -4,7 +4,9 @@
  */
 namespace Pressbooks\Admin\Dashboard;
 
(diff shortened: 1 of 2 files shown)


arzola (approved):
LGTM working as expected

PR #3097, opened 2022-12-21
feat: replace footnote icon with a more suitable one (fix #3019)

### LINKED ISSUE #3019: Replace existing footnote icon with better SVG
In the Pressbooks visual editor, we're now using a paperclip as the footnote icon. Ideally, we'd replace this with an SVG that better calls to mind a footnote. See https://twitter.com/kylemackie/status/1587820976393396224 for context.

greatislander:
Screenshot 2022-12-21 at 6 30 17 PM
I've used the Footnote icon from https://www.npmjs.com/package/@tinymce/oxide-icons-default (the icon set for newer versions of TinyMCE). It's released under an MIT license and that should be indicated but there doesn't seem to be a consistent place in Pressbooks to attribute other dependencies that AREN'T referenced in composer.json, package.json (or bundled in the `symbionts` directory).


SteelWagstaff (approved):

inc/editor/namespace.php
```diff
@@ -108,6 +108,7 @@ function admin_enqueue_scripts( $hook ) {
 		'editor', 'PB_FootnotesToken', [
 			'nonce' => wp_create_nonce( 'pb-footnote-convert' ),
 			'fn_title' => __( 'Insert Footnote', 'pressbooks' ),
+			'fn_icon' => PB_PLUGIN_URL . "assets/dist/images/footnote.svg",
```
SteelWagstaff, line 111:
```suggestion
			'fn_icon' => PB_PLUGIN_URL . 'assets/dist/images/footnote.svg',
```

PR #3100, opened 2022-12-22
fix: remote request POST parameters for completed cloning

richard015ar:
This change fixes the external cloning confirmation and sent the body POST parameters for completed clones to `pressbooks/v2/clone/complete` endpoint.


SteelWagstaff (approved):
:+1:

PR #3110, opened 2023-01-04
feat: upgrade scssphp

arzola:
This PR would update SASS php compiler later we would require a task to swap `@scssphp-import-once` in favor of `@use` in pressbooks-book in order to remove almost all the warnings during tests for now it can be merged to fix our current issue with the new buckram integration

The issue with the current update is that PHP 8 is throwing a TypeError fatal error that was not properly handled by the `class-sass.php` file instead of that we are silencing  the errors with `@` but that is not enough in PHP 8 environments

This would solve the issue for now but probably it could be better later to revisit this function and find a way to properly handle this escenarios instead of silencing the warnings.

We have a couple of bumps regarding coding standards that we could improve in a next iteration I marked those as a TODO

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
 		"monolog/monolog": "^2.2",
 		"pressbooks/mix": "^2.1",
 		"pressbooks/pb-cli": "^3",
-		"scssphp/scssphp": "~1.11.0",
+		"scssphp/scssphp": "^1.11.0",
 		"sinergi/browser-detector": "^6.1",
 		"vanilla/htmlawed": "^2.2",
 		"phpcompatibility/php-compatibility": "^9.3",
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
 		"monolog/monolog": "^2.2",
 		"pressbooks/mix": "^2.1",
 		"pressbooks/pb-cli": "^3",
-		"scssphp/scssphp": "~1.1.0",
+		"scssphp/scssphp": "~1.11.0",
 		"sinergi/browser-detector": "^6.1",
 		"vanilla/htmlawed": "^2.2",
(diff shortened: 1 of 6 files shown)


tests/test-sass.php
```diff
@@ -100,7 +100,8 @@ public function test_compile() {
 		$css = $this->sass->compile( $scss );
 		$expected = <<<EOF
 p {
```
arzola, line 102: New sass compiler prints an extra return

inc/class-sass.php
```diff
@@ -261,7 +261,11 @@ public function parseVariables( $scss ) {
 						$val = "{$fncall}({$fncall_params})";
 						break;
 					default:
-						$val = @( new \ScssPhp\ScssPhp\Compiler() )->compileValue( $item[2] ); // @codingStandardsIgnoreLine
+						try {
+							$val = @( new \ScssPhp\ScssPhp\Compiler() )->compileValue( $item[2] ); // @codingStandardsIgnoreLine
```
arzola, line 265: This is the actual fix for broken integrations, definitely a refactor is needed here to avoid silencing errors with some nodes

composer.json
```diff
@@ -50,7 +50,7 @@
 		"monolog/monolog": "^2.2",
 		"pressbooks/mix": "^2.1",
 		"pressbooks/pb-cli": "^3",
-		"scssphp/scssphp": "~1.1.0",
+		"scssphp/scssphp": "~1.11.0",
```
SteelWagstaff:
```suggestion
		"scssphp/scssphp": "^1.11.0",
```

SteelWagstaff (approved):
Resolves the fatal error for me locally. Will merge and open new issue to fix issue more reliably in upcoming sprint.

Issue #3044, opened 2022-11-29
Make compatible with PHP 8.1


SteelWagstaff:
Fix test matrix for PHP 8.1. We have a number of compatibility warnings in this repo: https://github.com/pressbooks/pressbooks/actions/runs/3568403390/jobs/5997235215

There are three types of deprecation notices:
1. We have 12 'automatic conversion of false to array' deprecation notices in In [inc/admin/covergenerator/namespace/php](https://github.com/pressbooks/pressbooks/blob/56be4dffcb09a83dc27cbcfece7d0969d3cf9228/inc/admin/covergenerator/namespace.php#L341-L581) (lines 341, 353, 365, 377, 389, 401, 417, 430, 512, 549, 560, & 581)

2. We have hundreds of deprecation notices in SCSSPHP: 
`substr(): Passing null to parameter #1 ($string) of type string is deprecated in /home/runner/work/pressbooks/pressbooks/vendor/scssphp/scssphp/src/Compiler.php on line 331`
`preg_match(): Passing null to parameter #4 ($flags) of type int is deprecated in /home/runner/work/pressbooks/pressbooks/vendor/scssphp/scssphp/src/Parser.php on line 1[23](https://github.com/pressbooks/pressbooks/actions/runs/3568403390/jobs/5997235215#step:9:24)`

We can probably resolve with https://github.com/pressbooks/buckram/issues/289 (or https://github.com/pressbooks/pressbooks/issues/2281)

3. There are 10 deprecation notices in rmccue/requests, a dependency of WP-CLI. For example:
`Return type of Requests_Cookie_Jar::offsetExists($key) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/runner/work/pressbooks/pressbooks/vendor/rmccue/requests/library/Requests/Cookie/Jar.php on line 63`

richard015ar:
Number 2 is about the scssphp version. We are using an older version (1.1.0), the latest 1.11.0, that fixes some PHP 8.1 notices.
I fixed it by changing the dependency requirement to: `"scssphp/scssphp": "^1.11.0"`

[...]

PR #3115, opened 2023-01-06
chore: make pressbooks compatible with php8.1

SteelWagstaff:
Issues:
- https://github.com/pressbooks/private/issues/1070
- https://github.com/pressbooks/pressbooks/issues/3044

This PR applies changes to make the plugin compatible with PHP 8.1 and avoid deprecation warnings related to this source code.

A new error handler was added to silence deprecation notices from 3rd parties. Those deprecation notices can be displayed in `development` environment, see: https://github.com/pressbooks/setup-development-environment/pull/10 for more details about it.

For more information in: https://docs.google.com/document/d/1Th56ipk19FNqdeXT32mdRgpAHoIV3wUQOtSrrEr8tjs/edit#

For testing, you will need: https://github.com/pressbooks/setup-development-environment/pull/10 in your Lando setup.


bin/install-wp-tests.sh
```diff
@@ -69,7 +69,7 @@ install_wp() {
 
 install_pressbooks_book() {
 	if [ -d $WP_CORE_DIR/wp-content/themes/pressbooks-book ]; then
-		cd $WP_CORE_DIR/wp-content/themes/pressbooks-book && git pull
+		cd $WP_CORE_DIR/wp-content/themes/pressbooks-book && git pull origin dev
```
richard015ar: After:
We now get:
```bash
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

If you wish to set tracking information for this branch you can do so with:

    git branch --set-upstream-to=<remote>/<branch> fix/scssphp-deprecated-directives
```
when installing tests.
It will fix the issue.

ho-man-chan (commented):
Tested the changes in pressbooks https://github.com/pressbooks/pressbooks/pull/3115 but was unable to reproduce the error silencing feature with the new environment variables.
Todo: will meet with [@richard015ar](https://github.com/richard015ar) to double check

inc/utility/namespace.php
```diff
@@ -964,7 +964,7 @@ function rcopy( $src, $dest, $excludes = [], $includes = [] ) {
  */
 function str_starts_with( $haystack, $needle ) {
 	$length = strlen( $needle );
-	return ( substr( $haystack, 0, $length ) === $needle );
+	return ( substr( $haystack ?? '', 0, $length ) === $needle );
 }
```
fdalcin, line 968: If we require PHP 8+ we could deprecate this method, and remove it soon

fdalcin (approved):
LGTM, I've been able to use pressbooks with php 8.1 without issues and it's possible to show deprecation warnings from third party packages with the new environment variable.

I noticed these deprecation warnings during while exporting, should we also silence them?

image

inc/utility/namespace.php
richard015ar: agree, it is widely used across multiple plugins, so it's going to be a challenge but yes, we could do it soon.

richard015ar:
@fdalcin, upgrading `jenssegers/imagehash`: https://github.com/jenssegers/imagehash solves this issue, looks like we had an old version of this package.
Do you know other places where we could use it? Would be nice to do some manual testing to ensure this upgrade does not break anything else.

This upgrade will cause some issues with PHP 8.0, see: https://github.com/pressbooks/pressbooks/actions/runs/3949268652/jobs/6760260661
Team discussion is suggested.

.github/workflows/standards-and-tests.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     - name: Install PHP dependencies
       run: |
         export PATH="$HOME/.composer/vendor/bin:$PATH"
-        composer install --no-interaction
+        composer update --no-interaction
```
richard015ar: To be discussed with @arzola

fdalcin:
> @fdalcin, upgrading `jenssegers/imagehash`: https://github.com/jenssegers/imagehash solves this issue, looks like we had an old version of this package. Do you know other places where we could use it? Would be nice to do some manual testing to ensure this upgrade does not break anything else.
> 
> This upgrade will cause some issues with PHP 8.0, see: https://github.com/pressbooks/pressbooks/actions/runs/3949268652/jobs/6760260661 Team discussion is suggested.

Export issue is in fact solved by upgrading `jenssegers/imagehash`. However I think we upgraded a symfony package that requires php 8.1, that will cause our pipelines or anything running on php 8 to fail.

fdalcin (approved):
LGTM now, working as expected locally with php 8.0 AND php 8.1. No deprecation warnings, exports/clone routines are working normally. I am also able to hide/display third party deprecation warnings using an environment variable.

PR #3121, opened 2023-01-10
fix: exclude ampersand from decoding action

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1038

This PR adds a new parameter to the widely used `decode` method to include or exclude the `&` character, specifically to be used in the export routine (PDF & EPUB) to include `&` as a part of the decoding action.
In addition, the authors' section in the chapter is not being escaped anymore since some authors may contain `&` or other special chars.

### Use case sample for testing
- For a book, create a contributor which contains `&`. Example: `Music and Visual departments`.
- In a chapter, add one or more sections with & and other special chars
- Add the author created as an author of the chapter.
- Add a subtitle for the chapter; ensure you include `&` in the text.
- In Appearance > Theme Options > Global Options, ensure the `Enable two-level table of contents (displays headings under chapter titles)` and `Display information about authors at the end of each chapter` options are checked.
- Export the book to PDF and EPUB. Both exported files should look as expected.

### DIFF
--- a/inc/modules/export/class-exporthelpers.php
+++ b/inc/modules/export/class-exporthelpers.php
@@ -163,7 +163,7 @@ trait ExportHelpers {
 				foreach ( $sections as $id => $subsection ) {
 					$subsections[] = [
 						'slug' => $is_slug ? "#{$id}" : "${data['href']}#{$id}",
-						'title' => Sanitize\decode( $subsection ),
+						'title' => Sanitize\decode( $subsection, false ),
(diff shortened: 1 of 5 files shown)


richard015ar:
@tw77 this PR could be the first change you could test locally if you want to. I would recommend an additional reviewer.

arzola (approved):
LGTM

Maybe not related but probably we should fix also here 

image

Issue #3123, opened 2023-01-10
Invalid post type” error message when creating a new chapter


jocelynpb2:
### Description

This bug was reported by a network manager from University of Queensland (See ticket https://pressbooks.zendesk.com/agent/tickets/13928) and has been replicated on Integrations (in both existing and new books).
When selecting “Add Chapter” from the Organize page, “Invalid Post Type” error is shown.
This does not occur when adding new front or back matter. 

When “Add chapter” is selected from the Organize menu, this results in the “Add new chapter” page loading correctly. 

### Steps to Reproduce

1. Create a new book 
2. Navigate to the ‘Organize’ tab
3. In the Main Body, click “Ads chapter” beside the “Add part” button

**Expected behavior:** "Add new chapter" page loads
**Actual behavior:** Invalid post type error message

fdalcin:
I believe the reason for this is that the button for some reason has the wrong URL. Checking the URL in the bottom left corner of the image shows `?post_type=part_19` whereas it should be something like `?post_type=chapter&startparent=19`

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/1761690/211653551-826b6cc2-6640-4459-a783-fa72d94b5b2e.png">

Probably the issue might be around this [code block](https://github.com/pressbooks/pressbooks/blob/56b59d228eeddc2d856ae7598b6ac583d96ad34c/templates/admin/organize.blade.php#L78-L85). I can work on a PR tomorrow after we talk more about the issue during bug scrub.

SteelWagstaff:
Working as expected on integrations. Ready for production release.

PR #3126, opened 2023-01-11
fix: add chapter button

fdalcin:
Fix for #3123 

This PR fixes the Add Chapter button inside the organize page. The issue was that we were setting the wrong link when adding a part (see https://github.com/pressbooks/pressbooks/issues/3123#issuecomment-1377827290).

**How to test**

1. Navigate to a book dashboard and visit the organize page
1. Click the "Add Chapter" button right after the part name
1. You should be redirected to the "Add New Chapter" page with the correct Part selected as the parent of the chapter

image
image

### DIFF
--- a/templates/admin/organize.blade.php
+++ b/templates/admin/organize.blade.php
@@ -76,11 +76,17 @@
             </h2>
             @if ($can_edit_posts)
                 <div class="page-title-actions">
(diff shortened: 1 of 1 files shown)


SteelWagstaff (approved):
Works as expected locally.

Issue #3136, opened 2023-01-20
Validation errors shown when exporting an EPUB:


SteelWagstaff:
We observed the following validation errors when exporting an EPUB on our integrations network. They seem to indicate an issue with how the theme CSS files are declared or referenced in the EPUB's package document.

Array
(
    [time] => Wed Jan 18 17:05:31 2023
    [user] => thomas
    [site_url] => https://integrations.pressbooks.network/testshapeshifter2023
    [blog_id] => 706
    [theme] => McLuhan
)
ERROR(OPF-014): /srv/www/integrations.pressbooks.network/releases/20230118044402/web/app/uploads/sites/706/pressbooks/exports/Test-Shape-Shifter-1674061522.epub/EPUB/mcluhan.css(0,0): The property "remote-resources" should be declared in the OPF file.
ERROR(RSC-006): /srv/www/integrations.pressbooks.network/releases/20230118044402/web/app/uploads/sites/706/pressbooks/exports/Test-Shape-Shifter-1674061522.epub/EPUB/mcluhan.css(3,9): Remote resource reference not allowed; resource must be placed in the OCF.
ERROR(RSC-006): /srv/www/integrations.pressbooks.network/releases/20230118044402/web/app/uploads/sites/706/pressbooks/exports/Test-Shape-Shifter-1674061522.epub/EPUB/mcluhan.css(4,9): Remote resource reference not allowed; resource must be placed in the OCF.

Check finished with errors

_Originally posted by @tw77 in https://github.com/pressbooks/private/issues/1080#issuecomment-1387423329_

richard015ar:
I am not able to replicate the validation issues mentioned.
Nevertheless, I can see new ones:
```bash
INFO(CSS-007): /app/web/app/uploads/sites/39/pressbooks/exports/LTI-Demo-Book-1674495306.epub/EPUB/mcluhan.css(5,3): Font-face reference "EPUB/assets/Lora-Bold.ttf" refers to non-standard font type "font/sfnt".
INFO(CSS-007): /app/web/app/uploads/sites/39/pressbooks/exports/LTI-Demo-Book-1674495306.epub/EPUB/mcluhan.css(11,3): Font-face reference "EPUB/assets/Lora-BoldItalic.ttf" refers to non-standard font type "font/sfnt".
[...]

PR #3140, opened 2023-01-23
Fix/epub font validation

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/3136

This PR specifies the media type for ttf fonts following the [EPUB 3.2 specs](https://www.w3.org/publishing/epub3/epub-spec.html#sec-cmt-supported).

### Context
For `ttf` fonts we should specify `font/ttf` or `application/font-ttf media`, see: https://github.com/w3c/epubcheck/issues/607#issuecomment-646350691

We get the media type from PHP using [finfo_file](https://www.php.net/manual/en/function.finfo-file.php) -using [mime_type](https://www.php.net/manual/en/function.mime-content-type.php) or `FILEINFO_MIME_TYPE ` constant, instead `FILEINFO_MIME` would be enough in https://github.com/pressbooks/pressbooks/blob/324733bcbe828b5592db0f3a6aec33bbe9645793/inc/media/namespace.php#L191. But this method is widely used-.  It returns `font/sfnt` for ttf fonts, which is correct. But EPUB 3.2 specs, expect a different format for the same mime type: `font/ttf` or `application/font-ttf media` for ttf fonts.

### Testing case
Export books to EPUB with multiple fonts. You should not see validation errors, except those from the book content itself.

One way to check validations is to watch logs while exporting:
```bash
lando ssh
tail -f /app/web/app/debug.log
```
Or you can enable the email logs.


richard015ar:
It contains PHP 8.1 changes. When I started it I was expecting to have the PHP 8.1 PR merged. Fixing it.

PR #3141, opened 2023-01-23
fix: ttf mime type following epub 3.2 specs

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/3136

This PR specifies the media type for ttf fonts following the [EPUB 3.2 specs](https://www.w3.org/publishing/epub3/epub-spec.html#sec-cmt-supported).

### Context
For `ttf` fonts we should specify `font/ttf` or `application/font-ttf media`, see: https://github.com/w3c/epubcheck/issues/607#issuecomment-646350691

We get the media type from PHP using [finfo_file](https://www.php.net/manual/en/function.finfo-file.php) -using [mime_type](https://www.php.net/manual/en/function.mime-content-type.php) or `FILEINFO_MIME_TYPE ` constant, instead `FILEINFO_MIME` would be enough in https://github.com/pressbooks/pressbooks/blob/324733bcbe828b5592db0f3a6aec33bbe9645793/inc/media/namespace.php#L191. But this method is widely used-.  It returns `font/sfnt` for ttf fonts, which is correct. But EPUB 3.2 specs, expect a different format for the same mime type: `font/ttf` or `application/font-ttf media` for ttf fonts.

### Testing case
Export books to EPUB with multiple fonts. You should not see validation errors, except those from the book content itself.

One way to check validations is to watch logs while exporting:
```bash
lando ssh
tail -f /app/web/app/debug.log
```
Or you can enable the email logs.

### DIFF
--- a/inc/modules/export/epub/class-epub.php
+++ b/inc/modules/export/epub/class-epub.php
@@ -642,14 +642,11 @@ class Epub extends ExportGenerator {
 		exec( $command, $output, $return_var );
 
 		foreach ( $output as $k => $v ) {
-			if ( strpos( $v, 'Picked up _JAVA_OPTIONS:' ) !== false ) {
-				// Remove JAVA warnings that are not actually errors
-				unset( $output[ $k ] );
-			} elseif ( strpos( $v, 'non-standard font type application/x-font-ttf' ) !== false ) {
-				// @see https://github.com/IDPF/epubcheck/issues/586, https://github.com/IDPF/epubcheck/pull/633
-				unset( $output[ $k ] );
-			} elseif ( strpos( $v, 'non-standard font type application/font-sfnt' ) !== false ) {
-				// @see https://github.com/w3c/epubcheck/issues/339
+			if (
+				str_contains( $v, 'Picked up _JAVA_OPTIONS:' ) ||
+				str_contains( $v, 'non-standard font type application/x-font-ttf' ) ||
+				str_contains( $v, 'non-standard font type application/font-sfnt' )
+			) {
 				unset( $output[ $k ] );
 			}
 		}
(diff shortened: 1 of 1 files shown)


inc/modules/export/epub/class-epub.php
```diff
@@ -642,14 +642,11 @@ public function validateGenerator() : \Generator {
...
 		foreach ( $output as $k => $v ) {
-			if ( strpos( $v, 'Picked up _JAVA_OPTIONS:' ) !== false ) {
-				// Remove JAVA warnings that are not actually errors
-				unset( $output[ $k ] );
-			} elseif ( strpos( $v, 'non-standard font type application/x-font-ttf' ) !== false ) {
-				// @see https://github.com/IDPF/epubcheck/issues/586, https://github.com/IDPF/epubcheck/pull/633
-				unset( $output[ $k ] );
-			} elseif ( strpos( $v, 'non-standard font type application/font-sfnt' ) !== false ) {
-				// @see https://github.com/w3c/epubcheck/issues/339
+			if (
```
richard015ar, line 645: This is just a quick refactor. It is not related to the fix.

SteelWagstaff:
@richard015ar am I correct in understanding that this PR fixes the problem described in this issue: https://github.com/pressbooks/pressbooks-book/issues/1119 rather than https://github.com/pressbooks/pressbooks/issues/3136?

richard015ar:
> @richard015ar am I correct in understanding that this PR fixes the problem described in this issue: [pressbooks/pressbooks-book#1119](https://github.com/pressbooks/pressbooks-book/issues/1119) rather than #3136?

Correct! Thank you! I've fixed it in the description. For some reason I can not connect it with the main issue on ZenHub.

fdalcin (approved):
LGTM.

Tested locally and got no validation errors related to the font mime.

Issue #2589, opened 2022-01-13
Add support for GA4


SteelWagstaff:
Google Analytics 4 has replaced Universal Analytics as the default GA tool for new properties: https://support.google.com/analytics/answer/10089681?hl=en.

GA 4 properties begin with `G-` prefixes, whereas University Analytics properties begin with `UA-` prefixes. We currently only support UA- prefixes: https://github.com/pressbooks/pressbooks/blob/dev/inc/admin/analytics/namespace.php. 

We should add support for GA 4 to Pressbooks, either by  

> If you use a web platform that requires you to provide a "UA-" ID  (i.e. won't accept a "G-" ID), you can still use Google Analytics 4 by pasting the global site tag (gtag.js) snippet manually into your website builder's/CMS's custom HTML field. Select your website builder from this table for platform-specific instructions.

or by actually supporting `G-` IDs and GA 4 features. See https://wordpress.org/plugins/ga-google-analytics/ for a well regarded plugin which does this.

tw77:
GA4 support requested by Western Sydney University (internal PB reference only): https://pressbooks.zendesk.com/agent/tickets/9580

SteelWagstaff:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pressbooks-Product-Backlog-60ae7d807f6e370010b3d5be/issues/pressbooks/pressbooks/2589?planning-poker) with ZenHub @arzola @cmurtagh @fdalcin @ho-man-chan @richard015ar @tw77

christopherphillips:
It is great to see this on the radar to add support for GA4 - is there any consideration to using [Google Tag Manager](https://support.google.com/tagmanager/answer/6102821?hl=en) as  a way to implement Google Analytics 4 which would provide a way for institutions to also collection additional variables? See 
[Track Custom Dimensions through Google Analytics](https://pressbooks.community/t/track-custom-dimensions-through-google-analytics/2262)

SteelWagstaff:
Desired changes for options page between now and July 1, 2023:
![Screenshot from 2023-01-24 10-50-04](https://user-images.githubusercontent.com/13485451/214382177-2c4aa9aa-606f-4c28-93b4-5936ec37f065.png). 
Description for first option should read:
> The Google Analytics UA ID for your network, e.g. ‘UA-01234567-8’. Google will [stop processing data](https://support.google.com/analytics/answer/11583528) for sites which use UA on July 1, 2023.

[...]

PR #3144, opened 2023-01-25
feat: google analytics 4 compatibility

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/2589

This PR adds a new input field for Google Analytics ID v4. 

A new class was added to manage Google Analytics settings and fields, which can be reused here: 

#### Network
- Go to Network Admin > Settings > Google Analytics
- You should be able to see 2 inputs fields, one for each GA version. Add one or both GA IDs and save them.
- Inspect the HTML in admin pages, you should be able to see v3 or v4 Google Analytics integrations, depending which one you added. BOTH CAN BE ADDED. Examples of how them look like:
Version 3:
```html
<!-- Google Analytics -->
<script>
(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
ga('create', 'UA-1234567-8', 'auto');
ga('send', 'pageview');
</script>
<!-- End Google Analytics -->
```
Version 4:
```html
<!-- Google Analytics -->
<script async="" src="https://www.googletagmanager.com/gtag/js?id=G-96TN18KRQR"></script>
<script>
window.dataLayer = window.dataLayer || [];
function gtag(){dataLayer.push(arguments);}
gtag('js', new Date());
gtag('config', 'G-96TN18KRQR');
</script>
<!-- End Google Analytics -->
```

#### Book level
- You will need to activate the subdomain installation constant, you can do it by changing this constant: https://github.com/pressbooks/setup-development-environment/blob/8cc6c83fc99dcfb5574d388ec4104b801a182ecb/config/application.php#L153
- Now, go to Go to Network Admin > Settings > Google Analytics and activate the checkbox: `Site-Specific Tracking`.
- Go to some book where you are an administrator, and go to Book Admin > Settings > Google Analytics
- Add 1 or both GA ID.
- Go to the book home page and you should be able to see the GA integrations there. Those are similar to the code displayed above, but with `'bookTracker'` parameter in the `ga` JS params.

#### Review data was properly sent to Google Analytics
I would need to test it, it is not possible to test it from local environment.


inc/class-googleanalytics.php
```diff
@@ -0,0 +1,422 @@
...
+
+	public static string $google_id_v4_option = 'ga_4_mu_uaid';
+
+	private string $v3_input_legend;
+
+	private string $v4_input_legend;
+
+	private string $v3_input_label;
+
+	private string $v4_input_label;
```
fdalcin: This could be replaced by an `array` I think. 

```php
$ga_versions = [
  'v3' => [
    'label' => ...,
    'legend' => ...,
    'id_option' => ...,
  ],
  'v4' => [
    'label' => ...,
    'legend' => ...,
    'id_option' => ...,
  ]
]
```

We could replace the methods to return data from the array instead of having multiple variables declared. What do you think?

inc/class-googleanalytics.php
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	public static function getGoogleIDSiteOption( int $version, bool $for_book ): false|string {
+		return ! $for_book ?
+			get_site_option( $version === 4 ? self::$google_id_v4_option : self::$google_id_v3_option ) :
+			get_option( $version === 4 ? self::$google_id_v4_option : self::$google_id_v3_option );
+	}
+
+	public function getInputLabel( int $version ): string {
+		return $version === 4 ? $this->v4_input_label : $this->v3_input_label;
```
fdalcin: Here we could use the array structure by passing the expected version

```php
return $ga_versions[ $version ]['label'];
```

inc/class-googleanalytics.php
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	public function getInputLegend( int $version ): string {
+		return $version === 4 ? $this->v4_input_legend : $this->v3_input_legend;
```
fdalcin: Same as above

inc/class-googleanalytics.php
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	static public function init(): GoogleAnalytics {
```
fdalcin: This should be `public static` to follow best practices

inc/class-googleanalytics.php
```diff
@@ -0,0 +1,422 @@
...
+
+		add_settings_field(
+			self::$google_id_v3_option,
+			$this->v3_input_label,
+			[ $this, 'analyticsInputCallback' ],
+			$this->network_page,
+			$section,
+			[
+				'legend' => $this->v3_input_legend,
+				'version' => 3,
```
fdalcin: I'd prefer having the version number as a constant so we don't have to guess what the number is in the future when passing parameters.

```php

public const GA_VERSION_3 = 3;
public const GA_VERSION_4 = 4;

```

And then when we have to use it, we could simply use
```php
GoogleAnalytics::GA_VERSION_4;
```

inc/class-googleanalytics.php
```diff
@@ -0,0 +1,422 @@
...
+		?>
+		<div class="wrap">
+			<h2><?php _e( 'Google Analytics', 'pressbooks' ); ?></h2>
+			<?php
+			$nonce = ( ! empty( $_REQUEST['_wpnonce'] ) ) ? $_REQUEST['_wpnonce'] : ''; // @codingStandardsIgnoreLine
+			if ( ! empty( $_POST ) ) {
+				if ( ! wp_verify_nonce( $nonce, 'pb_network_analytics-options' ) ) {
+					wp_die( 'Security check' );
+				} else {
+					$this->saveNetworkIDOption( 'ga_3', 3 );
```
fdalcin, line 284: Do we need both parameters? Couldn't we guess/build the key based on the version number?

fdalcin (changes requested):
Testing locally and seems to be working as expected. I'm able to update both keys and they are reflected in the frontend.

I added a couple of comments with some ideas (not required), let me know what you think. If you feel like there's no need to change, I'm ok merging as it is.

richard015ar:
@fdalcin I agree with your suggestions. I thought about the array of private properties but decided on granularity instead. But I think it's okay a static array would also work fine.
Rest of changes I'll submit them today, thanks.

For https://github.com/pressbooks/pressbooks/pull/3144#discussion_r1088135478, I think would be better to allow a custom key (string) to give independence & flexibility to the front end to use whatever 'name' or key in the inputs. Otherwise, we will force the interface to use the key we expect in GA class.

fdalcin (approved):

PR #3135, opened 2023-01-19
fix: encode PDF and EPUB fields for export routine

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1038

This PR renders decoded book fields for PDF and EPUB export routines, allowing to include or exclude ampersands in some cases when needed, especially on EPUB exports.
It also decodes the footer for PDF for titles with ampersands.

### Testing case
We want to test HTML tags and `&` in some books fields and inspect the output for EPUB and PDF exports.
- Pick a book and add `<em>` or similar allowed tags to the title of a chapter. 
- Add titles to the chapter (Headings 1) with `&` within them.
- Add a new contributor to the book and ensure the name contains `&`.
- For a chapter, adds the just created contributor as an author of the chapter.
- Add the contributor as illustrator, editor, author and other contributors' role for the book (In the Book Info page).
- In Appearance > Theme Options > Global Options, ensure `Enable two-level table of contents (displays headings under chapter titles)`, `Display information about authors at the end of each chapter` and `Display attributions at the end of a chapter` are enabled.
- In the PDF Options tab, use the following settings:
    Screen Shot 2023-01-19 at 2 40 47 PM
- Export the book to PDF and EPUB.

#### Analyzing the exports
Make sure everything looks as expected.
In the TOC, check the titles, authors and sections, especially those with ampersands.
At the footer of each page, you be able to see the chapter titles. Check the ones you added an ampersand within the title.
Check the About the authors' section at the end of each chapter, especially the one with the author that contains an ampersand in it.

PR #3145, opened 2023-01-26
fix: improve ampersand replacement regex

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1038

This PR improves the regex to replace `&` in certain metadata for EPUB and PDF export. Previously we only replaced the character when it is in the middle of a string (as usually expected), but it does not replace it if the character is at the end or at the beginning.

For context and more use cases for testing, please see https://github.com/pressbooks/pressbooks/pull/3135

### Testing case
Use a similar use case described in https://github.com/pressbooks/pressbooks/pull/3135 but add `&` at the end of titles, authors, contributors, parts, titles, etc.


fdalcin (approved):
LGTM, tested locally and seems to be working as expected.

Issue #3146, opened 2023-01-26
Encoding problems on organize page when title has special HTML chars


SteelWagstaff:
There are encoding issues on the organize page when the book title has special characters. See 
![Screenshot from 2023-01-25 16-17-38](https://user-images.githubusercontent.com/13485451/214723988-f36aa182-37e2-4e71-9eeb-d7c0097f55fd.png)
https://dev.pressbooks.pub/clonetest2/wp-admin/admin.php?page=pb_organize

_Originally posted by @SteelWagstaff in https://github.com/pressbooks/private/issues/1038#issuecomment-1404392072_

tw77:
I notice this fix on dev.pressbooks.network and it appears to solve this for book, chapter and part titles on the Organize page.

Just to note that there are similar encoding issues on the Organize page with author names:
<img width="182" alt="Screenshot 2023-01-26 at 10 57 34 AM" src="https://user-images.githubusercontent.com/44182629/214925119-50c79d27-e36f-448a-9585-4739229fa833.png">

Also, may be outside scope, but there are similar encoding issues in the chapter title in the chapter editor:
<img width="965" alt="Screenshot 2023-01-26 at 10 55 01 AM" src="https://user-images.githubusercontent.com/44182629/214924637-a43edf1b-5a9e-467c-897c-8d1925fa36c9.png">

SteelWagstaff:
Thanks @tw77 I can fix the author name issue with a new PR. @richard015ar and I noticed the encoding issue in the chapter title in the editor as well. I think that needs its own ticket.

SteelWagstaff:
New PR ready if @richard015ar or @tw77 is available to review: https://github.com/pressbooks/pressbooks/pull/3148. It's a tiny fix in a single line.

PR #3148, opened 2023-01-27
fix: improve author name encoding on organize page

SteelWagstaff:
Fix for https://github.com/pressbooks/pressbooks/issues/3146#issuecomment-1405467727

To test:
1. assign a contributor with special HTML characters in their name as an author of a chapter
2. Visit organize page and observe correct encoding

Before:
Screenshot from 2023-01-26 16-33-07

After:
Screenshot from 2023-01-26 16-32-51


fdalcin (approved):
LGTM

Issue #3151, opened 2023-01-28
Allow network managers to add many users to the network at once


SteelWagstaff:
As a super admin/network manager, I want to be able to add multiple users to my network as quickly and efficiently as possible.

All I want to provide for these new users are their email addresses. If an email address is already in use, I should be notified upon submission of the form. If an email address is not in use, a new account should be created using the part before the @ in the email address as their user name. If that user name is already in use, we should use current disambiguation methods for the user name. All newly invited users should receive invitations to set a password for their new accounts. Once they accept these invitations, these users should be added to the network as 'lowly users' (i.e. subscribers to the root site). 

I propose replaceing the current 'add users' page available to super admins )`https://YOURNETWORKURL/wp/wp-admin/network/user-new.php` with a simplified version of the bulk add user page we've added at the book level, i.e. what's visible at `https://YOURNETWORKURL/BOOKURL/wp-admin/users.php?page=user_bulk_new`.) This page should allow network managers to add many users as 'lowly users' all at once.
![Screenshot from 2023-01-25 12-41-10](https://user-images.githubusercontent.com/13485451/214685558-7c01a707-87ca-4031-9fbc-69d89cb9fb03.png)

See https://github.com/pressbooks/pressbooks/blob/dev/inc/admin/users/class-userbulk.php and https://github.com/pressbooks/pressbooks/blob/dev/templates/admin/user_bulk_new.blade.php for how the bulk add user page currently works in pressbooks.

## Requirements:
- [ ] Network managers can enter as many emails as they want into the form, one per line.
- [ ] Upon submission of the form, the invitation routine distinguishes between email addresses which already in use and those which are not. For those which are already in use, no action is taken. For those which are not in use, we attempt to use the part before the `@` in the email address as their user name. If that username is already taken, we use currently available disambiguation methods (add a # at the end, etc) to create a valid user name
- [ ] Upon submission of the form, the network network manager see sees a descriptive success/failure message (use the bulk user message for books as model)
- [ ] Upon submission of the form, new users receive the standard new user account email inviting them set a password and join the network
- [ ] Once an invited user sets a password and accepts the invitation, they are added to the network as a 'lowly user'

SteelWagstaff:
Ready for testing in dev environment.

tw77:
This is working as expected. The one thing I notice is that the message shown when attempting to add an email address that already exists on the network reads like a 'success' message:

> User(s) successfully added to the network.
> {email address}

[...]

PR #3161, opened 2023-01-31
feat: add bulk add user page for network admins

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/3151

This PR enables the `Users bulk add` page for network admins following the specs in the related issue.

### Testing case
- A new action in Network Admin > Users must be present: Bulk Add
- You should see Emails (textarea) and an `Add users` button.
- Enter multiple emails in the text area (1 per line).
- 1 activation email per user must be sent to the new users. You can see them in mailhog. You can check the mailhog URL by inspecting `lando info`.
- Activate 1 or more users.
- The activated users should contain `subscriber` role.

For more use cases, you could add invalid/repeated emails in the textarea. Also, you could try to add an email using the first part of the string, before the `@` with an existing username in your system.

### DIFF
--- a/inc/admin/users/class-userbulk.php
+++ b/inc/admin/users/class-userbulk.php
@@ -49,7 +49,9 @@ class UserBulk {
 	 * @param UserBulk $obj
 	 */
 	static public function hooks( UserBulk $obj ) {
-		add_action( 'admin_menu', [ $obj, 'addMenu' ] );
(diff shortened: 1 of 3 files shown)


fdalcin (approved):
LGTM, tested with a couple of emails and invalid emails/usernames, and the process works as expected.

PR #3172, opened 2023-02-06
fix: exclude integrations unit tests from tests suite

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1096

This PR excludes integrations tests from the tests suite to temporarily fix the pipeline issue we're having.

Permanent solutions to be implemented in https://github.com/pressbooks/private/issues/1860

For context, see: https://docs.google.com/document/d/16zFMx6so-27UQxhkGeOFMGBnAtIWESdsN4a5mij5QOA/edit#heading=h.281ury1ft3z5

### DIFF
--- a/.github/workflows/standards-and-tests.yml
+++ b/.github/workflows/standards-and-tests.yml
@@ -73,7 +73,7 @@ jobs:
       run: bash bin/install-wp-tests.sh wordpress_test root root localhost ${{ matrix.wordpress }}
 
     - name: Run PHP Tests
-      run: composer test
+      run: composer test -- --exclude=integrations
       if: matrix.experimental == true
 
     - name: Run PHP Tests and PCOV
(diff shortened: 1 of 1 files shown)


.github/workflows/standards-and-tests.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       run: bash bin/install-wp-tests.sh wordpress_test root root localhost ${{ matrix.wordpress }}
 
     - name: Run PHP Tests
-      run: composer test
+      run: composer test -- --exclude=integrations
```
richard015ar, line 76: not happy about this, but as we discussed it is a temporary fix.

fdalcin: Are we going to ignore integrations only for `experimental` pipelines?

fdalcin (approved):
LGTM

Issue #3173, opened 2023-02-07
Old-style user catalogs not displaying books


tw77:
### Description

Old-style user catalogs with the URL ending in `/catalog/{username}` are appearing as a blank page in production networks:

<img width="1429" alt="Screenshot 2023-02-07 at 9 46 17 AM" src="https://user-images.githubusercontent.com/44182629/217323858-cd6e5108-de81-4562-a8a8-094b55e006a0.png">

On dev.pressbooks.network, this stack trace is shown:

`Fatal error: Uncaught Error: Call to undefined function Pressbooks\analytics\print_analytics() in /srv/www/bedrocks/pressbooksedu-golden-bedrock/releases/20230202215401/web/app/plugins/pressbooks/templates/pb-catalog.php:170 Stack trace: #0 /srv/www/dev.pressbooks.network/releases/20230202215401/web/wp/wp-includes/template-loader.php(106): include() #1 /srv/www/dev.pressbooks.network/releases/20230202215401/web/wp/wp-blog-header.php(19): require_once('/srv/www/dev.pr...') #2 /srv/www/dev.pressbooks.network/releases/20230202215401/web/index.php(6): require('/srv/www/dev.pr...') #3 {main} thrown in /srv/www/bedrocks/pressbooksedu-golden-bedrock/releases/20230202215401/web/app/plugins/pressbooks/templates/pb-catalog.php on line 170`

On integrations.pressbooks.network, a 'There has been a critical error on this website' page is shown.

### Steps to Reproduce

1. Sign in to a network.
2. Navigate to 'My Catalog'
3. Include books to be displayed in this catalog, if there aren't any included already.
4. Click the public link to this catalog.

**Expected behavior:** We see the old user catalog with selected books shown.

**Actual behavior:** This catalog is blank or unavailable.

SteelWagstaff:
This broke because we removed the `inc/analytics/namespace.php` file in our 6.7.0 release. This was done as part of a GA refactor: https://github.com/pressbooks/pressbooks/pull/3144 but appears to have had unintended consequences, since this catalog page can't find the namespace it needs for this function anymroe: https://github.com/pressbooks/pressbooks/blob/c42987efb5016ab748f2f851297c0f156dbe7d07/templates/pb-catalog.php#L170

fdalcin:
[...]

PR #3174, opened 2023-02-07
fix: replace broken GA function on custom catalog page

SteelWagstaff:
Fix for https://github.com/pressbooks/pressbooks/issues/3173

To test:
1. Add GA4 and/or UA values in network settings
2. Open your individual catalog page at `https://pressbooks.test/catalog/admin`
3. Check to see that there's no fatal error and that the page head includes the expected GA scripts
4. Remove or alter GA4 and/or UA values
5. Repeat steps 2 and 3.


richard015ar (approved):
LGTM
Tested and works as expected, GA script is printed for both versions.

I added a friendly commit to use the init method (singleton instance) instead. 

Thanks!

Issue #3189, opened 2023-02-15
Fix location of noto sans fonts


SteelWagstaff:
Google archived their noto sans GitHub repo at the end of January 2023: https://github.com/notofonts/noto-fonts. These fonts are now being distributed at https://notofonts.github.io/ with https://github.com/notofonts/notofonts.github.io as the source repo. We should update our font URL references accordingly.

SteelWagstaff:
All of the changed Noto fonts appear to work as expected on dev networks for web and exports. I see a problem with Ancient Greek, but it's unrelated to this issue, so I'll file a new bug for that one.

PR #3190, opened 2023-02-15
fix: google fonts locations & tests

SteelWagstaff:
Fix for https://github.com/pressbooks/pressbooks/issues/3189

### DIFF
--- a/tests/test-globaltypography.php
+++ b/tests/test-globaltypography.php
@@ -84,6 +84,7 @@ class GlobalTypographyTest extends \WP_UnitTestCase {
 						$this->assertTrue( false, "Cannot download: {$url}" );
 					}
 					$this->assertStringContainsString( "/notofonts/noto-cjk/", $font_url );
+					continue;
 				}
 				if ( str_contains( $baseurl, 'https://cdn.jsdelivr.net/gh/notofonts/notofonts.github.io/fonts/' ) ) {
 					if ( $headers && isset( $headers['content-type'] ) ) {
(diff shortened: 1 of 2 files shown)


tests/test-globaltypography.php
```diff
@@ -77,12 +77,22 @@ public function test_fontPacks() {
 			foreach ( $val['files'] as $font => $font_url ) {
 				$url = $baseurl . $font_url;
 				$headers = wp_get_http_headers( $url );
-				if ( $headers && isset( $headers['location'] ) ) {
-					$font_url = $headers['location'];
-				} else {
-					$this->assertTrue( false, "Cannot download: {$url}" );
+				if ( str_contains( $baseurl, 'https://github.com/notofonts/noto-cjk/raw/main/Sans/OTF/' ) ) {
```
arzola, line 80: I have a refactor suggestion about those conditionals

```php
if (str_contains($baseurl, 'https://github.com/notofonts/noto-cjk/raw/main/Sans/OTF/')) {
    $font_url = $headers['location'] ?? $this->assertTrue(false, "Cannot download: {$url}");
    $this->assertStringContainsString('/notofonts/noto-cjk/', $font_url);
     continue; // we don't want to check again in the next if so we can end this iteration and move to the next one
}

if (str_contains($baseurl, 'https://cdn.jsdelivr.net/gh/notofonts/notofonts.github.io/fonts/')) {
    $content_type = $headers['content-type'] ?? $this->assertTrue(false, "Cannot download: {$url}");
    $this->assertEquals('font/otf', $content_type);
}
```

SteelWagstaff: love it. Didn't know about `continue;` -- will apply it now

arzola (approved):
LOKTM

Issue #3183, opened 2023-02-14
Update dashboard and nav icons


SteelWagstaff:
We've begun using new icons for several items in the new dashboards and network manager side-level nav. We should also update the side nav icons for the book dashboard at the same time, for consistency. Wireframe in Figma.

## Notes
It would be great if we could use Heroicons & Michelle' custom SVG icons seamlessly as if they were the regular dashicons. Something like: https://giteroptimized.com/how-to-remove-dashicons-in-wp/ might work here?

arzola:
I submitted a new branch with the first iteration of this to discuss https://github.com/pressbooks/pressbooks/tree/feat/icons-integration

SteelWagstaff:
@arzola https://github.com/pressbooks/pressbooks/pull/3193#issuecomment-1432274973
> this is a really interesting approach -- the problem I'm seeing now is that the markup results in the SVG files being referenced as part of the src bath for an img elements:

> This is a problem because I can't apply CSS to change the fill color of the SVG file for hover and active states. I think we need these icons to be rendered as SVG elements or as characters within a font-face (the way WP does this with dashicons), rather than <img> elements, so that we can alter fill. See https://stackoverflow.com/a/30419873

arzola:
@SteelWagstaff yes I agree I was thinking how to properly map those svgs because they don't provide a webfont format but actually I found the following https://github.com/savatia/heroicons-css so probably this would be our best bet

arzola:
Well the heroicons-css is pretty outdated and a lot of new icons are not included, I tried to generate the font/svg map with `fantasticon` but the generated files are corrupted

SteelWagstaff:
Ideas: https://blog.adobe.com/en/publish/2017/11/15/weekoficons-prepare-svg-icons-for-icon-fonts + https://mediatemple.net/blog/design-creative/creating-implementing-icon-font-tutorial/. Might be better, more modern tooling, but that's the basic concept.

SteelWagstaff:
@arzola spoke with Michelle about this task. She said she's capable of turning icons into a font family as described in the Adobe article and could do it in an afternoon, if desired. I told her we'd talk about it at stand up and get back to her on that. Let me know if useful?

arzola:
That would be amazing,  probably it will be limited to the icons that we are using only, but sounds really helpful

arzola:
[...]

PR #3193, opened 2023-02-15
Feat/icons integration

SteelWagstaff:
Fix for https://github.com/pressbooks/pressbooks/issues/3183


SteelWagstaff:
@arzola this is a really interesting approach -- the problem I'm seeing now is that the markup results in the SVG files being referenced as part of the src bath for an img elements:
Screenshot from 2023-02-15 16-13-22
This is a problem because I can't apply CSS to change the fill color of the SVG file for hover and active states. I think we need these icons to be rendered as SVG elements or as characters within a font-face (the way WP does this with dashicons), rather than `` elements, so that we can alter fill. See https://stackoverflow.com/a/30419873

PR #3205, opened 2023-02-23
fix: reorder view option on topbar

fdalcin:
When editing a chapter of the book the "View Chapter" option is placed first in the top bar options when it should be placed last. This PR addresses that by updating the order we display menu options.

**Before**
image

**After**
image


SteelWagstaff (approved):
:+1:

Issue #3152, opened 2023-01-28
Improve top-level navigation for all Pressbooks users


SteelWagstaff:
As a network manager, I want to clearly see and understand the various high-level administrative options I have from my top-level navigation menu (administer network, my books, create a book, clone a book, add users, edit my profile).

As a regular end user of Pressbooks, I want to be able to see a list of all the books I have accounts in and key actions I can take (create a book, clone a book, edit my profile)

To that end, we should simplify and improve the top-level navigation menu to make it easier for users to find and perform the actions they want to do.

## Design notes 
See https://www.figma.com/file/CU0d5QeDcCRK9LK2xUuYQp/Network-Manager-Dashboard?node-id=640%3A503&t=m955UZbuNqroGbCK-0 and https://docs.google.com/document/d/1227gW5CmccJNUhxfTduViB3kkKpzJKd_UdWP49oE5pk/edit

Key features:
- [x]  remove PB logo menu for all users
- [x]  Combine the two admin menus visible for super admins & network managers into a single menu called 'Administer Network'
- [x]  Display a my books menu for all users. Clicking on it opens new user dashboard. If they have no books, that's all it does. If they have books, display books in list like we do now. Clicking on a book opens book dashboard like it does now.
- [x]  Add new 'create a book' and 'clone a book' buttons for all users with permissions to do this
- [x]  Add new 'add users' button for network managers. Clicking this button should take them to a new 'bulk add user' page at network level that is a simplified version of our current bulk add user page at the book level. Let's them enter email addresses to add many users as 'lowly users' all at once.
- [x]  Change appearance of the 'profile' link in top right

SteelWagstaff:
I pushed several CSS changes to the WiP branch. I think this could still use a bit of clean up and consolidation, but it's getting closer to my desired appearance.

SteelWagstaff:
This has been deployed to dev environments and is ready for testing/review by @tw77 and @MichelleWeremczuk.

tw77:
As a regular user who hasn't created any books yet, 'My Books' in the top nav links to the user dashboard:

<img width="1214" alt="Screenshot 2023-02-27 at 8 53 47 AM" src="https://user-images.githubusercontent.com/44182629/221628224-c84fcd81-37c0-4856-8c85-7db2e433b60f.png">

[...]

PR #3209, opened 2023-02-28
Improve styling of top nav

SteelWagstaff:
Fix for https://github.com/pressbooks/pressbooks/issues/3152#issuecomment-1447210560

To test:
1. Check to see if the PB logo appears in top nav bar for all logged in users. Link should point to pressbooks.com. Screen reader text should read 'About Pressbooks'
2. Check to see if profile avatar receives thin white border upon hover, focus, and active states
3. Check to see if profile avatar has more right padding

Screenshot from 2023-02-27 17-33-33


fdalcin (changes requested):
LGTM, however I have a couple of suggestions.

Could we define the border as `2px solid transparent` to the profile and then just change the colour on hover? Feels weird that when we hover the profile it slightly moves things to the left.

Also, would it be possible to adjust the PB icon on mobile? Hide it on mobile devices perhaps?

fdalcin (approved):

PR #3211, opened 2023-03-01
feat: use heroicons

arzola:
This PR adds the ability to use heroicons seamlessly as if they were dashicons. Fix for https://github.com/pressbooks/pressbooks/issues/3183

This PR removes heroicons svg files in favour of a new heroicons custom webfont, this way will allow us to use the heroicons svgs using an `<i>` tag this would allows to use the heroicons SVGs as if they were fonts.

```html
<i class="pb-heroicons pb-heroicons-book-open"></i>
```

I also included a css map (heroicons-bridge.scss) to override WP dashicons with the heroicon version, for instance if we try to use the class `dashicons-dashboard` that :before content icon will be replaced by `pb-heroicons-home`

### WP Menu example

```php
add_menu_page( 
__( 'Publish', 'pressbooks' ), 
__( 'Publish', 'pressbooks' ), 
'edit_posts', 'pb_publish', 
[ $page, 'render' ], 
'dashicons-admin-users', 
17 );
// dashicons-admin-users will be replaced by pb-heroicons-users automagically
```

Note: This PR removes the Icons class used to locate SVG files


richard015ar (changes requested):
Thanks, @arzola. We reviewed it with @fdalcin and we found some issues that would need to be fixed:

- Add users icon should be solid not outlined in the topbar
- Analytics icon is missing in the network dashboard "Updated Homepage" section
- Topbar active book `text-indent` moves the icon all the way out of `visible` area screen and `(max-width: 782px)`. Inspecting it we see:
```css
#wpadminbar #wp-admin-bar-my-sites > .ab-item, 
#wpadminbar #wp-admin-bar-site-name > .ab-item, 
#wpadminbar #wp-admin-bar-site-editor > .ab-item, 
#wpadminbar #wp-admin-bar-customize > .ab-item, 
#wpadminbar #wp-admin-bar-edit > .ab-item,
#wpadminbar #wp-admin-bar-my-account > .ab-item
```
It seems to be the WP core rule that can be overridden.
- Changes in class-sidebar.php seems to be an older version which contains some bugs that were fixed. You would need to update this file to the latest version.

fdalcin (approved):
Thank you @arzola, this looks good to us. @richard015ar and I reviewed the PR and added just a friendly commit because one icon was not being replaced properly in all cases.

We did some exploratory testing checking different dashboards and sidebars with different plugins enabled (e.g. koko analytics, network analytics)

PR #3222, opened 2023-03-06
Fix heroicon color

SteelWagstaff:
This PR changes the color of the newly added heroicons on the network manager and book dashboards to match the Figma board. 

To test:
1. Visit the network manager and book dashboard and observe the heroicons are #3c434a (grey);
2. Hover over the link or tab through so that the link is in the active state. Observe the icons are now #1a1d20; (darker grey)


arzola (approved):
tested LGTM

PR #3227, opened 2023-03-07
Move network admin notices dashboard

SteelWagstaff:
Partial fix for https://github.com/pressbooks/pressbooks-saml-sso/issues/144. This PR moves the WordPress admin notices on the new network manager dashboard so that they reliably appear below the 'This network has X books and Y users' declaration in the second row.


templates/admin/dashboard/network.blade.php
```diff
@@ -26,6 +25,7 @@ class="button button-primary"
 						</a>
 					@endif
 				</div>
+				<div class="wp-header-end"></div>
```
SteelWagstaff, line 28: WordPress uses the .wp-header-end class as a reference point for placing admin notices. See https://github.com/WordPress/WordPress/blob/4959927285104dd4b032ffabd9fe5f58a0fe9db7/wp-admin/js/common.js#L1081-L1091

richard015ar (approved):
LGTM!

Issue #3231, opened 2023-03-09
No scroll in 'My Books' dropdown


tw77:
### Description

The dropdown menu of books under 'My Books' in the top nav bar is no longer scrollable. This is an obstacle when a user belongs to many books, exceeding the number that can be shown onscreen at once.

<img width="701" alt="Screenshot 2023-03-09 at 7 25 03 AM" src="https://user-images.githubusercontent.com/44182629/224071147-4ebca9d1-3f85-4333-b6de-8114431f00f7.png">

### Steps to Reproduce

1. Belong to many books (30 or more to replicate at 100% zoom)
2. Hover over 'My Books' in the top nav bar of the dashboard.

**Expected behavior:** We can scroll up and down the 'My Books' dropdown menu.

**Actual behavior:** We don't have the ability to scroll up and down this list.

SteelWagstaff:
We fixed this previously in https://github.com/pressbooks/pressbooks/issues/1901. See https://github.com/pressbooks/pressbooks/pull/1906 + https://github.com/pressbooks/pressbooks/pull/1900

tw77:
Fix deployed and confirmed working.

PR #3234, opened 2023-03-09
fix: restores scroll behavior to my books dropdown

fdalcin:
Issue #3231 

This PR restores the scroll behaviour to the "My Books" dropdown.

image

**How to test**

1. Add the same user to multiple books
2. Hover the "My Books" dropdown and make sure the list is scrollable if not all books are showing

> **Note**
> If you don't have a good amount of books in your local network you can use the following snipped to mimic a large list of books. Simply replace [these lines](https://github.com/pressbooks/pressbooks/blob/3f8f9dfdb4266e94eeeae635fe6fe4e8243aff80/inc/admin/menus/class-topbar.php#L221-L226) in the `class-topbar.php` with the following code:
```php
foreach (range(1, 50) as $id) {
	$bar->add_node([
		'parent' => 'pb-my-books-list',
		'id' => "book-{$book->userblog_id}-{$id}",
		'title' => "<span class='blavatar'></span> {$title}",
		'href' => get_admin_url($book->userblog_id),
	]);
}
```


richard015ar (approved):
LGTM.

Tests work following the code replacement suggested, since I did not have enough book in my local environment.

Thanks @fdalcin !

Issue #3241, opened 2023-03-13
Book creation for super admins / network managers unintentionally limited by network setting


tw77:
### Description

There is a bug at the moment where unchecking "Allow registered users to create and clone new books" in **Network Settings > Book & User Registration Settings** also removes the ability for super admins / network managers to create new books, hiding the Create Book buttons from them in the top nav and the dashboard.

This network setting is only intended to limit regular users from book creation, but it is being applied to super admins / network managers as well.

With this network setting unchecked, directly navigating to the book creation URL as a super admin / network manager does not work either, instead showing "Registration has been disabled" or "You are logged in already. No need to register again!" depending on whether the "Allow user self-registration" setting is checked.

### Steps to Reproduce

1. Log in as a super admin / network manager.
2. Uncheck "Allow registered users to create and clone new books" in **Network Settings > Book & User Registration Settings** 
3. Try to create a book.

**Expected behavior:** As a super admin / network manager I still see the Create Book buttons in the top nav and the dashboard, and I still have the ability to create books, while regular users on the network don't have this ability.

**Actual behavior:** 
<img width="1421" alt="Screenshot 2023-03-13 at 8 30 07 AM" src="https://user-images.githubusercontent.com/44182629/224749160-36f054b8-debb-4026-a4dd-27bb33e2d876.png">
Nobody on the network has the ability to create books or see the buttons for book creation. Navigating to the book creation URL either shows "Registration has been disabled" or "You are logged in already. No need to register again!" depending on whether the "Allow user self-registration" setting is checked.

tw77:
Testing fix on dev networks as a super admin with "Allow registered users to create and clone new books" unchecked, I notice that the Create Book button is visible again in the top nav and directs me to the intended URL.

However, the "Create a book" block in the "My Books" dashboard is still not appearing for me:  

<img width="1440" alt="Screenshot 2023-03-13 at 1 22 03 PM" src="https://user-images.githubusercontent.com/44182629/224823506-43f5af6b-0ac6-48e9-ab64-8183bb17a45d.png">

Also, I'd expect the "Clone Book" button in the top nav to appear in this scenario. That and the "Clone a book" block in the "My Books" dashboard aren't visible at the moment.

tw77:
[...]

PR #3242, opened 2023-03-13
fix: always allow super admins to create books

fdalcin:
Issue #3241 

This PR allows super admins to create books even when the option to create books is unchecked.

**Expected behaviour**

- Book creation is enabled: users should be redirected to the regular `{$NETWORK}/wp-signup.php` page to create a new book
- Book creation is disabled: super admins/network managers should be redirected to `{$NETWORK}/wp-admin/network/site-new.php` to create a new book.

**How to test**

1. Follow the steps described in the ticket to validated the expected behaviour


arzola (approved):
LGTM

PR #3245, opened 2023-03-14
fix: show create a book to super admins when book creation is disabled

fdalcin:
Issue #3241 

Following conversations on the ticket, I've opened this PR so the user dashboard blocks behave the same way the top bar buttons.

- **Create a book:** top bar button and user dashboard block should be displayed when book creation is enabled or if the user is a super admin/network manager. In cases where the option is disabled, the redirect link should be `{$NETWORK}/wp-admin/network/new-site.php`.
- **Clone a book:** top bar button and user dashboard block should be displayed when cloner is enabled **AND** book creation is enabled or if the user is a super admin/network manager.

**How to test**

Play around with network settings, enabling book creation and the cloner. Make sure the buttons reflect the expected behaviour.

### DIFF
--- a/inc/admin/menus/class-sidebar.php
+++ b/inc/admin/menus/class-sidebar.php
@@ -34,7 +34,7 @@ class SideBar {
 			$this->usersSlug = 'pb_network_analytics_userlist';
 
 			$this->settingsCallback = [ \PressbooksNetworkAnalytics\Admin\Options::init(), 'printMenuSettings' ];
-			$this->settingsSlug = 'pb_network_analytics_options';
+			$this->settingsSlug = $this->getContextSlug( 'admin.php?page=pb_network_analytics_options', false );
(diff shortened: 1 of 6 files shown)


inc/admin/menus/class-sidebar.php
```diff
@@ -233,16 +233,27 @@ private function addMenuItems(): void {
...
 				__( 'Settings', 'pressbooks' ),
 				'manager_network',
-				$this->settingsSlug,
+				'settings.php',
 				$this->settingsCallback,
 				'dashicons-admin-settings',
 				7
 			);
+			if ( ! is_network_admin() ) {
+				add_submenu_page(
```
richard015ar, line 247: We decided with Steel that a submenu item works.

richard015ar (approved):
LGTM!

Issue #3250, opened 2023-03-15
Allow super admins/network managers to view old 'my catalog' page from URL


SteelWagstaff:
We have a legacy 'my catalog' feature that creates a custom page for user catalogs: `wp-admin/index.php?page=pb_catalog`. As of our latest release, super admins/network managers see a `Sorry, you are not allowed to access this page.` message. We want super admins and network manager to be able to view this page from the URL. We don't need to add this page to the sidebar menus, however.

We also want to restore the 'User Activation Keys' sub menu item to the User menu in the sidebar for super admins when the User Activation Keys plugin is active.

SteelWagstaff:
Tested and works as expected on dev networks. Ready for release. Thanks all!

PR #3251, opened 2023-03-15
fix: add pb catalog page & users activation keys submenu

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/3250

This PR enables the `pb_catalog` pages to make it accessible via URL.
It also adds the Users Activation Keys submenu item when the proper plugin is active, for super admins.

Note: it also adds the `Bulk Add` submenu item in Users menu when Network Analytics plugin is inactive.

We had a pair programming session with @fdalcin (reviewer), he knows the testing process of this.

### DIFF
--- a/inc/admin/menus/class-sidebar.php
+++ b/inc/admin/menus/class-sidebar.php
@@ -360,7 +360,7 @@ class SideBar {
 				__( 'User Activation Keys', 'pressbooks' ),
 				'edit_users',
 				'act_keys',
-				[ $ds_wp3_user_activation_keys, 'ds_delete_stale' ]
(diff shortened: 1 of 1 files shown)


fdalcin (approved):
LGTM, works as expected locally.

Tested accessing the **My Catalog** link at `{$NETWORK}/wp-admin/index.php?page=pb_catalog` and it works even though we don't show the link.

User activation keys is displayed as a submenu in the users menu and works as expected.

I tested activating/deactivating the following plugins: **User Activation Keys** and **Pressbooks Network Analytics**

PR #3252, opened 2023-03-16
fix: deprecation notices

SteelWagstaff:
This PR fixes PHP deprecation notices that display on CI test runs. It doesn't address the use of mb_convert_encoding in https://github.com/pressbooks/pressbooks/blob/8d8ed90ba313774146842cf438c6d32356a8e382/inc/class-htmlparser.php#L47. See https://php.watch/versions/8.2/mbstring-qprint-base64-uuencode-html-entities-deprecated#html for details on that one.


fdalcin (approved):
LGTM

Issue #3268, opened 2023-01-26
Implement autosave in Pressbooks


arzola:
As an author, I'd like in progress work to be auto-saved while in progress, so I don't lose my work if my browser crashes or internet connection is disrupted. 

### Feature Description
WordPress core has an autosave feature, in which drafts of posts/pages are automatically saved to the database as part of the revision history based on the AUTOSAVE_INTERVAL defined in the `wp.config.php` file. See https://jetpack.com/blog/wordpress-autosave/ + https://www.wpbeginner.com/glossary/revisions/

Pressbooks used to support this feature (I believe), but autosave appears to be broken, and is only triggered now when users 'preview' or 'save' their posts. See https://github.com/pressbooks/private/issues/1019#issuecomment-1284035146

Autosave in WordPress relies on 'Heartbeat': https://make.wordpress.org/core/2013/01/19/autosave-and-post-locking/. When we refactored the cloning page to allow users to clone from anywhere, we deregistered heartbeat on all admin pages instead of just the cloning page. See: https://github.com/pressbooks/pressbooks/commit/6677e6a32882493247473158aa4c049ae69a6e5e. If we were to only remote heartbeat from the cloning page, that should resolve this issue. See https://github.com/pressbooks/pressbooks/issues/3258 (open source user confirmed they could restore autosave functionality by commenting out this line: https://github.com/pressbooks/pressbooks/blob/7b7d3be450880c6af5cec7826142bc3c987e673e/inc/admin/laf/namespace.php#L638

### Other Notes

Some of our users were confused when autosave disappeared from Pressbooks, and were disappointed when they lost work beacuse the feature was no longer present.

Autosaving appears to work normally for me in the classic editor of my vanilla WordPress multisite install, which leads me to believe that it's not working in Pressbooks either because of our use of custom post types or because of how we've modified the core WordPress 'publish' block. 

WP core maintainers aren't aware of any problems with core WP related to this either:
![Screenshot from 2023-01-26 08-55-06](https://user-images.githubusercontent.com/13485451/214898871-107c1f40-c745-46ce-9ad2-b8f1b14037f3.png)

There is a plugin called [wp-autosave](https://wordpress.org/plugins/wp-autosave/?ref=commonninja) that adds the feature but is not working properly with our custom post types. We can update or fork the `wp-autosave` plugin in order to implement this feature for our users

SteelWagstaff:
[...]

PR #3269, opened 2023-04-03
Fix: restore heartbeat on non-import/export/cloning admin pages

SteelWagstaff:
This PR addresses https://github.com/pressbooks/pressbooks/issues/3268. We need to add a check to only add pb-clone scripts and remove the heartbeat scripts on the cloner page, instead of all admin pages.


richard015ar (approved):
LGTM!
I cloned a book from network and book-level pages and it worked. I also checked the autosave for a chapter and it works as expected. Thank you, @SteelWagstaff!

Screen Shot 2023-04-11 at 3 48 36 PM

PR #3274, opened 2023-04-12
chore: trigger bedrock updates

arzola:
This PR is just an update to allow triggering autoupdate actions when production branch is updated

https://github.com/pressbooks/private/issues/1128


fdalcin (approved):
LGTM

PR #3289, opened 2023-04-25
fix: authors cleaned up during collection

arzola:
Related https://github.com/pressbooks/pressbooks-network-catalog/issues/195

To test this

1. Save the book information adding some authors
2. Run the data collector to copy metadata on blog meta (https://github.com/pressbooks/pressbooks-network-analytics#setup)
3. Review the catalog has the authors after the data collector run
4. Review the endpoints has the authors as expected

### DIFF
--- a/inc/datacollector/class-book.php
+++ b/inc/datacollector/class-book.php
@@ -307,12 +307,6 @@ class Book {
 		$in_catalog = empty( get_option( get_in_catalog_option() ) ) ? 0 : 1;
 		update_site_meta( $book_id, self::IN_CATALOG, $in_catalog );
 
-		// pb_authors is being skipped by the getBookInformation() method when reading metadata from the database
-		// we need to check if it's empty and if so, read it from the database
-		if ( count( $metadata['pb_authors'] ) === 0 ) {
-			$metadata['pb_authors'] = get_site_meta( $book_id, self::AUTHORS );
-		}
-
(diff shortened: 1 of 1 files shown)


inc/datacollector/class-book.php
```diff
@@ -428,12 +428,13 @@ public function copyBookMetaIntoSiteTable( $book_id ) {
...
 	private function saveArrayMetadata( int $blog_id, string $meta_key, string $array_key, array $metadata ): void {
-		// clean up meta key before adding
-		delete_site_meta( $blog_id, $meta_key );
-
 		if ( isset( $metadata[ $meta_key ] ) && is_array( $metadata[ $meta_key ] ) ) {
 			foreach ( $metadata[ $meta_key ] as $value ) {
-				add_site_meta( $blog_id, $meta_key, $value[ $array_key ] );
+				// clean up meta key before adding
+				delete_site_meta( $blog_id, $meta_key, $value[ $array_key ] );
+				if ( is_array( $value ) ) {
```
arzola, line 435: I added this extra check because sometimes the $value is not an array

richard015ar (approved):
LGTM. Tests passed as expected.
I was checking why sometimes the metadata in this function would be as string and it might be the case when [it is in the cache](https://github.com/pressbooks/pressbooks/blob/26c681eb84de9e4298aed2be0d60a03036c28c90/inc/class-book.php#L107).

Issue #3299, opened 2023-05-01
Make images smaller, add TablePress icon


SteelWagstaff:
Several users have reported that the decorative images we've added to the network manager, user, and book dashboards are too large and are slightly distracting from the text and links around them. We should make them smaller.

Also, when book admins activate the TablePress plugin, no icon appears in the book dashboard. We should add the heroicons table icon.

SteelWagstaff:
Tested on dev. Images are smaller, but TablePress icon still does not appear. Perhaps we compiled the new font incorrectly?

SteelWagstaff:
Worked on dev, but i had to clear my browser cache for some reason. I thought fonts had cache busting built in, but i guess not somehow?

SteelWagstaff:
Ready for release

PR #3300, opened 2023-05-01
fix: make decorative dashboard images smaller, add TablePress icon

SteelWagstaff:
This PR provides a fix for https://github.com/pressbooks/pressbooks/issues/3299. Still need to add the actual icon to the heroicons font set as `\f11d` before this can be merged.


assets/src/styles/heroicons.scss
```diff
@@ -127,3 +127,7 @@ i[class^="pb-heroicons-"]::before, i[class*=" pb-heroicons-"]::before {
 .pb-heroicons-wrench::before {
   content: "\f11c";
 }
+
+.pb-heroicons-tablepress::before {
+  content: "\f11d";
```
SteelWagstaff, line 132: @arzola we need to add a new icon to the dashicons font with this character reference

arzola: I updated the font family and now includes the table cells icon

SteelWagstaff: what's the character reference for the new icon?

SteelWagstaff: Found it, nevermind

PR #3321, opened 2023-05-17
fix: repair internal hyperlinks after importing epub file

dominik:
Fix for #3326.

### DIFF
--- a/inc/modules/import/epub/class-epub201.php
+++ b/inc/modules/import/epub/class-epub201.php
@@ -207,12 +207,28 @@ class Epub201 extends ImportGenerator {
 		$selected_for_import = $this->selectedForImport( $xml, $match_ids );
 		$total = count( $selected_for_import );
 		$y = new PercentageYield( 40, 95, $total );
+
+        $posts = [];
+        $urls = [];
+
 		foreach ( $selected_for_import as $id ) {
 			yield from $y->tick( __( 'Importing', 'pressbooks' ) );
 			// Insert
 			$href = $this->basedir . $this->manifest[ $id ]['href'];
-			$this->kneadAndInsert( $href, $this->determinePostType( $id ), $chapter_parent, $current_import['default_post_status'] );
+			$pid = $this->kneadAndInsert( $href, $this->determinePostType( $id ), $chapter_parent, $current_import['default_post_status'] );
+			// Collect info for replacement of hrefs
+			$posts[$pid] = get_post( $pid );
+			$urls[$pid]['new'] = $posts[$pid]->guid;
+			$urls[$pid]['old'] = $href;
 		}
+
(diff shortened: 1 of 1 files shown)


inc/modules/import/epub/class-epub201.php
```diff
@@ -364,6 +384,29 @@ protected function kneadAndInsert( $href, $post_type, $chapter_parent, $post_sta
...
+
+	/**
+	 * Fix hrefs in content
+	 *
+	 * @param WP_Post $post
+	 * @param array $urls
+	 */
+	protected function fixHrefs( $post, $urls ) {
+
+		$use_relative_hrefs = false;
```
arzola, line 399: @dominik could you explain why this variable is needed, the following condition it will be always false so the str_replace never will be reached

arzola: I think it could be really helpful to add a test for this function because it doesn't contain side effects it's just mutating the post object, you can add this in test-utility.php

```php
use Pressbooks\Modules\Import\Epub\Epub201;

public function testFixHrefs()
    {
        // Create a mock Post object
        $post = $this->createMock(stdClass::class);

        // Set up the initial post_content
        $post->post_content = 'Test content with hrefs to be replaced';

        // Define urls to test
        $urls = [
            ['old' => 'old_url', 'new' => 'new_url'],
        ];

        // Create the object that has the fixHrefs method 
        $processor = new Epub201();

        // Use Reflection to set accessibility to the fixHrefs method
        $reflection = new ReflectionClass(Epub201::class);
        $method = $reflection->getMethod('fixHrefs');
        $method->setAccessible(true);

        // Call the fixHrefs method
        $result = $method->invokeArgs($processor, [$post, $urls]);

        // Check that the URLs have been replaced in the post_content
        $this->assertSame('Test content with href="new_url" and href=\'new_url\'.', $result->post_content);
    }
```

arzola (changes requested):

SteelWagstaff:
hi @dominik our dev team left two comments/suggestions for revision we'd like to have answered or addressed before this is merged. Are you willing and able to respond?

dominik:
I'm on vacation - will provide answers next week ;-)

Am Di., 30. Mai 2023 um 16:11 Uhr schrieb Steel Wagstaff <
***@***.***>:

SteelWagstaff:
hi @dominik is this PR still of interest for you? Our team is willing to re-review once changes/questions are addressed. If not, we'll close this as the PR is becoming stale.

dominik:
Yes - will see that I can aubmit an update until end of month ...

Am Mo., 27. Nov. 2023 um 19:24 Uhr schrieb Steel Wagstaff <
***@***.***>:

Issue #3312, opened 2023-05-09
Implement filters parameters for /books API endpoint


richard015ar:
As a developer I would like to get a filtered list of books from `<NETWORK_URL>/wp-json/pressbooks/v2/books` endpoint. These filters should allow me to specify certain values and operators for certain properties.

Example: retrieve books with `bookDirectoryExcluded` false, with licenses code different than `All Rights Reserved` and with `wordCount` metadata greater than 5000.

The parameters I'd like to use are:
- bookDirectoryExcluded [boolean]
- license code [string]
- name [string]
- wordCount [integer]

These filters must be inclusive and must accept different operators like: `!= / not contains, >/greater than, ==/equals, etc`.

Filters we should support are the following:
- T/F for booleans
- contains/does not contain for strings
- `>=`/`<=` for integers

DoD: extend/modify the current books API endpoint to accept and process these filters.

richard015ar:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pressbooks-Product-Backlog-60ae7d807f6e370010b3d5be/issues/pressbooks/pressbooks/3312?planning-poker) with Zenhub @arzola @fdalcin

richard015ar:
Update: I got a working version at https://github.com/pressbooks/pressbooks/tree/feat/extend-books-endpoint.

Right now in the books endpoint, we get the [paginated books](https://github.com/pressbooks/pressbooks/blob/9ea8cb18b7d11877062b4f2605f9d6de418ace57/inc/api/endpoints/controller/class-books.php#L331) ids from the `wp_blogs` table, and then, [we read most of the metadata](https://github.com/pressbooks/pressbooks/blob/9ea8cb18b7d11877062b4f2605f9d6de418ace57/inc/api/endpoints/controller/class-books.php#L257) from `wp_blogmeta`.

In order to filter books what I did as a POC (testing code) is to join the `wp_blogmeta` table, where title, license code, words count, etc are.
The testing code I did, works well.
The problem is that this data is from the data collector, populated by network analytics, which is not available for open source, so this solution won't work as it is.

So I looked at the source of these metadata, and it is in `wp_<BLOG>_posts` for `post_type = metadata `. In the data collector, we get it using the getBookInformation method, which switches the blog.

So, basically, these parameters won't make much sense for open source users, in fact, open source users shouldn't see much metadata in the book endpoint, except for the `bookDirectoryExcluded` field, which is differently handled.

[...]

PR #3331, opened 2023-05-25
feat: allow filters for books api endpoint

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/3312

This PR adds necessary changes to handle new API parameters for `books` endpoint. The parameters handled are:
- `license_code[]` (array) to filter books by multiple license codes. For example `Public Domain` or `CC BY`.
- `title[]` (array), to filter or exclude books by multiple titles. for example `LTI Demo` to get books containing `LTI Demo` string (insensitive case) or `-biology` to exclude books with `biology` within the title.
- `in_directory` (boolean) to filter books included or excluded from PB Directory. For example: `in_directory=true`.
- `words` (string) to filter books `>=` or `<=` to a certain number of words. For example: `words=gte_1000` will bring book with `>=1000` words, and `words=lte_60000` will bring books with `<= 60000` words.

### Testing case
Basic book API endpoint in your local: `https://pressbooks.test/wp/wp-json/pressbooks/v2/books`

Make sure you have 5 or more public books to consider basic cases.

`in_directory` parameter:
- Add some books to the catalog (Books list).
- In Network Level > Settings > Sharing & Privacy check the `Exclude non-catalogued public books from Pressbooks Directory.` option and save it.
- add `?in_directory=true` parameter to the API endpoint. You must see only books with `bookDirectoryExcluded: false` in your list of books.
- now, change it to `?in_directory=false` parameter to the API endpoint. You must see only books with `bookDirectoryExcluded: true` in your list of books.
- Now, in Network Level > Settings > Sharing & Privacy uncheck the `Exclude non-catalogued public books from Pressbooks Directory.` option and save it.
- Go to some of your books and some of them in Settings > Sharing and Privacy, set `Pressbooks Directory` option to `Yes, I want this book to be listed in Pressbooks directory`, and for some of them, set it to `No`.
- in the basic API book endpoint you should see some `bookDirectoryExcluded` field as `true` and others as `false`.
- Use `in_directory` parameter as true and then as false to ensure you get the expected results.

`license_code[]` parameters:
- Add different licenses to your books.
- Use `?license_code[]` parameter to filter them by license code. You could use multiple licenses like `?license_code[]=CC BY&license_code[]=Public Domain` for example.
- Make sure the licenses match with the parameters.

`titles[]` parameters:
- Filter books by multiple titles like `title[]=test&title[]=demo` to get books containing `test` or `demo` within the title.
- You can exclude some books by specifying strings you don't want to be contained within the title. For example `title[]=lti&title[]=-demo'

`words` parameter:
- Check the above description to see how you can use it.


inc/api/endpoints/controller/books/class-books.php
```diff
@@ -1,50 +1,32 @@
 <?php
 
-namespace Pressbooks\Api\Endpoints\Controller;
+namespace Pressbooks\Api\Endpoints\Controller\books;
```
arzola: I feel this should be Capitalized to keep uniformity

inc/api/namespace.php
```diff
@@ -141,7 +141,7 @@ function init_root() {
 	\Pressbooks\Metadata\init_book_data_models();
 
 	// Register Books
-	( new Endpoints\Controller\Books() )->register_routes();
+	( new Endpoints\Controller\books\Books() )->register_routes();
```
arzola: Also here

inc/api/endpoints/controller/books/class-booksquerybuilder.php
```diff
@@ -0,0 +1,80 @@
...
+		$limit = ! empty( $request['per_page'] ) ? $request['per_page'] : 10;
+		$offset = ! empty( $request['page'] ) ? ( $request['page'] - 1 ) * $limit : 0;
+
+		$this->placeholder_values[] = $offset;
+		$this->placeholder_values[] = $limit;
+
+		return $this;
+	}
+
+	private function setupParameters(): void {
```
arzola, line 57: fancy, I like it

arzola (changes requested):
LGTM nice solution, just a small change

Issue #3313, opened 2023-05-10
Network managers unable to reach the User Activation Keys page


tw77:
### Description

Network managers can no longer reach the User Activation Keys page even when directly navigating to the `https://{domain}/wp/wp-admin/network/admin.php?page=act_keys` URL. Only super admins still have access to this page.

It would be helpful for network managers to have the ability to activate users right away from this page. Two network managers have reported scenarios where the user account confirmation email was filtered by their institution's email provider. If they were able to access the User Activation Keys page, they could activate users who didn't receive their confirmation email with the activation link.

This is especially relevant because the page for adding users to the network doesn't include the option to skip the confirmation email, which could be a useful separate feature outside the scope of this issue.

### Steps to Reproduce

1. Sign in to a network as a network manager, not a super admin.
2. Navigate to `https://{domain}/wp/wp-admin/network/admin.php?page=act_keys`

**Expected behavior:** We're brought to the User Activation Keys page.

**Actual behavior:** 

<img width="1076" alt="Screenshot 2023-05-10 at 8 59 09 AM" src="https://github.com/pressbooks/pressbooks/assets/44182629/067d03b0-06ae-4d53-a07a-0105b2c3baad">

SteelWagstaff:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pressbooks-Product-Backlog-60ae7d807f6e370010b3d5be/issues/pressbooks/pressbooks/3313?planning-poker) with Zenhub @arzola @fdalcin @richard015ar

richard015ar:
Update: I was able to display it for network managers. But now, I can see in this from the super admin:
![Screen Shot 2023-05-29 at 8 57 32 AM](https://github.com/pressbooks/pressbooks/assets/13248424/39ce96e2-aa82-4932-be2e-d695f2db042d)

I will work on this today to fix it.

SteelWagstaff:
Network manager can now access the User Activations Keys page at `https://NETWORK.URL/wp/wp-admin/network/admin.php?page=act_keys` -- it does not appear in their user menu for network managers or super admin users, however. Was this by design @richard015ar? How difficult would it be to add this page as a sub-menu item for both types of users?

richard015ar:
@SteelWagstaff Yes, it was designed only to access it through the URL. We discussed it during the planning meeting and sized it as a 3 task.
[...]

PR #3336, opened 2023-05-29
fix: allow access to user activation keys page to network managers

richard015ar:
Issue: #3313 

This PR adds the User Activation Keys access for network managers and removes it as a submenu item for Users menu for super admin users.
The only way to access to the page is through `<NETWORK_URL>/wp-admin/network/admin.php?page=act_keys`.

### Testing case
- Log in as a super admin user.
- `User Activation Keys` submenu item should NOT be present in any menu item, including Users.
- You should be able to access it through `<NETWORK_URL>/wp-admin/network/admin.php?page=act_keys`.
- Log in as a Network Manager user.
- `User Activation Keys` submenu item should NOT be present in any menu item, including Users.
- You should be able to access it through `<NETWORK_URL>/wp-admin/network/admin.php?page=act_keys`.

### DIFF
--- a/inc/admin/menus/class-sidebar.php
+++ b/inc/admin/menus/class-sidebar.php
@@ -84,14 +84,12 @@ class SideBar {
 				'sites.php',
 				'users.php',
 				'users.php',
-				'users.php',
 			],
 			[
 				'pb_network_analytics_booklist',
 				'site-new.php',
 				'pb_network_analytics_userlist',
 				'user-new.php',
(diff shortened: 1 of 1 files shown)


fdalcin (approved):
LGTM.

Tested locally and am able to access the *User Activation Keys* page with both Super Admins and Network Managers by manually typing the URL. I don't see the submenu item in the sidebar as expected.

Issue #3213, opened 2023-03-02
in insertContent mode standard, standardsidebar, shaded and shadedsidebar textboxes should have <p/> inside <div/> like educational textboxes


thomasdumm:
In insertContent for standard, standardsidebar, shaded and shadedsidebar textboxes there should be `<p/>` tag inside `<div/>` like in the case of educational textboxes:

_pressbooks/assets/src/scripts/textboxes.js_

This would prevent leaving the textbox when pressing return and would lead to the same behavior as with educational textboxes.

<img width="408" alt="mceInsertContent" src="https://user-images.githubusercontent.com/3658808/222465858-332728cb-0661-4fc6-a47d-67d0855a80c2.png">

SteelWagstaff:
Ready for release!

PR #3337, opened 2023-05-29
fix: wrap textbox content in `p` tags

fdalcin:
Issue #3213

This PR aims to fix a bug where standard texboxes don't add a `p` tag around the content. This change prevents one from leaving the textbox area when pressing enter while writing the content, keeping the same behaviour as the Educational textboses.

**How to test**

1. Edit a book chapter
2. Choose to insert one of the following textboses: Standard, Standard (sidebar), Shaded, Shaded (sidebar)
3. Make sure the textbox content is wrapped with a `p` tag
4. When pressing enter you should be able to continue editing inside the textbox in a new line


SteelWagstaff (approved):
LGTM.

PR #3340, opened 2023-05-31
fix: transifex client installation and pot file generation

arzola:
This PR would use the latest transifex client written in GO to generate the .mo files and it will generate a pull request with the generated files.

Also this modifies the pot generation action to create a pull request instead


SteelWagstaff (approved):
:+1:

PR #3355, opened 2023-06-08
chore(l10n): update languages/pressbooks.pot

pressbooks-ops:
This pull request updates the POT file for the Pressbooks project.


arzola:
This is good I'll address the automerge in the following sprint, we hare having a race condition with all of the checks in place

Issue #3363, opened 2023-06-13
Add license exclusion Book API parameter option


richard015ar:
From: We added `licenses` GET parameters for the Book API endpoint. 
As a developer, I would like to exclude licenses by adding `-` sign, just like the title exclusion option. For example, to get all books except those with `CC BY` license, I would like to make the following GET call: `<BOOK_API_URL>?licenses[]=-CC BY`.

DoD: Implement the new Book API Parameter to exclude books with specific licenses.

SteelWagstaff:
Tested and works as expected on staging. Ready for release.

PR #3365, opened 2023-06-13
feat: add licenses exclusion API param for books endpoint

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/3363

This PR adds the possibility to prepend `-` characters in `license_code` GET parameter for `books` API endpoint in order to exclude books with specific licenses. 

### Test case
- Create books with multiple licenses
- Call the Book API Endpoint with an exclusion parameter, for example: `https://pressbooks.test/wp/wp-json/pressbooks/v2/books?license_code[]=-CC BY&license_code[]=-Public Domain` would exclude all books with `CC BY` and `Public Domain` licenses.


inc/api/endpoints/controller/books/class-books.php
```diff
@@ -151,7 +152,18 @@ public function get_collection_params() {
...
 			'validate_callback' => function ( $param, $request, $key ) {
-				return is_array( $param );
+				if ( ! is_array( $param ) ) {
+					return false;
+				}
+
+				$licensing = new Licensing;
+
+				$supported_types = $licensing->getSupportedTypes();
+				$supported_codes = array_map( fn ( $license ) => $license['abbreviation'], $supported_types );
```
arzola: I feel it could be handy to perform this operation in a method like #licensing->getSupportedCodes()

SteelWagstaff (approved):
LGTM. Thanks Ricardo!

PR #3414, opened 2023-07-24
fix: output copyright year in CC licenses (resolves #3249)

### LINKED ISSUE #3249: Copyright year not included in exports when copyright license other than default / All Rights Reserved is specified
https://github.com/pressbooks/pressbooks/blob/5a20ea04f2e103ca381733a03c7b67f3da9b789c/inc/modules/export/class-export.php#L449

https://github.com/pressbooks/pressbooks/blob/d10da5d2679fbe57ad6da703e3ef68db8a3b5a0e/inc/class-licensing.php#L287

A fix for this would be to always print the copyright year regardless of which license is selected.

greatislander:
See #3249.


SteelWagstaff (approved):
:+1:

Issue #3412, opened 2023-07-18
Add support for download events to GA4 and remove support for Google UA


SteelWagstaff:
Google's Universal Analytics product has been replaced with GA4. We can now remove support for Google UA from Pressbooks and related plugins.

Also, a holdover from UA that we should address: as reported by eCampusOntario, networks using GA4 are unable to capture download events. They have identified the function to update:

> We have added the GA-4 code but we are unable to capture the downloads event from the pressbooks.
> 
> I did an inspection of the code and found out that we are using the old GA function. Please see the screenshot below.
> 
> ![image](https://github.com/pressbooks/private/assets/44182629/ebb86938-681c-4be8-abc6-e4ebe34cce96)
> 
> Can you please update the function to make it work with the GA-4?
> 
> I have an example how we can do it:
> 
> ![image](https://github.com/pressbooks/private/assets/44182629/77b2ab16-0185-4a59-b240-1b91d8b682e8)
> 
> Please let me know if you have any questions.

(Thomas adds: if there are any questions from the team for the person who wrote in here, I can relay them)

richard015ar:
I updated GoogleAnalytics class to remove multiple versions handling and I left only v4.
Changes in: https://github.com/pressbooks/pressbooks/compare/feat/download-action-ga4?expand=1
To do: update unit tests.

I removed ga3 references in pressbooks network analytics plugin. Changes in: https://github.com/pressbooks/pressbooks-network-analytics/compare/feat/remove-ga3-option?expand=1 

I changed and tested the new way to track download events. Tests were performed in dev.pressbooks.network GA account from my local env. Changes in: https://github.com/pressbooks/pressbooks-book/compare/fix/track-download-ga4?expand=1 

I'll add unit tests and submit PRs as soon as I can.

tw77:
This is working on dev networks. I'm able to track downloads of book exports in GA4's realtime overview. UA field is removed as expected from PB interface.

PR #3428, opened 2023-07-31
feat: remove google analytics v3

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/3412

This PR removes Google Analytics 3 implementation. It does not add any new features or bug fixes, but it implements breaking changes in pressbooks-book. This PR should be tested with https://github.com/pressbooks/pressbooks-network-analytics/pull/322 checked out.

### Test case
- Go to Network Level > Settings > Google Analytics and make sure only 1 Google Analytics field is available to fill/update.
- Add a Google ID
- Go to any public or admin page and ensure the Google Analytics script is added in the DOM as expected.
Ask the team for a Google ID to check if the events are appropriately sent to Google Analytics. It should not be necessary, since any part of the feature was changed/modified. Nevertheless, it was tested by me.

### DIFF
--- a/inc/class-googleanalytics.php
+++ b/inc/class-googleanalytics.php
@@ -27,8 +27,8 @@ class GoogleAnalytics {
 		$this->settings['input_label'] = __( 'Google Analytics ID', 'pressbooks' );
 		$this->settings['input_legend'] = __( 'The Google Analytics ID for your network, e.g &lsquo;G-A123B4C5DE6&rsquo;.', 'pressbooks' );}
 
-	public function getGoogleIDSiteOption( bool $for_book ): false|string {
-		return ! $for_book ?
+	public function getGoogleIDSiteOption( bool $for_book_context ): false|string {
+		return ! $for_book_context ?
 			get_site_option( $this->settings['option'] ) :
 			get_option( $this->settings['option'] );
 	}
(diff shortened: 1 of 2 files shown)


inc/class-googleanalytics.php
```diff
@@ -339,52 +277,17 @@ public function displayBookAnalyticsSettings(): void {
...
 
 	public function printScripts(): void {
-		$network_google_v3_code = self::getGoogleIDSiteOption( 3, false );
-		if ( ! empty( $network_google_v3_code ) ) {
-			$book_google_v3_code = get_site_option( self::$is_allowed_option ) ?
-				$this->getGoogleIDSiteOption( 3, true ) : '';
+		$network_google_code = $this->getGoogleIDSiteOption( false );
+		if ( ! empty( $network_google_code ) ) {
+			$book_google_code = get_site_option( self::$is_allowed_option ) ?
+				$this->getGoogleIDSiteOption( true ) : '';
```
fdalcin: Could we use named arguments here? I had to check the method to understand what `true|false` means in the context.

fdalcin (approved):
LGTM, added a comment about named arguments though. Let me know if you think it's valid, I can merge it otherwise.

inc/class-googleanalytics.php
richard015ar: it makes sense to me, it is not so intuitive. Thanks, Dalcin, I'll add it.

PR #3444, opened 2023-08-16
fix: change name of Mount Royal University

SteelWagstaff:
Mount Royal College has been Mount Royal University since 2009. See https://pressbooks.community/t/institutions-list-how-to-add-to-it/2174/9 for user request.

### DIFF
--- a/symbionts/institutions/institutions.json
+++ b/symbionts/institutions/institutions.json
@@ -329,7 +329,7 @@
           },
           {
             "code": "CA-AB-010",
-            "name": "Mount Royal College",
+            "name": "Mount Royal University",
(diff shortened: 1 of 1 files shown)


symbionts/institutions/institutions.json
```diff
@@ -329,7 +329,7 @@
           },
           {
             "code": "CA-AB-010",
-            "name": "Mount Royal College",
+            "name": "Mount Royal University",
```
cmurtagh, line 332: Oof, that's a big change. We should probably do regression tests on that.

richard015ar (approved):
LGTM

Issue #3449, opened 2023-05-17
Entries for pb_authors, pb_editors, etc. are not being cleaned up as expected in wp_blogmeta table


SteelWagstaff:
Authors are not displayed in the correct order in the catalog for open source users (i.e. users who are running Pressbooks and Pressbooks Network Catalog but not our private Pressbooks Network Analytics plugin & its data collector cron job).

First, it appears that we are not properly updating `pb_authors` in the `wp_blogmeta` table after book info is updated. For example, when a book's author is deleted/removed from book info, the corresponding record is not deleted from wp_blogmeta. Second, the catalog author display order is not the same as used in book info, on the book homepage or in the catalog when network analytics and its cron are working. I don't know why.

See https://github.com/pressbooks/pressbooks-network-catalog/issues/173 and this short video I made with Ed Beck, the open source user who reported the issue:

https://github.com/pressbooks/pressbooks-network-catalog/assets/13485451/ce993b87-aa02-4764-8967-f1f36eb83d40

DoD:
- [ ] Only the listed authors and editors in book info are displayed on the catalog listing page (when book is cloned or created new)
- [ ] wp_blogmeta is updated correctly when book info is updated (especially when users are _removed_ from pb_authors or pb_editors in book info)

We believe this is connected to https://github.com/pressbooks/pressbooks-network-catalog/issues/236 and https://github.com/pressbooks/pressbooks-network-catalog/issues/230

MichSpieler:
Behaviour that could be connected to this was also reported in this ticket (for internal reference) [#16912](https://pressbooks.zendesk.com/agent/tickets/16912)

MichSpieler:
Behaviour that could be connected to this was also reported in this ticket (for internal reference) [#17424](https://pressbooks.zendesk.com/agent/tickets/17424)

MichSpieler:
Behaviour that could be connected to this was also reported in this ticket (for internal reference) [#17493](https://pressbooks.zendesk.com/agent/tickets/17493)

MichSpieler:
Behaviour that could be connected to this was also reported in this ticket (for internal reference) [#17491](https://pressbooks.zendesk.com/agent/tickets/17491)

SteelWagstaff:
[...]

PR #3456, opened 2023-08-28
fix: contributors taxonomy on collectors

arzola:
See #3449 

This is a very obscure issue I don't know how we didn't realize the getAll method on contributors was not retuning the required information due the contributors taxonomy not being registered on noon book contexts.

The solution enable us to cleanup entirely the contributors associated to the book and then read them again to register the latest data in the collector.

How to test:

1. Save a few contributors, try to delete one of them and then try to add a new one.
2. Contributors should look OK in the catalog after saving metadata because is refreshed at this point
3. Run the collector manually and the records should match the stored configuration on book info, no orphans pb_authors should remain on the `wp_*_postmeta` tables

Notes:

Be aware the `pb_network_analytics_booklist_sync` action has a transient to cache the execution just in case

### DIFF
--- a/inc/class-taxonomy.php
+++ b/inc/class-taxonomy.php
@@ -51,9 +51,8 @@ class Taxonomy {
 	/**
 	 * @param Taxonomy $obj
 	 */
-	static public function hooks( Taxonomy $obj ) {
+	public static function hooks( Taxonomy $obj ) {
 		if ( Book::isBook() ) {
-			add_action( 'init', [ $obj, 'registerTaxonomies' ] );
 			add_action( 'init', [ $obj, 'maybeUpgrade' ], 1000 );
 			add_action( 'user_register', [ $obj->contributors, 'addBlogUser' ] );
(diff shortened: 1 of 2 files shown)


inc/class-taxonomy.php
```diff
@@ -69,6 +68,7 @@ static public function hooks( Taxonomy $obj ) {
 			add_filter( 'license_row_actions', [ $obj, 'removeTaxonomyViewLinks' ], 10, 2 );
 			add_filter( 'contributor_row_actions', [ $obj, 'removeTaxonomyViewLinks' ], 10, 2 );
 		}
+		add_action( 'init', [ $obj, 'registerTaxonomies' ] );
```
arzola, line 71: At least contributors taxonomy should be registered also on NON book context in order to functions like get_term_by functions properly because those functions do a taxonomy lookup before trying to query the terms

fdalcin (approved):
LGTM, tested it locally and authors are now updated correctly. Thank you!

PR #3497, opened 2023-10-24
fix: translation domain namespace

arzola:
The integrations menu should work for any kind of "integrations" it should not be translated on the lti-provider namespace


fdalcin (approved):
LGTM, thank you @arzola

PR #3512, opened 2023-11-08
chore: bump wordpress dependency

fdalcin:


fdalcin:
Close because it was a test PR

PR #3541, opened 2023-12-29
fix: use ARIA dialog pattern for glossary definitions (resolves #3507)

### LINKED ISSUE #3507: Glossary definition modal is not fully accessible
Pressbooks processes glossary shortcodes by printing a button. See the ARIA pattern for button: https://www.w3.org/WAI/ARIA/apg/patterns/button/. When clicked, this button causes a div containing the definition to become visible on the screen. When the user presses escape or clicks outside of the definition modal, the definition disappears. The patterns resembles the dialog (modal) pattern described here: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/, but with a few problems. 

The button pattern suggests:
> Following button activation, focus is set depending on the type of action the button performs. For example:
If activating the button opens a dialog, the focus moves inside the dialog. 

In our case, the user's focus does not shift to the glossary definition, which poses problems as users of assistive technology generally cannot access or hear the term's definition. 

> The button has an accessible label. By default, the accessible name is computed from any text content inside the button element. However, it can also be provided with [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) or [aria-label](https://w3c.github.io/aria/#aria-label).

> If a description of the button's function is present, the button element has [aria-describedby](https://w3c.github.io/aria/#aria-describedby) set to the ID of the element containing the description.

Our button does not have a description of the function, but does use aria-descriedby. The aria-describedby points to the ID of the glossary__tooltip which contains the definition. It might be better to have an accessible description that describes the button's function (`display term definition` or similar)?

We probably also want to provide a clear indicator or 'close' button inside the tooltip definition div which allows the user to exit modal/popup and return to the prior page navigation with their focus on the button they initially launched. 

We may want to add aria-haspopup="dialog" to the button, to make it clear that clicking the button will open a popup which includes dialog? Not sure if this is important or will make a difference. 

This description of accessible popups: https://theadminbar.com/accessibility-weekly/how-to-make-accessible-popups/ and accompanying CodePen: https://codepen.io/amberhinds/pen/gOBpBBo provide an accessible pattern that could be adapted for our glossary terms.

## Notes
The user who reported this concern to us provided this video: https://drive.google.com/file/d/1Zl8zSoQ0dlHyCVhCqYjZpsigHudG_fnz/view and shared some suggestions:

>  Please note, these hover option “definitions” are not links they are buttons.  They look like links on the page, but they are programmed and announced as buttons. I just want to add that this distinction of links versus buttons matters because it affects user expectations. If a screen reader announces an element as a “link” or “button,” users have expectations about what will happen when they click that element. If something else happens, this can be disorienting.

> Also, from the screen reader perspective the pop window appears to never receive focus once it is opened up by the user engaging with the link text coded as a button with their keyboard. The easy thing to try first is to code them as links and see how that works and have somebody test it with a screen reader. Yet, they need to code those links so that focus goes into the popup window content and that a screen reader user can easily close out those windows and have focus continues to read exactly where they left off.
 
> If that doesn’t work, then maybe they need to be linked directly to the glossary page where that definition resides.  The key is to make sure that a screen reader user can return to the exact same spot where they left off reading when they opened the glossary. And when I say first trying to change the buttons to links so that you have the same premise of a pop-up window with that definition. And it still stands that if that is something that will work, a screen reader user is able to easily close out that pop-up window and continue reading from where they left off so that they don’t lose any understanding of the content that is being read out to them.

greatislander:
This PR changes the display of glossary definitions to use a link with `aria-haspopup="dialog"` and displays the definition in a dialog. The definition dialog can be dismissed by clicking outside it, pressing the <kbd>ESC</kbd> key, or clicking/tapping the close button in the definition dialog.

Fix for #3507


fdalcin:
Thank you for your contributions, @greatislander.
I'm currently reviewing the PR and am getting some unexpected behaviour when trying to open the popup.

The overlay and the definition are created and displayed, however the definition's position seems to be off of the screen.

https://github.com/pressbooks/pressbooks/assets/1761690/570e0f74-12e1-4392-924f-449169a612e4

greatislander:
@fdalcin I believe my latest commit has resolved the issue you encountered. Please let me know if it's working as expected for you now.

fdalcin (approved):
Thank you @greatislander. Tested it locally and now it works as expected with a single or multiple definitions.

Issue #3508, opened 2023-10-31
Private content can't be viewed (by users other than Editors and Administrators) unless Show in Exports is checked


tw77:
### Description

When content is set to private and 'show in exports' is not selected, authenticated users are unable to view the content, even if their permissions should allow them to see the content.

For example, in this globally public book, they have selected the setting that should allow book users added with the role of Subscriber/Collaborator/Author to view pages that they have set to private.
<img width="500" alt="Screenshot 2023-10-31 at 7 52 23 AM" src="https://github.com/pressbooks/pressbooks/assets/44182629/da9e348e-d24a-4beb-842b-4e6b396c6faa">

There seems to be a bug here: users with these roles are able to view private front/back matter and chapters in the table of contents when 'Show in Exports' is selected, but they aren't able to load this private content.

### Steps to Reproduce

1. Log in as a test user added to the affected book with the role of Subscriber, Collaborator, or Author.
2. Navigate to the TOC.
3. Open one of the private pages in the TOC with a lock symbol next to it, e.g.: <img width="487" alt="Screenshot 2023-10-31 at 7 55 10 AM" src="https://github.com/pressbooks/pressbooks/assets/44182629/f790c08f-e0b8-46d6-8b1b-01b158ad8dff">

**Expected behavior:** The private chapter or front/back matter loads.

**Actual behavior:** <img width="801" alt="Screenshot 2023-10-31 at 7 55 27 AM" src="https://github.com/pressbooks/pressbooks/assets/44182629/0fdccab2-1611-4fcc-9116-f399642edf58">

tw77:
It turns out this is reproducible anywhere: when a page is not shown in exports and not shown in web, it does not load for users with these roles.

tw77:
Tested on https://dev.pressbooks.network/ and fix is working as expected.

PR #3550, opened 2024-01-11
fix: permissive content display

arzola:
This PR addresses #3508

I added a little extra hook to allow logged users to view the content as expected because when a post is set as **not show in web** is being set as a draft and drafts automatically are excluded from the wp loop.

### DIFF
--- a/hooks.php
+++ b/hooks.php
@@ -4,6 +4,7 @@
  * @license GPLv3 (or any later version)
  */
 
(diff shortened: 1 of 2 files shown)


SteelWagstaff:
@arzola looks like linter is nagging us. Maybe run `composer fix` and commit changes?

fdalcin (approved):
LGTM, thank you @arzola.

PR #3554, opened 2024-01-16
feat: handle new lti plugin

arzola:
This PR adds temporary support for the new PB LTI plugin.

### How to test

1. Enable any of the two LTI plugins and check if the integrations menu contains platforms and settings on network admin, main site and book admin if enabled.

A refactor is suggested to avoid tons of conditionals to register menus on the integrations menu.

### DIFF
--- a/inc/admin/menus/class-sidebar.php
+++ b/inc/admin/menus/class-sidebar.php
@@ -2,6 +2,7 @@
 
 namespace Pressbooks\Admin\Menus;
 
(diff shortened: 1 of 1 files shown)


fdalcin (approved):
@arzola thank you, LGTM. Tested locally and seems to work as expected.

@tw77 to test this you could check dev.pressbooks.network and ltidev.pressbooks.network. They should have an integrations menu with a couple of options inside. You could try deactivating some plugins to see if they are reflected as expected.

PR #3569, opened 2024-01-29
feat: setup eloquent ORM

arzola:
We've been using Eloquent in our new plugins and this is the first step to align and centralice Eloquent's setup, our next step would be research and create a POC using https://mantle.alley.com/

### how to test

1. Try to create a model in any plugin using Laravel's Base model and Fetch the data as needed.

```php
<?php
namespace Plugin\Models;

class MyModel extends Model 
{
protected $table = 'pressbooks_lti_users';
}
```
2. Retrieve your data using Fluent

```php
$data = MyModel::query()->first();
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,8 @@
 		"illuminate/http": "^8.0",
 		"illuminate/support": "^8.0",
 		"illuminate/view": "^8.0",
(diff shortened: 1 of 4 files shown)


fdalcin (approved):
LGTM, thank you @arzola

PR #3573, opened 2024-02-01
chore: add Dutch universities

tw77:
As requested by University of Groningen network manager.

### DIFF
--- a/symbionts/institutions/institutions.json
+++ b/symbionts/institutions/institutions.json
@@ -1443,40 +1443,40 @@
         "name": "Netherlands",
         "institutions": [
           {
-            "code": "EU-NL-001",
+            "code": "EU-NLD-001",
             "name": "Delft University of Technology",
             "url": "https:\/\/www.tudelft.nl/\/"
           },
           {
-            "code": "EU-NL-002",
+            "code": "EU-NLD-002",
             "name": "Leiden University",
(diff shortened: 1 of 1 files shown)


SteelWagstaff (approved):
LGTM!

SteelWagstaff:
@tw77 Tests are failing, but I don't think it's related to this PR. We need to investigate before merging.

SteelWagstaff:
@tw77 There were a couple of small errors with the original PR that were causing tests to fail. See fixes here: https://github.com/pressbooks/pressbooks/pull/3573/commits/ce978373ddb84ac826a8fc21cf0262b91a875ac8. Basically we need to use three letter ISO code for countries -- NL was already in use by Newfoundland in Canada. There was also an extra comma at the end of the new institution. I also went ahead and added the remaining members of the https://en.wikipedia.org/wiki/Universiteiten_van_Nederland in alphabetical order so we won't have to do that later.

Another small note -- in the future the branch should be have a prefix and the PR title should as well, indicating whether the change was a feature, a fix, a chore, etc. We want to use Conventional Commit standards for this: https://www.conventionalcommits.org/en/v1.0.0/. Not sure whether this would be considered a `chore` or a `fix`. In this case, I went with chore. That being the case, the branch should have been titled `chore/add-dutch-universities`

PR #3587, opened 2024-02-13
feat: append filters to handle groups

arzola:


arzola:
This wouldn't not be required, we will build another dashboard and probably it would be handled differently

PR #3594, opened 2024-02-17
fix: resolve issues with glossary definition modal (resolves #3572)

### LINKED ISSUE #3572: Unexpected placement for glossary term pop-up window and unexpected behaviour upon closing of the pop-up window
### Description

Several clients have commented on an unexpected behaviour with the glossary term pop-up window. They expect the pop-up modal to appear over the glossary term, but instead the modal appears at the top of the page (the reading position in the background is lost). Upon closing the modal, the user is taken back to the placement of the first instance of the glossary term.

### Steps to Reproduce for the location of the pop-up window

1. Create a glossary term in a longer chapter.
2. Click on the glossary term.
3. Close the pop-up window.

**Expected behavior:** the window view of the reading interface (in the background) does not change throughout the whole process.

**Actual behavior:** the window view jumps to the top of the page when the modal is opened and back to the glossary term when the modal is closed.

A second issue is that if the same term is included multiple times in a given chapter, focus is always returned to the first instance of the term upon closing the modal, even if the user launched the definition from say the second or third occurrence of that term.

### Steps to Reproduce for the location after closing the pop-up window

1. Create a glossary term in a longer chapter and reuse it further down in the text.
2. Click on the glossary term's second occurrence.
3. Close the pop-up window.

**Expected behavior:** after closing the pop-up window, the user returns to the location of the second occurrence of the glossary term.

**Actual behavior:** the user returns to the location of the first occurrence of the glossary term.

https://github.com/pressbooks/pressbooks/assets/13485451/b3d313db-987d-4d09-aaf0-3ecd1a518e15

Finally, if the glossary term is formatted and the shortcode around it is not (i.e. bolded, a `span` element is added between the two pieces of shortcode), the glossary term pop-up will not appear. [Steel's note: I think this may be a prexisting issue and is unrelated to recent changes?]

### Steps to Reproduce for the formatted glossary term

1. Create a glossary term
2. Make the glossary term bold (but not the short code)
3. Click on the glossary term.

**Expected behavior:** a pop-up window appears.

**Actual behavior:** nothing happens.

greatislander:
Resolves #3572. Also, I removed the old `glossary-tooltip.js` files which I forgot to do in the earlier PR.


SteelWagstaff (approved):
Works for me locally. Will merge for further testing in dev environments. thanks @greatislander!

PR #3609, opened 2024-02-26
Add filters for pb_sanitize_webbook_content config and spec

matthias-noetzli:
The `sanitize_webbook_content` filter introduced in [#1462](https://github.com/pressbooks/pressbooks/pull/1462) filters webbook HTML using HTMLawed. This PR adds filters for the config and spec parameters passed to HTMLawed by `sanitize_webbook_content`, enabling users to customize this behavior according to their needs.

We encountered a specific issue where a Vimeo embed resulted in an iframe with the `allow` attribute being stripped out, which prevented the fullscreen option from being displayed. With this PR, this issue can be addressed as follows:
```
add_filter('pb_sanitize_webbook_content_spec',  'demo_pb_sanitize_webbook_content_spec'));

function demo_pb_sanitize_webbook_content_spec($spec)
    {
        $spec .= 'iframe=allow;';
        return $spec;
    }
```

### DIFF
--- a/inc/sanitize/namespace.php
+++ b/inc/sanitize/namespace.php
@@ -757,11 +757,11 @@ function sanitize_webbook_content( $content ) {
 	// Remove deprecated table borders
 	$spec = '';
 	$spec .= 'table=-border;';
-	
-	$spec = apply_filters('pb_sanitize_webbook_content_spec', $spec);
-	$config = apply_filters('pb_sanitize_webbook_content_config', []);
-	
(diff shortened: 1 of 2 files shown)


fdalcin (approved):
Thank you for your contribution and sorry for the delay reviewing it.
I took the liberty of adding a friendly commit with a couple of tests.

arzola:
cc @matthias-noetzli

PR #3613, opened 2024-02-27
chore: update autobuild asset workflow

SteelWagstaff:
This PR simply commits the built assets to the dependabot PR branch instead of opening and automerging a PR. Fix for https://github.com/pressbooks/composer-autoupdate-bedrock/issues/62


arzola (approved):
LGTM

PR #3617, opened 2024-02-29
chore: use reusable workflow to generate .mo files

SteelWagstaff:
Partial fix for https://github.com/pressbooks/private/issues/1340


arzola (approved):
LGTM

PR #3621, opened 2024-02-29
chore: Add Boğaziçi University (Bosphorus University) to institutions list

cmurtagh:
As requested in the community forum

### DIFF
--- a/symbionts/institutions/institutions.json
+++ b/symbionts/institutions/institutions.json
@@ -3221,6 +3221,21 @@
     ]
   },
   {
+    "country": "Turkey",
+    "regions": [
+      {
+        "name": "Istanbul",
+        "institutions": [
+          {
+            "code": "TR-IST-001",
+            "name": "Bosphorus University",
+            "url": "https:\/\/bogazici.edu.tr\/"
+          }
+        ]
+      }
+    ]
+  },
+  {
     "country": "United States of America",
     "regions": [
       {
--- a/symbionts/institutions/institutions.json
+++ b/symbionts/institutions/institutions.json
@@ -2391,7 +2391,6 @@
       {
         "name": "Netherlands",
         "institutions": [
-
           {
             "code": "EU-NLD-001",
             "name": "Delft University of Technology",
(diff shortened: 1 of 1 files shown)


SteelWagstaff (changes requested):
Thanks for submitting this, Christopher. The JSON structure is a bit wacky in this file. Although the titles are 'country' and 'region' -- for most countries outside the US and Canada these are actually used for Continent and Country. So in this case we'd want to set Europe as the Country and Turkey as the Region. Their institution code would use EU for the Country and the three letter ISO code for Turkey (TUR) in the region place. 

In hindsight, a better naming convention would probably have been 'region' and 'subregion', but these labels aren't displayed to users anywhere in the front-end, so refactoring it hasn't been a priority.

SteelWagstaff:
I pushed a friendly commit that updates the values. The other question I have is whether we should use the university's official name 'Boğaziçi University' rather than Bosphorus University. I think probably yes?

cmurtagh:
I'm all for having the official name, yes. 

Question: Most of Turkey isn't in Europe, but this institution is. If we get another request from an institution in Ankara, would we put that in Asia/Turkey?

SteelWagstaff:
Good question. We have a few options here. We can indicate that Turkey is in Europe, indicate that Turkey is in Asia, or simply display Turkey (without a parent region).
Here's what the user would see with the PR as it stands now:
Screenshot from 2024-02-29 08-04-21
Here's what would be displayed if we put Turkey in Asia or without a geographic subregion:
Screenshot from 2024-02-29 08-04-56

What do you think is best here? This is a complex geographical question, and a pretty trivial choice for us :joy:

cmurtagh:
Can we have both? Turkish institutions in Europe and also in Asia?

cmurtagh:
My vote would be to keep the PR as it is now (Europe/Turkey) and when we add another institution that isn't from Western Istanbul, we'll figure it out.

SteelWagstaff:
We've decided to use the official name and to go with Turkey as a country without regions for now. Will display like this:
Screenshot from 2024-02-29 08-14-56

SteelWagstaff (approved):

PR #3635, opened 2024-03-13
fix: menu order workaround

arzola:
This would allow us to position the institutions menu below users.

Definitely this is a workaround, meanwhile we refactor the class-sidebar


richard015ar (approved):
LGTM

PR #3646, opened 2024-03-20
fix: check user exists

fdalcin:
This PR aims to fix a warning noted here https://github.com/pressbooks/private/issues/1385. 

The reason we are seeing the warning is because the user might have been deleted. If that's the case WP returns `false` in the `get_user_by` method and we are attempting to get the `ID` of a `boolean` value.


richard015ar (approved):
LGTM, thank you @fdalcin

Issue #3638, opened 2024-03-14
Fix menu order for Pages in root site context


SteelWagstaff:
When I enter the root site context, the pages menu item has moved from its expected location/order:
(on dev network) -- jumped to bottom instead of being displayed between appearance & plugins
![Screenshot from 2024-03-14 10-46-42](https://github.com/pressbooks/pressbooks-multi-institution/assets/13485451/5f9b19ec-6722-48d3-bfa0-06d670b06262)
(on integrations) -- stays in between appearance and plugins
![Screenshot from 2024-03-14 10-46-29](https://github.com/pressbooks/pressbooks-multi-institution/assets/13485451/34c6e2f3-a9b4-4128-81bd-98544b6879c1)

To reproduce: visit https://dev.pressbooks.network/wp/wp-admin/edit.php?post_type=page and click the pages link from the admin dashboard. Notice that the location of the pages menu item moves to the bottom of the list. This shouldn't happen.

Perform the same actions at https://integrations.pressbooks.network/wp/wp-admin/edit.php?post_type=page to see expected behaviour.

See https://us02web.zoom.us/rec/share/Ci9NJWHwtMeWfP75D49m9gId1PDbsW1-I-T9LYuuDdk4UgdMn_K1be3H0WgJESYt.GhOUM_Y4SvJ7xVRN for a brief demonstration.

_Originally posted by @SteelWagstaff in https://github.com/pressbooks/pressbooks-multi-institution/issues/96#issuecomment-1998009159_

richard015ar:
This PR: https://github.com/pressbooks/pressbooks-multi-institution/pull/124

Is ready and working on Ricardodev, but I will wait for the [Code Improvements](https://github.com/pressbooks/pressbooks-multi-institution/issues/76) to include those changes there.

I extended my PR to add more changes, described in https://github.com/pressbooks/pressbooks-multi-institution/pull/124.

tw77:
Works as expected on dev.pressbooks.network.

PR #3654, opened 2024-03-25
fix: revert menu order

richard015ar:
This PR reverts some of the position items defined in: https://github.com/pressbooks/pressbooks/pull/3635 and changes hooks priorities in order to preserve the previous order and avoid: https://github.com/pressbooks/pressbooks/issues/3638

### DIFF
--- a/inc/admin/menus/class-sidebar.php
+++ b/inc/admin/menus/class-sidebar.php
@@ -61,7 +61,7 @@ class SideBar {
 
 		if ( ! is_restricted() ) {
 			add_filter( 'custom_menu_order', '__return_true' );
-			add_filter( 'menu_order', [ $this, 'reorderSuperAdminMenu' ], 999 );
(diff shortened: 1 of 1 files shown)


inc/admin/menus/class-sidebar.php
```diff
@@ -61,7 +61,7 @@ public function hooks(): void {
 
 		if ( ! is_restricted() ) {
 			add_filter( 'custom_menu_order', '__return_true' );
-			add_filter( 'menu_order', [ $this, 'reorderSuperAdminMenu' ], 999 );
+			add_filter( 'menu_order', [ $this, 'reorderSuperAdminMenu' ], 998 );
```
richard015ar, line 64: decreasing priority so we can re-order if we need it through another hook.

arzola (approved):
LGTM works as expected

arzola (approved):
LGTM

PR #3659, opened 2024-03-31
fix: resolve contrast issue with active menu item

greatislander:
Resolves [WCAG SC 1.4.3 Contrast (minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum) issue identified in https://github.com/pressbooks/pressbooks-multi-institution/issues/129#issuecomment-2027683916.


SteelWagstaff (approved):
LGTM! Thanks Ned.

Issue #3665, opened 2024-04-02
Update to v4 of CodeCov, Upload & Download Artifact Actions


SteelWagstaff:
In order to update the code coverage for this repo, we run PCOV, save the code coverage output as an XML file using an upload-artifact action: https://github.com/pressbooks/pressbooks/blob/0aa3c8fd53378e21c31dba2fad532981d21ee56a/.github/workflows/tests.yml#L83

Later in the pipeline, we download that coverage artifact: https://github.com/pressbooks/pressbooks/blob/0aa3c8fd53378e21c31dba2fad532981d21ee56a/.github/workflows/tests.yml#L139

and upload it to Codecov using the codecov-action: https://github.com/pressbooks/pressbooks/blob/0aa3c8fd53378e21c31dba2fad532981d21ee56a/.github/workflows/tests.yml#L143

Each of these actions is current set to v3. A v4 release is available for all of them, which includes breaking changes. For the upload/download artifact actions, see https://github.com/actions/download-artifact/blob/main/README.md#breaking-changes and https://github.com/actions/upload-artifact/blob/main/README.md#breaking-changes

For the Codecov action, see https://docs.codecov.com/docs/adding-the-codecov-token#github-actions and https://github.com/codecov/codecov-action/releases/tag/v4.0.0

SteelWagstaff:
Works as expected after the change.

PR #3666, opened 2024-04-02
chore: bump action versions and add CODECOV_TOKEN secret

SteelWagstaff:
Addresses #3665. I followed these instructions: https://docs.codecov.com/docs/adding-the-codecov-token#github-actions to generate the required Codecov token and added it as a repository secret.


fdalcin (approved):
LGTM

PR #3667, opened 2024-04-03
fix: remove network manager collector

fdalcin:
This PR is part of pressbooks/bi-data-manager#468. Accompanies pressbooks/pressbooks-network-analytics#356

We're dropping the data collection for `pressbooks_network_managers_ids` since it's not being kept in sync with `pressbooks_network_managers`. 

This might be a breaking change if others use the method or if they are reading `pressbooks_network_managers_ids` meta value from the database.

### DIFF
--- a/inc/datacollector/class-user.php
+++ b/inc/datacollector/class-user.php
@@ -6,8 +6,6 @@
 
 namespace Pressbooks\DataCollector;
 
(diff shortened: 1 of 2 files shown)


richard015ar (approved):
LGTM

PR #3668, opened 2024-04-04
fix: remove patterns sub-menu items for book admin context

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1395.

This PR removes the `Patterns` sub-menu item under `Appearance` added in WP 6.5.

### Testing case
- Go to any book admin context and make sure Appearance does not have any sub-menu items called `Patterns`.
- If you manually enter the `<BOOK_URL>/wp-admin/edit.php?post_type=wp_block` URL, you should see the `Sorry, you are not allowed to access this page.` page.

### DIFF
--- a/inc/admin/menus/class-sidebar.php
+++ b/inc/admin/menus/class-sidebar.php
@@ -54,7 +54,8 @@ class SideBar {
 
 	public function hooks(): void {
 		if ( ! is_main_site() ) {
-			add_action( 'admin_menu', [ $this, 'removePatternsSubMenuItem' ], 999 );
+			add_action( 'admin_menu', [ $this, 'removePatternsSubMenuItem' ] );
+			add_action( 'admin_init', [ $this, 'restrictPatternsPageAccess' ] );
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


inc/admin/menus/class-sidebar.php
```diff
@@ -67,6 +73,20 @@ public function hooks(): void {
...
+	}
+
+	public function restrictPatternsPageAccess(): void {
+		global $pagenow;
+
+		if ( $pagenow !== 'edit.php' || ! isset( $_GET['post_type'] ) || $_GET['post_type'] !== 'wp_block' ) {
+			return;
+		}
+
+		wp_die( __( 'Sorry, you are not allowed to access this page.', 'pressbooks' ), 403 ); // phpcs:ignore Pressbooks.Security.EscapeOutput.OutputNotEscaped
```
richard015ar, line 87: Lint issue:
```bash
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 87 | ERROR | All output should be run through an escaping function
    |       | (see the Security sections in the WordPress Developer
    |       | Handbooks), found '__'.
    |       | (Pressbooks.Security.EscapeOutput.OutputNotEscaped)
----------------------------------------------------------------------
```
Here we have 2 options:
- ignore phpcs
- Use an escaping data method, see: https://developer.wordpress.org/apis/security/escaping/. I don't think any of them would make sense in this case

fdalcin: I don't see the point to escape this string since it does not have variables nor html, so ignoring phpcs makes more sense to me.

fdalcin (approved):
LGTM, testing it locally and I don't see the Patterns menu when accessing the book context.

PR #3688, opened 2024-04-15
fix: Rename variable to trigger bedrock update

ho-man-chan:
Resolves: pressbooks/private#1405
This PR includes changes to rename a variable which is necessary to trigger a specific update in the bedrock configuration.


arzola (approved):
LGTM

PR #3727, opened 2024-06-19
chore: remove unused features

arzola:
This PR removes ODT and HTMLBook exports, feature is broken and is not used


fdalcin (approved):
LGTM

PR #3747, opened 2024-07-23
fix: user menu priority

fdalcin:
This PR aims to fix an issue with the User Avatar after upgrading to WP 6.6.1.

See: https://core.trac.wordpress.org/ticket/61615


arzola (approved):
LGTM

Issue #3763, opened 2024-08-15
Change default 'Private Content' setting to 'All logged in users including subscribers'


SteelWagstaff:
We have a setting that determines which user roles are able to view private content in a book. The default setting for new books is 'Only logged in editors and administrators'. This causes problems when private content is launched via LTI, since students are mapped as subscribers and cannot view private content, defeating the purpose of using the LTI integration!

We should set the default value for new books to 'All logged in users including subscribers'. 

Code for this setting: https://github.com/search?q=repo%3Apressbooks%2Fpressbooks%20permissive_private_content&type=code

See previous discussion of this in https://pressbooks.com/pressbooks/pressbooks-lti/issues/57: 

> Works in Canvas with https://ltidev.pressbooks.network/pageorder/?post_type=chapter&p=26&preview=true. As a student in Canvas (mapped to the role of subscriber in PB) I'm able to launch this hidden from web, hidden from exports 'draft' chapter -- provided that the book has the setting that enables subscribers to view private content:

<img width="509" alt="Screenshot 2024-01-29 at 7 28 16 AM" src="https://github.com/pressbooks/pressbooks-lti/assets/44182629/997bf016-ca8b-43a2-8175-0e8fe4aabe3b">

> But if the privacy setting _doesn't_ enable subscribers to view private content, limiting it to editors and administrators, then I can't launch the chapter as a student (subscriber) from the LMS, and I get the Oops! That content can't be found message instead. In a way, this is what I would expect; the book's privacy settings / user roles are respected and not overridden by the LTI launch.

> It may be worth noting though that the default privacy setting for a new book is to limit private content to editors and administrators, while the default user role mapping on a PB network is to map students from the LMS to the role of subscriber. This could be a potential stumbling block for instructors who assume that any content from a book can, regardless of its state, be shared with participating students in an LMS course just by being brought into that course.

_Originally posted by @tw77 in https://github.com/pressbooks/pressbooks-lti/issues/57#issuecomment-1914954380_

cmurtagh:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pressbooks-Product-Backlog-60ae7d807f6e370010b3d5be/issues/pressbooks/pressbooks/3763?planning-poker) with Zenhub @arzola @fdalcin @ho-man-chan @richard015ar @tw77

ho-man-chan:
Potential solution:
[...]

PR #3767, opened 2024-08-22
feat: add default value for permissive_private_content option

richard015ar:
Issue: https://github.com/pressbooks/pressbooks/issues/3763

This PR set the `permissive_private_content` option to `1` for new books. 
Currently, this option is not being stored or defined during book creation. In order to display posts, if it is not defined (default behaviour), it will consider this value as `false`.
With this change, we set the value to `1` during book creation.

### Testing case
- Create a new book
- Go to Settings > Sharing & Privacy
- Make sure `Private Content` is defined as `All logged in users including subscribers.` option


fdalcin (approved):
LGTM, works as expected locally

Issue #3191, opened 2023-02-15
Remove users from network manager array when super admin status is revoked


SteelWagstaff:
If a user has been added as a network manager, they exist as a restricted super admin. If they are removed from the 'network manager' list by simply taking away their super admin status, their user ID remains in the network manager array list until the next time a super admin is restricted/unrestricted as a network manager https://github.com/pressbooks/pressbooks/blob/997b82cb90d64d0db3f3deb6195b9df77a916e15/inc/admin/networkmanagers/namespace.php#L91

We never want to have IDs in the network manager restricted array for users who are no longer super admins. To ensure better database integrity, when a user has their super admin status removed (at the network level), we should check to see if they're in the network manager restricted array. If so, we should remove them from this array.

## Notes
See https://developer.wordpress.org/reference/functions/revoke_super_admin/#hooks and https://github.com/WordPress/WordPress/blob/0bccdce4428c394e698e793e7dec38172c3d7027/wp-includes/capabilities.php#L1143-L1167 for how this function is invoked in WordPress.

SteelWagstaff:
@tw77 is this still a problem, or has the underlying problem been fixed? If so, we can close this issue.

fdalcin:
This can be definitely close since there was a fix for this implemented here https://github.com/pressbooks/pressbooks/pull/3771

PR #3771, opened 2024-08-27
fix: remove user from pb_network_managers on super admin revoke or de…

arzola:
This PR address #3191 

The site option would remove the ID from the `pressbooks_network_managers` this would also solve the discrepancy in the bi-data-manager, probably a cleanup process would be required for existing networks.

This will also perform a cleanup of Orphan IDs on `pressbooks_network_managers` IDs that are no longer super_admins or deleted users that doesn't exist anymore.


fdalcin (approved):
LGTM, works as expected locally

PR #3773, opened 2024-08-29
fix: append hooks in noon admin context

arzola:
This will allow to run the cleanup function for network managers using WP CLI.

Running `wp super-admin remove USER` will cleanup pressbooks_network_managers site_meta also it will do the same on user deletion


fdalcin (approved):
LGTM

PR #3777, opened 2024-09-04
Fix: hide full plugin menu in books for network managers

SteelWagstaff:
Network managers can currently see all plugins on the network when they view the plugins page in individual books, which sometimes leads to confusion. See https://github.com/pressbooks/pressbooks-ecampus-ontario/issues/48 for example.

This change makes it so that they see only the 'approved' plugins that book admins can activate/deactivate.

Before:
Screenshot from 2024-09-04 14-26-07

After:
Screenshot from 2024-09-04 14-25-27

### DIFF
--- a/inc/admin/plugins/namespace.php
+++ b/inc/admin/plugins/namespace.php
@@ -11,7 +11,7 @@ namespace Pressbooks\Admin\Plugins;
 use function Pressbooks\add_notice;
 
 /**
- * Hide plugins that aren't prefixed with `pressbooks-` (only applies to books).
+ * Hide unapproved plugins for book admins & network managers (only applies to books).
  * To show all plugins to all users, place the following in a plugin that loads before Pressbooks:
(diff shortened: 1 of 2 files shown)


arzola (changes requested):
I think this is not enough to solve the issue

It will still display the Pressbooks Plugins because the foreach will append the approved plugins + pressbooks-* plugins (it will discard all the others but still will include pressbooks ones)

I think we should apply an extra condition in the plugins foreach

I tested it on my local with a network manager and it filters out all other plugins like classic editor, etc... and show pressbooks-* and allowed ones (maybe I'm testing it wrong)

```php
foreach ( $plugins as $slug => $value ) {
			if ( strpos( $slug, 'pressbooks-' ) !== false || in_array( explode( '/', $slug )[0], $slugs, true ) ) {
				$approved[ $slug ] = $value;
			}
		}
```

arzola (approved):
LGTM

PR #3779, opened 2024-09-05
chore: add missing import

arzola:
is_restricted is namespaced so is not in the same namespace, so an import is required


fdalcin (approved):
LGTM, thank you @arzola

PR #3795, opened 2024-09-26
fix: avoid applying shortcodes during mathjax content analysis

richard015ar:
This PR fixes the fact that footnotes are displayed twice in the web format of a chapter when the footnotes contain latex shortcode.

At the same time, if process latex formulas within the TablePress [table] shortcode when it contains latex formulas. Even the table may not contain formulas, the code considers it may have.

For testing:
- Add tablePress without any formulas in a chapter
- Add tablePress table with formulas in another chapter
- Add footnotes with and without latex formulas in different chapters

Make sure the formulas and tables are rendered as expected.


fdalcin (approved):
LGTM. Testing locally it works as expected and no more duplicate footnotes.

PR #3826, opened 2024-11-05
feat: update post_modified fields when menu_order is updated

arzola:
This will help us to track menu_order changes.

### Problem
`menu_order` gets updated when chapters are reordered in the organize page but we are not touching the timestamp of each chapter so we don't have a way to track menu_order changes.

### Solution 
Touch timestamp fields when menu_order changes

### How to test
Reorder, delete, create new posts and check timestamp fields to see if those gets updated

### DIFF
--- a/inc/admin/organize/namespace.php
+++ b/inc/admin/organize/namespace.php
@@ -122,6 +122,7 @@ function reorder() {
 		parse_str( $_POST['old_order'], $old_order );
 		$new_parent = (int) $_POST['new_parent'];
 		$old_parent = (int) $_POST['old_parent'];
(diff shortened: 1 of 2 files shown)


richard015ar (approved):
Editing the chapter such as metadata, title or content, already changes the `post_modified` and `post_modified_gmt` values as expected, likely by default through WP core. However, changing the order in the organize page (dragging and dropping chapter) does change the `menu_order`, but it does not modify the `post_modified` and `post_modified_gmt`  fields.

Using this branch, values are updated. As our server (dev and prod) are on UTC 0, both dates are equal, which is also expected.

This PR LGTM and works.

PR #3834, opened 2024-11-06
chore: update Lemoore College name 

tw77:
As requested by an OER librarian. The college’s name changed from West Hills College Lemoore to Lemoore College.


fdalcin (approved):
LGTM

Issue #3648, opened 2024-03-21
Issue where a user is listed in a book with the role of 'Administrator, Subscriber' and cannot view items in the Media Library


tw77:
### Description
We've had a few reports of a strange issue where a regular user with the role of Administrator in a book is unable to view the media items that are present in the book's Media Library. The user can access the Media Library page and they can even upload items there, but the items immediately disappear from their view.

Although no media items are shown for the user from the Media Library, the media items exist in the book. Other users (for example, network managers or super admins) can still view the items in the Media Library.

Each time this has happened, we've seen that the user encountering the bug is listed in the book user list with the role of 'Administrator, Subscriber'. It seems that, while being added to the book's user list, the Subscriber role has been automatically applied. (In terms of their actual permissions, this 'Administrator, Subscriber' user is an Administrator.)

The user's role can be reset to just 'Administrator', dropping the 'Subscriber', by removing the user from the book user list and then adding them back in.

Each time this has happened, removing the user from the book user list and adding them back in has solved the issue for them. Once the user is re-added as an Administrator, they can view items in the Media Library page again.

### Notes
The automatic, in this case unexpected provision of the Subscriber role may be related to a change described at https://github.com/pressbooks/pressbooks/issues/3628#issuecomment-1995296065. 

We don't know if a user being an 'Administrator, Subscriber' always means that this bug exists for them.

This issue has been reproduced once on a testing network, but not reliably, so finding steps to replicate will take some more investigation.

tw77:
This bug can be recreated:

1. Navigate to a book user list that includes a user with the role of Administrator. (This can be your own user.)
2. Click **Edit** under this user's username.
3. Edit some aspect of this user's profile. (For example, add a website in the Website field.)
4. Save changes.

At this point, the user becomes an 'Administrator, Subscriber' in the book with the attendant loss of any ability to view media in the Media Library.

richard015ar:
I think here's the issue we need to look at: https://github.com/pressbooks/pressbooks/blob/6089313b5df8e52954c943d57a2080bcbeab000d/inc/class-activation.php#L68

fdalcin:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pressbooks-Product-Backlog-60ae7d807f6e370010b3d5be/issues/pressbooks/pressbooks/3648?planning-poker) with Zenhub @arzola @ho-man-chan @richard015ar @tw77

tw77:
[...]

PR #3841, opened 2024-11-11
fix: prevent subscriber role when not needed

fdalcin:
Issue #3648 

This PR aims to fix an issue where `subscriber` role was added to a user while updating them. Users would have two roles in place causing some issues with permissions.

**How to test**

- Follow the steps from this comment: https://github.com/pressbooks/pressbooks/issues/3648#issuecomment-2430301503
- Users should be updated without having an additional role
- In addition, follow the steps from this PR https://github.com/pressbooks/pressbooks/pull/2269#issue-950986539 to make sure lowly users are still able to create/clone books

### DIFF
--- a/inc/class-activation.php
+++ b/inc/class-activation.php
@@ -438,11 +438,21 @@ class Activation {
 	}
 
 	public function addUserDefaultRole( $id ) {
-		if ( is_numeric( $id ) ) {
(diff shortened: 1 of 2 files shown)


richard015ar (approved):
LGTM and it works as expected

PR #3844, opened 2024-11-13
chore: update heroicons webfont, includes entire solid and outline

arzola:
This PR includes the entire heroicons iconset


fdalcin (approved):
LGTM

PR #3847, opened 2024-11-14
chore: remove duplicated prefix

arzola:
I hated the duplicated Heroicons prefix so I fixed it


fdalcin (approved):
LGTM, thanks for the refinement

Issue #3852, opened 2024-11-19
Preserve appearance of generic login page in WP 6.7 upgrade


tw77:
Upgrading to WP 6.7 on dev and staging changes the appearance of the generic login page from this:

![image](https://github.com/user-attachments/assets/5c8a443a-85b7-4ece-a736-1059e96d16af)

to this:

![image](https://github.com/user-attachments/assets/c760e2ef-1df5-474b-b3d2-ff0fca79b1b5)

We'll need to preserve the current appearance of the login page in the upcoming release (including continued custom logo support).

Relevant link: https://core.trac.wordpress.org/ticket/51786

PR #3853, opened 2024-11-19
fix: custom login logo wp 6.7

arzola:
Address #3852


richard015ar (approved):
LGTM. I'll wait for some discussion on this in our next bug scrub before merging it.

PR #3860, opened 2024-12-02
Add missing restore_current_blog call in getCoverThumbnail

matthias-noetzli:
The getCoverThumbnail method in Pressbooks\DataCollector\Book calls switch_to_blog( $book_id ); to get the attachment information but then fails to call restore_current_blog. In my case this lead to issues with other plugins trying to save metadata.
This pull request simply adds said restore_current_blog call to the function.


arzola:
We'll take a look, I think you are right thanks @matthias-noetzli

arzola (approved):
LGTM, tested and tests are passing

Definitely a good practice to do specially during switching contexts, we do that on other methods

PR #3862, opened 2024-12-03
chore: reuse objects to csv function

arzola:
This PR adds this function used by results to convert object arrays to csv without any library


fdalcin (approved):
LGTM

PR #3865, opened 2024-12-10
fix: fonts test cdn fallback

arzola:
jsdeliver files are randomly returning Package size exceeded the configured limit of 50 MB on subsequent requests, but each file download works after doing a hard refresh


richard015ar (approved):
LGTM

PR #3867, opened 2024-12-14
fix: add header_bg to customizer colors array

SteelWagstaff:
Adds header background to the array of color variables that can be overridden with the theme customizer. Partial fix for https://github.com/pressbooks/pressbooks-aldine/issues/198


arzola (approved):
LGTM

PR #3872, opened 2024-12-19
fix: drag and drop menu order

fdalcin:
Issue: pressbooks/private#1560

This pull request includes changes to the `reorder` function in the `assets/src/scripts/organize.js` file to improve code readability and maintainability by simplifying variable assignments and regex replacements.

Improvements to code readability and maintainability:

* Simplified variable assignments by extracting `id`, `old_parent`, and `new_parent` from `item` and `pb.organize` respectively.
* Replaced inline calls to `sortable('serialize')` with pre-computed `old_order` and `new_order` variables.
* Used a regex to consistently replace part identifiers in `old_order` and `new_order`.


assets/src/scripts/organize.js
```diff
@@ -228,16 +228,27 @@ function updateControls( table ) {
...
+	const id = item.id;
+	const old_parent = pb.organize.oldParent.replace( /^part_([0-9]+)$/i, '$1' );
+	const new_parent = pb.organize.newParent.replace( /^part_([0-9]+)$/i, '$1' );
+	let old_order = $( `#${ pb.organize.oldParent }` ).sortable( 'serialize' );
+	let new_order = $( `#${ pb.organize.newParent }` ).sortable( 'serialize' );
+
+	const regex = /part_([0-9]+)/g;
+
+	old_order = old_order.replaceAll( regex, 'part' );
+	new_order = new_order.replaceAll( regex, 'part' );
```
fdalcin, line 240: The request previously sent something like `part_5[]=31&part_5[]=27`. However, when moving a chapter between parts this would become inconsistent by sending something like `part_5[]=31&part_4[]=27` which would assign the incorrect menu order to the chapter.

Since we don't make use of the key `part_([0-9]+)` in the background, we replace them with `part` so the chapters are assigned the correct menu order by the process.

richard015ar (approved):
changes LGTM

PR #3875, opened 2024-12-19
fix: set media attributions to display by default

SteelWagstaff:
Sets media attribution global setting to 'on' by default. Fix for https://github.com/pressbooks/ideas/issues/220


arzola (approved):
LGTM

arzola:
cc @tw77

PR #3893, opened 2025-01-07
fix: social media default values when empty

arzola:
This is the easiest solution for this problem.

The Options class isn't ready to work with checkbox arrays, for now our best bet is to check for stored values before displaying it because the way we built the checkboxes is not working well with default values behaviour.

### Problem
If you uncheck all the social media options, that will be saved on the database successfully but the UI will display the items checked (default values)

image

We can improve later this if required, hooking pb_theme_options_web_defaults or reworking the form and refactor the builtin renderCheckbox method with checkboxes group support


fdalcin (approved):
LGTM

PR #3908, opened 2025-01-31
wip: PB PSR4 new era (build in public)

arzola:
Pressbooks, built several years ago, adheres to outdated standards that don’t align with modern [PHP frameworks](https://www.php-fig.org/). However, we’ve developed several tools that enhance developer productivity and provide new tools for those who want to build on top of Pressbooks.

## Discovery phase and refactor TODO

> The WordPress ecosystem has evolved, introducing new methods and hooks to manage plugin dependencies. Manual dependency checks are no longer necessary.

- [x] Improve Pressbooks Plugin Registration with a clear separation of concerns.
- [x] Use Laravel Pint zero configuration setup for coding standards (Pint defaults are sane defaults).
- [x] Remove the custom HumanMade autoloader and rely on Composer PSR4.
- [x] Improve Service Container and create ServiceProvider Interface with deferred boot.
- [ ] Carefully clean up and set up helper functions; maybe separate them into different helper files.
- [ ] Eliminate all namespace.php files and create Skeleton classes for Services, Entities, Helpers, etc.
- [x] Use the Laravel Container to register services that could/should be available globally, like (Blade, DB Facade, Menu Manager?).
- [ ] Map and reorganize tests by scope and include all of them in the new format.
- [ ] Create/improve tests regarding all plugin registration stuff.
- [ ] Double-check the need for suppressing errors with `@`.
- [ ] Think about useful hooks on service providers or maybe some observables.
- [ ] Map Plugins Pressbooks's dependencies like functions, classes, and files used by them to rewrite those.
- [ ] Maybe remove deprecated code that is way too old?
- [ ] While Re configuring Cloner, look for room for improvement (maybe delegate expensive cloning process to a faster script, maybe Node?).
- [ ] Maybe it could be worth taking a look at Prince, DocRaptor’s new features to enhance our PDF export capabilities. 
- [ ] I believe this is an excellent opportunity to dismantle the **current menu’s override hacks** and implement them in a sustainable manner.
- [ ] We should consider making the **dashboard’s override optional**, enabling users to customize their dashboards according to their preferences, including the addition of widgets and other features.


arzola:
TODO:

- [x] Start refactoring tests for new classes and migrated classes

arzola:
Today I have an interesting idea. Sometimes it's kinda tricky to know which hook could be the best to hook some actions and classes, so I was thinking that maybe our objects that interact with WP hooks should adhere to the best interface suited for that. That way, we can have a super clean way to build PB objects.

For instance this could be a contract to build admin pages

```php
interface WpPage {
   public function init();
   public function menuRegistration();
   public function pageTemplate();
   public function afterTheme();
   public function context();
   public function permissions();
   public function render();
}
```

And so on...

```php
interface WpTaxonomy {
   public function pluginsLoaded();
}
```

arzola:
**BladeServiceProvider working as expected** 

```php
app('Blade')->render('admin.organize');
```

**DbServiceProvider working as expected** 
```php
app('db')->table('users')->get()
```

arzola:
@fdalcin feedback

- [ ] moving the switch case to a match operator
- [ ] booting providers instead of calling static methods
- [ ] instead of calling new on our classes, let the container build them. This would allow for DI on those classes as well.

PR #3937, opened 2025-03-03
fix: improve conditional before removing H5P menu

SteelWagstaff:
Updates hook so that it's only called on admin pages. Adds checks to ensure a user exists, is not a network manager, and ONLY has the subscriber role before proceeding into the conditional action. Should fix the 'fatal error' when users who aren't logged in attempt to view an embedded H5P activity.

To test:
1. View the URL for an embedded H5P activity while logged out of a book (e.g. `https://university.pressbooks.pub/ltidemo/wp-admin/admin-ajax.php?action=h5p_embed&id=1`). Ensure that you can view the activity
2. Visit the book dashboard where H5P is activated as a super admin or network manager who has only the subscriber role in a given book. Ensure that you see the H5P plugin menu in the sidebar
3. Visit the book dashboard where H5P is activated as a user is not a network manager but has both the Administrator and Subscriber role (see https://github.com/pressbooks/pressbooks/issues/3648) -- or any two roles as long as one of them is Subscriber. Ensure that you see the H5P plugin menu in the sidebar.
4. Visit the book dashboard where H5P is activated as a user who is not a network manager but does have the 'Subscriber' role. Ensure that you do not see the H5P plugin menu in the sidebar. Visit the URL for an embedded H5P activity (as in step 1) and ensure that you can still view the activity as expected.

### DIFF
--- a/hooks-admin.php
+++ b/hooks-admin.php
@@ -322,7 +322,11 @@ if ( $is_book ) {
 		}
 	);
 
-	add_action( 'admin_init', [ SideBar::class, 'handleH5pMenu' ] );
+	add_action( 'admin_init', function() {
+		if ( is_admin() ) {
+			SideBar::handleH5pMenu();
+		}
+} );
 
 	// Hide welcome screen
 	add_action(
--- a/hooks-admin.php
+++ b/hooks-admin.php
@@ -322,11 +322,7 @@ if ( $is_book ) {
 		}
 	);
 
-	add_action( 'admin_init', function() {
-		if ( is_admin() ) {
-			SideBar::handleH5pMenu();
-		}
-} );
+	add_action( 'admin_init', [ SideBar::class, 'removeH5pMenuForSubscribers' ] );
 
 	// Hide welcome screen
 	add_action(
--- a/inc/admin/menus/class-sidebar.php
+++ b/inc/admin/menus/class-sidebar.php
@@ -727,16 +727,19 @@ class SideBar {
 			( $is_main_site_page ? $page : network_admin_url( $page ) );
 	}
 
-	public static function handleH5pMenu(): void {
-		$user_id = get_current_user_id();
-		if ($user_id === 0) {
+	public static function removeH5pMenuForSubscribers(): void {
+		$user = get_user_by( 'ID', get_current_user_id() );
+
+		if (
+			! $user ||
+			! is_admin() ||
+			is_super_admin( $user->ID ) ||
+			$user->roles !== [ 'subscriber' ] ||
+			! is_plugin_active( 'h5p/h5p.php' )
+		) {
 			return;
 		}
-		$user = get_user_by( 'id', $user_id );
-		if (
-			$user && is_plugin_active( 'h5p/h5p.php' ) && !is_super_admin($user->ID) && $user->roles === ['subscriber']
-		) {
-			remove_menu_page( 'h5p' );
-		}
+
+		remove_menu_page( 'h5p' );
 	}
 }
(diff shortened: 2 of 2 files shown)


hooks-admin.php
```diff
@@ -322,7 +322,11 @@
 		}
 	);
 
-	add_action( 'admin_init', [ SideBar::class, 'handleH5pMenu' ] );
+	add_action( 'admin_init', function() {
+		if ( is_admin() ) {
```
SteelWagstaff: Makes sure the function is only triggered on admin pages (not admin-ajax)

inc/admin/menus/class-sidebar.php
```diff
@@ -728,10 +728,13 @@ private function getContextSlug( string $page, bool $is_main_site_page ): string
 	}
 
 	public static function handleH5pMenu(): void {
-		$user = get_user_by( 'id', get_current_user_id() );
+		$user_id = get_current_user_id();
+		if ($user_id === 0) {
+			return;
```
SteelWagstaff: Bail early if no user

inc/admin/menus/class-sidebar.php
```diff
@@ -728,10 +728,13 @@ private function getContextSlug( string $page, bool $is_main_site_page ): string
...
+		}
+		$user = get_user_by( 'id', $user_id );
 		if (
-			is_plugin_active( 'h5p/h5p.php' ) &&
-			in_array( 'subscriber', $user->roles, true )
+			$user && is_plugin_active( 'h5p/h5p.php' ) && !is_super_admin($user->ID) && $user->roles === ['subscriber']
```
SteelWagstaff: Clarified checks: 1) user must exist, 2) H5P plugin must be active, 3) ensure the user is not a super admin/network manager and 4) make sure that the ONLY role they have in the book is subscriber. 

We have weird situations where some users are both administrators and subscribers in the same book -- this will ensure that the menu is not removed for them.

richard015ar: that makes sense. The only thing I'd change is that we don't need to check for the $user here, it is safe to assume the user exists at this point since it was checked here: https://github.com/pressbooks/pressbooks/pull/3937#discussion_r1978361722.
I'll push a friendly commit

richard015ar: another think I'll propose is to check the roles with `in_array`, since `WP_User` object returns the role as an array: https://github.com/WordPress/wordpress-develop/blob/8a52d746e9bb85604f6a309a87d22296ce1c4280/src/wp-includes/class-wp-user.php#L80

SteelWagstaff: Yes, it can be an array, but we only care if the array has a single value and it is subscriber. That's why I checked for `['subscriber']` -- the brackets indicate an array, right?

SteelWagstaff: This way we don't have to worry if there are multiple array values (in the case of a user who accidentally has multiple roles). We're only proceeding if it's a single item array with the value of 'subscriber'

SteelWagstaff: I suppose we could also do something like $user->roles[0] === 'subscriber' but I thought this might be simpler

richard015ar: yes, I've decided to leave as it is, since it makes sense. We fixed the problem of multiple roles, however some legacy users still might have more than 1 role for a book. So, it is better to check as it is, instead of assuming a single role

richard015ar (approved):
LGTM

PR #3943, opened 2025-03-10
Removes filter from footnotes content.

alexdryden:
This soloves a legacy problem and prevents wptexturize() from performing various desired functions, like making appropriate curly quotation marks. See https://pressbooks.community/t/smart-quotes-in-footnotes/2432/7


alexdryden:
Apologies! I opened this up against the wrong repo.

Issue #3959, opened 2025-03-18
Improve display math on exports


arzola:
Written by @tw77 

I found two things related to the requirements in the TODO comment above:

`\[ ... \]` is being handled as display math in webbooks (centred on its own line as expected), but in PDFs and EPUB exports, it's being handled the same way as inline math. If we could centre display math on its own line in our exports, we would have complete support for this delimiter.

The same applies for webbook footnotes, where display math is centred:
<img width="505" alt="Image" src="https://github.com/user-attachments/assets/67a95aec-6a35-438d-9da5-e72fe5151a41" />

But not PDF or EPUB footnotes:
<img width="288" alt="Image" src="https://github.com/user-attachments/assets/0f2cf925-ec65-4795-ad4b-5554f616e797" />

With footnotes specifically, I'm not certain which is preferable but suggest something consistent here.

--

The other thing is that we're handling `$$` as inline math everywhere except webbook glossary terms, where `$$` is being handled as display math. 

In text: <img width="342" alt="Image" src="https://github.com/user-attachments/assets/5ff292d5-4cc1-40a7-ac99-20cabcc773ab" />

In a webbook glossary term (third entry here, others included for comparison):
<img width="508" alt="Image" src="https://github.com/user-attachments/assets/98a0dd9e-97a5-413d-87eb-38a9e11ddcc3" />

I think we should go with a consistent approach here. This could be using `$$` as inline math across the board (if we're using `$$` in place of `$`. I understand `$$` was once used for display math in LaTeX outside of Pressbooks but that `\[ ... \]` is recommended in modern LaTeX)

PR #3960, opened 2025-03-18
fix: display math wrapper

arzola:
Address #3959


fdalcin (approved):
LGTM. Thank you once more @arzola 🎉

PR #3961, opened 2025-03-19
fix: resolve bug with taxonomy terms that use non-latin characters

### LINKED ISSUE #3886: Contributor names in Hindi not showing up in metadata
### Feature Description

When the book's language is already set to Hindi with script support added, authors and contributors in Hindi cannot be added to metadata. 

More specifically, the book info page is not able to save any contributors in Hindi. After saving, the author in Hindi then turned blank. 
image

See test book here: https://kpu.pressbooks.pub/testkz/wp-admin/index.php?page=book_dashboard

If a non-Hindi author is added, the Hindi author will be added as "--2" which was made a new contributor in addition to the original contributor name in Hindi. (see https://opentextbc.ca/isvinternationalhi/)

### Feature Use Case

This will support publication and translation in Hindi. 

### Other Notes

April 30 update: similar situation found with contributors in Punjabi language as well

greatislander:
Resolves #3886.

### Description

Contributors are managed using the [`TaxonomyReorderableMultiselect`](https://github.com/pressbooks/pressbooks/blob/dev/inc/admin/fields/class-taxonomyreorderablemultiselect.php) field. The way this works is that it sets the selected terms and also saves a meta field of comma-separated term slugs in the user-determined order. The problem is that WordPress urlencodes slugs, so — for example — the slug `अरुंधति-राय` would be returned as `%E0%A4%85%E0%A4%B0%E0%A5%81%E0%A4%82%E0%A4%A7%E0%A4%A4%E0%A4%BF-%E0%A4%B0%E0%A4%BE%E0%A4%AF` within the return value of the `get_terms()` function. This was causing the meta value to be stored in a way that prevented the term from being retrieved on the other end— so something was saved but `get_term_by('%E0%A4%85%E0%A4%B0%E0%A5%81%E0%A4%82%E0%A4%A7%E0%A4%A4%E0%A4%BF-%E0%A4%B0%E0%A4%BE%E0%A4%AF')` was not finding the term with the slug `अरुंधति-राय`. My solution here is to run the term slugs through `urldecode` when setting up the field options for `TaxonomyReorderableMultiselect` and `TaxonomySelect` to make sure that the correct value is always stored and can be used by `get_term_by` on the other end.

Users affected by this bug will need to re-enter their contributors once this fix is deployed.

### DIFF
--- a/inc/admin/fields/class-taxonomyreorderablemultiselect.php
+++ b/inc/admin/fields/class-taxonomyreorderablemultiselect.php
@@ -38,7 +38,7 @@ class TaxonomyReorderableMultiselect extends Field {
 		$options = [];
 
 		foreach ( $terms as $term ) {
-			$options[ $term->slug ] = $term->name;
+			$options[ urldecode( $term->slug ) ] = $term->name;
 		}
 
 		return $options;
--- a/inc/admin/fields/class-taxonomyselect.php
+++ b/inc/admin/fields/class-taxonomyselect.php
@@ -33,7 +33,7 @@ class TaxonomySelect extends Field {
 		}
 
 		foreach ( $terms as $term ) {
-			$options[ $term->slug ] = $term->name;
+			$options[ urldecode( $term->slug ) ] = $term->name;
 		}
 
 		return $options;


richard015ar (approved):
Good catch! LGTM, tested and it works. Thank you, Ned!

PR #3963, opened 2025-03-20
fix: add aria-current to theme options page

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1632

This pull request includes a change to the `inc/modules/themeoptions/class-admin.php` file to improve the accessibility and readability of the theme options navigation tabs.

Accessibility and readability improvements:

* Added `aria-current="page"` attribute to the active navigation tab to improve accessibility for screen readers.
* Reformatted the HTML structure for the navigation tabs to enhance readability by breaking long lines and properly indenting nested elements.

### Testing case
- In an admin book area go to Appearance > Theme options
- inspect the tags and ensure that the current active tag contains the `aria-current="page"` attribute.
- The `aria-label` for the `nav` element should be `aria-label="Theme Options` instead of `aria-label="Theme Options Tabs`.


fdalcin (approved):
LGTM

PR #3965, opened 2025-03-20
fix: ensure that top nav items are accessible in mobile

greatislander:
Previously, text labels for top admin menu items were hidden using `display: none` on small screens. This PR instead applies the `screen-reader-text` styles to visually hide them while preserving them so these items have accessible names.

Resolves pressbooks/private#1633.


fdalcin (approved):
LGTM

PR #3966, opened 2025-03-20
fix: properly label checkboxes on export page

greatislander:
See pressbooks/private#1636.

Checkboxes for selecting exported files and pinning exported files should now have meaningful labels.


fdalcin (approved):
LGTM

PR #3969, opened 2025-03-26
fix: recognize all supported math delimiters in web book interface

arzola:
This pull request includes an enhancement to the `sectionHasMath` function in the `inc/class-mathjax.php` file to improve performance and accuracy in detecting math content within sections. The most important changes include adding caching for previously checked sections, extending the types of math content detected, and refining the regex used for LaTeX delimiters.

Performance improvements:

* Added caching mechanism to store and retrieve results of previously checked sections using the `$this->sectionHasMath` array.

Detection improvements:

* Extended the detection to include specific math tags such as `[table id=`, `[/latex]`, `$latex`, `[/asciimath]`, `$asciimath`, and `</math>`.
* Refined the regex pattern to include additional LaTeX delimiters and inline math expressions.

> I just bring back a couple of caching conditionals we used to have on this method and I added the support for previous and our new supported delimiters

### Additional things to consider for testing

* Test every supported delimiter in isolation on a chapter to confirm this fix works as expected

### DIFF
--- a/inc/class-mathjax.php
+++ b/inc/class-mathjax.php
@@ -229,15 +229,32 @@ class MathJax {
 			return true;
 		}
 
+		$id = $post->ID;
+		if ( isset( $this->sectionHasMath[ $id ] ) ) {
+			return $this->sectionHasMath[ $id ];
+		}
+
 		$content = $post->post_content;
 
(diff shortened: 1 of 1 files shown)


richard015ar (approved):
LGTM

PR #3976, opened 2025-04-02
feat: h5p activities will be printed in exports

arzola:
This pull request introduces several updates, including dependency changes, enhanced image validation, new utility methods, and improvements to the EPUB export process. The changes focus on adding support for H5P content extraction, improving maintainability, and modernizing the codebase.

### Dependency Updates
* Updated `composer.json` to include a new repository for `H5PExtractor` and added `snordian/h5p-extractor` as a dependency. The PHP version requirement was also updated from `^8.1` to `^8.2`. [[1]](diffhunk://#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R30-R37) [[2]](diffhunk://#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R64)

### Image Validation Enhancements
* Added a filter `pb_validate_svg` to allow SVG files to pass validation in the `is_valid_image` function.

### H5P Content Support
* Introduced several private methods in `class-h5p.php` to handle H5P content, including `getVendorPath`, `textAddonMatches`, `generateContentSlug`, `createH5PExport`, `ensureH5Export`, and `getH5PRepresentation`. These methods enable H5P content extraction and rendering. [[1]](diffhunk://#diff-7e3631bc1344793e025238d75c1ba7692bc2853d7b36f65aed09c31e8e332983R116-R431) [[2]](diffhunk://#diff-7e3631bc1344793e025238d75c1ba7692bc2853d7b36f65aed09c31e8e332983L161-R505)
* Modified the `replaceShortcode` method to render H5P content using a new Blade template (`interactive.h5pextractor`) when an H5P representation is available.

### EPUB Export Improvements
* Replaced fully qualified exception names (e.g., `\Exception`) with imported `Exception` for consistency and readability. [[1]](diffhunk://#diff-4d36e712f4455de8592afc2af3953c30f7cf23921703a5c4ce5980930b1ed895L402-R418) [[2]](diffhunk://#diff-4d36e712f4455de8592afc2af3953c30f7cf23921703a5c4ce5980930b1ed895L820-R831) and others)
* Added a new `protected string $embbededCss` property to the `Epub` class to manage embedded CSS.

### Utility Functions and Code Cleanup
* Added new utility imports in `class-epub.php`, such as `create_tmp_file` and `put_contents`, to simplify code and improve readability. [[1]](diffhunk://#diff-4d36e712f4455de8592afc2af3953c30f7cf23921703a5c4ce5980930b1ed895R12-R21) [[2]](diffhunk://#diff-4d36e712f4455de8592afc2af3953c30f7cf23921703a5c4ce5980930b1ed895L839-R849)


richard015ar (changes requested):
@otacke thank you for these changes, the proposal is looking great.
However, I've found some things I'd like to share to understand better.

I noticed that "Advanced FIll in the Blanks" activity is not supported by this feature. Is that right? If not, do we have a list of supported or non-supported h5p types so we could document them.

See:
Screenshot 2025-04-07 at 2 16 03 PM
Screenshot 2025-04-07 at 2 15 22 PM

On the other hand, the EPUB export file presents some broken images (in addition to the Advanced fill in the blanks message) with an "?" image, see:
Screenshot 2025-04-07 at 2 17 56 PM
Screenshot 2025-04-07 at 2 17 43 PM
Screenshot 2025-04-07 at 2 17 31 PM
Screenshot 2025-04-07 at 2 17 11 PM
Screenshot 2025-04-07 at 2 16 59 PM

arzola:
@richard015ar thanks, I'll take a look to see if we can fix the EPUB thing for @otacke

otacke:
@richard015ar The heavy lifting is done by the H5PExtractor library (https://github.com/otacke/H5PExtractor) which contains "generators" for quite a bunch of H5P content types/libraries - but not all of them yet, that's true.

https://github.com/otacke/H5PExtractor/tree/master/app/generators should have you covered.

Content types were prioritized by the sponsor. There will be more as time/funding/pull-requests trickle(s) in.

inc/sanitize/namespace.php
```diff
@@ -775,7 +775,18 @@ function sanitize_webbook_content( $content ) {
 
 function filter_export_content( $content ) {
 	remove_filter( 'the_content', '\Pressbooks\Sanitize\sanitize_webbook_content' );
+
+	// Content can be very large which can cause convert_smilies in formatting.php to crash. It runs preg_replace which
```
richard015ar: I will look at alternatives of `init_set`, which has global impact, it also depends on environments and configurations (it might not be possible to change global parameter like this).
I will explore if we can chunk the content, something like:
```php
 function filter_export_content( $content ) {
       remove_filter( ‘the_content’, ‘\Pressbooks\Sanitize\sanitize_webbook_content’ );

       $chunks = str_split( $content, 10000 ); // Adjust chunk size as needed
       $filtered_chunks = array_map( function( $chunk ) {
           return apply_filters( ‘the_content’, $chunk );
       }, $chunks );

       $content = implode( ‘’, $filtered_chunks );

       add_filter( ‘the_content’, ‘\Pressbooks\Sanitize\sanitize_webbook_content’ );
       return $content;
   }
```

I also wonder if `conver_smilies` is relevant to the export process (will look at that). If not, we could filter `the_content`.

otacke: @richard015ar The `convert_smilies` functions is what was called and crashed when content was large. Had to dig to find it. The approach is what I found to be used here (https://github.com/pressbooks/pressbooks/blob/ea2b6bf3738dc54ac8d79505e6a7e2c809bfc854/inc/modules/import/html/class-xhtml.php#L41-L48), so I assumed taking it was okay. 

Citing https://snordian.slack.com/archives/C0156UC7NRG/p1726045948910349?thread_ts=1725978816.609599&cid=C0156UC7NRG:

"The static HTML can become very large though. Its size lead WordPress' `convert_smilies` function that runs on all content and replaces ASCII smilies with emoji characters to fail. It uses `preg_split` to parse the text, and it can return `false` for large inputs, because the preg_backtrack_limit can be reached. There is no guard against not returning an array later in the code however, and the function crashes.

https://github.com/WordPress/WordPress/blob/2f943897c90733b2c45eadd40954e1eda76e4cb7/wp-includes/formatting.php#L3494-L3495

Could in theory also happen with large books without H5P, and A. [name redacted] thinks he has encountered this before.

As a workaround, I temporarily increase the `pcre.backtrack_limit` on the server inside the export_content filter. That approach was also taken by Pressbooks for importing books, so I assume it's not a no-no in general.
A better approach would be to create a fix for WordPress as a pull request: Instead of crashing, the function could check the return value type of `preg_split` and throw/log an error and/or return the input content directly. Returning the input content directly would simply leave ASCII smilies untouched.

After that, with other contents, I still hit a problem with a dependency (forgot the name) that is used for sanitizing URLs. It failed with large base64 encoded assets. It simply would not handle them.

In order to fix this, I added a new feature to H5PExtractor. You can now pass the base URLs of H5P's core, libraries and content data which normally hold the assets. If these URLs are set, H5PExtractor will not base64-encode assets, but construct the URLs pointing to the assets on the server and use these in the HTML output.

Alltogether, the workaround for temporarily increasing the `pcre.backtrack_limit` may not be required anymore, because without the base64 encoding, content will be much smaller. Since the bug in WordPress still exists and since the same issue could occur for large books in general, it may still serve a purpose until the issue in WordPress gets tackled."

richard015ar: Got it, that makes sense. Thank you for the clear explanation. I will consult this with ops team to ensure this is not a problem. If it is, we can just remove it or discuss how to approach it.

richard015ar:
I am trying to fix the EPUB export routine. I cannot find the reason for the "?" images in the H5P rendering.

Something I noticed is that with this the EPUB checker (validator) fails every time using this branch. As soon as I move to `dev` branch, the validation passes successfully.
The EPUB checker logs indicate the following:
```
ERROR(CSS-020): /app/web/app/uploads/sites/70/pressbooks/exports/New-LTI-Book-1744220909.epub/EPUB/chapter-001-chapter-1.xhtml(48,411): CSS font selector declaration uses unexpected font-size value "unset".
ERROR(RSC-006): /app/web/app/uploads/sites/70/pressbooks/exports/New-LTI-Book-1744220909.epub/EPUB/chapter-001-chapter-1.xhtml(48,93430): Remote resource reference not allowed; resource must be placed in the OCF.
ERROR(RSC-007): /app/web/app/uploads/sites/70/pressbooks/exports/New-LTI-Book-1744220909.epub/EPUB/chapter-001-chapter-1.xhtml(48,508196): Referenced resource "EPUB/src" could not be found in the EPUB.
ERROR(RSC-008): /app/web/app/uploads/sites/70/pressbooks/exports/New-LTI-Book-1744220909.epub/EPUB/chapter-001-chapter-1.xhtml(48,74685): Referenced resource "https://pressbooks.ngrok.dev/app/uploads/sites/70/h5p/libraries/H5P.MultiChoice-1.16/css/../fonts/icomoon.ttf" is not declared in the OPF manifest.
```

These 3 validation errors appear in every book multiple times. It seems to be a problem with the CSS injection, but I am not quite sure yet.

richard015ar:
@otacke after a few hours of debugging I couldn't find the problem. Do you have a gist about where it could be the issue? This might need a separate issue on our side to investigate and resolve.

otacke:
@richard015ar The first message I can explain, I guess. As far as I revall, the H5P Extractor uses `font-size: unset` somewhere to shield the H5P content from the stylesheet that is applied to the book. This got necessary, because the H5P renders do not sit in a protective iframe anymore. It's not an obscure or experimental value (https://developer.mozilla.org/en-US/docs/Web/CSS/unset), but it's one of the printing engine's CSS related shortcoming that I learned about the hard way.

This message should not be related to what you're experiencing though.

The other messages look more relevant, but I do not know what they mean. It seems that loading the content type icon image from a "remote" location fails, but I could not tell what "remote" is supposed to mean here. All the files are on the server. There's probably a simple solution if you manage to interpret that message correctly.

I assume that a workaround could be to base64-encode the svg content type icon images that are shown if no renderer is available instead of referencing their respective file. They are small, so the HTML would not grow in size by much. But that's not a real option for the other images that may be used in H5P contents. The HTML size would explode if you base-64-encoded all of those. Led to other issues before. 

No idea what an OPF manifest is that's mentioned in connection with the font file of Multiple Choice.

richard015ar:
@otacke I was inviestigating a little bit more about this EPUB checker message:
```
ERROR(RSC-006): /app/web/app/uploads/sites/70/pressbooks/exports/New-LTI-Book-1744220909.epub/EPUB/chapter-001-chapter-1.xhtml(48,93430): Remote resource reference not allowed; resource must be placed in the OCF.
```
and it means that somewhere in the `chapter-001-chapter-1.xhtml` file (which I could get by unzipping the epub file), there’s a reference to an external (remote) resource, such as an image, CSS, JavaScript, or font, that is hosted outside the EPUB package. I am attaching this zipped file, uncompress it and look for the EPUB/chapter-001-chapter-1.xhtml file. Check the line 49 which is basically a single line with most of the content of this test chapter you will find some references there.

[New-LTI-Book-1744220909.zip](https://github.com/user-attachments/files/19684159/New-LTI-Book-1744220909.zip)

If you do a search there, you will find the inserted css with some things like this:
```
 @font-face { font-family: 'H5PFontIcons'; src: url('https://pressbooks.ngrok.dev/app/uploads/sites/70/h5p/libraries/H5P.FontIcons-1.0/styles/../fonts/h5p.ttf') format('truetype');
 @font-face { font-family: 'h5p'; src: url('') format('truetype'); // There several lines with empty source
// Other lines like:
.h5p-question-minus-one { background-image: url(https://pressbooks.ngrok.dev/app/uploads/sites/70/h5p/libraries/H5P.Question-1.5/styles/../images/minus-one.svg); } 
```

EPUB files are required to be self-contained. That means all assets (images, styles, fonts, etc.) must be included inside the EPUB archive itself, not loaded from the internet. Anything pointing to http:// or https:// is considered a remote resource, which violates EPUB’s OCF rules.

I am not sure if these lines are from the extractor, are those?
If that's the case, we'd need to place the fonts and images within folders in the EPUB's directory (fonts/ or images/ folders) and update the references in the .xhtml to point to local files. 

The other error:
```
RROR(RSC-007): /app/web/app/uploads/sites/70/pressbooks/exports/New-LTI-Book-1744220909.epub/EPUB/chapter-001-chapter-1.xhtml(48,508196): Referenced resource "EPUB/src" could not be found in the EPUB.
```
indicates that there's a reference to EPUB/src which is incorrect because that folder does not exist within our exported EPUB folder.
I am not quite sure, but I think it refers to the lines containing:
```
<img width="100%" height="100%" class="h5p-placeholder" src="src#fixme" alt="image"/>
```
Looking [at the code](https://github.com/pressbooks/pressbooks/blob/fe31b00cf96b8dba68d86039c78cd5bccc318edf/inc/modules/export/epub/class-epub.php#L2069), it seems that we use that for broken links, but the `$url` is defined in the document.

Finally, for:
```
ERROR(RSC-008): Referenced resource "https://pressbooks.ngrok.dev/app/uploads/sites/70/h5p/libraries/H5P.MultiChoice-1.16/css/../fonts/icomoon.ttf" is not declared in the OPF manifest.
```
It has two different issues: we are using external reference (URL) and the `icomoon.tff` file is not included in the epub. Look for the line:
```
{ font-family: 'icomoon-multichoice'; src: url('https://pressbooks.ngrok.dev/app/uploads/sites/70/h5p/libraries/H5P.MultiChoice-1.16/css/../fonts/icomoon.ttf') format('truetype');
```
For this problem, we should download the font and place it in the `fonts` folder. We could use [this funcion](https://github.com/pressbooks/pressbooks/blob/fe31b00cf96b8dba68d86039c78cd5bccc318edf/inc/modules/export/epub/class-epub.php#L1072) to make it work.

Here's the file exported without using this branch, which does not contain these validation problems:
[New-LTI-Book-1744286509.zip](https://github.com/user-attachments/files/19684350/New-LTI-Book-1744286509.zip)
Inpecting it, I could not find the references mentioned, which would helps us to debug the problem and compare.

otacke:
@richard015ar I see.

Yes, the HTML that is generated for the export uses "remote" references then. The HTML requires fonts, stylesheets from the original H5P libraries and media from the respective H5P content files which are all located on the server and can be retrieved via the respective URL. When I understand correctly, then for `ERROR(RSC-006)` it should suffice for Pressbooks to copy those into the EPUB archive and to update the occurrences inside the xhtml accordingly. I have not looked at EPUB generation before, however.

The `RROR(RSC-007)` error is something that I'll have to look into, I guess. Can't tell from the `src="src#fixme"` what it might have been before. But from glancing at the code, the faulty value should come directly from the H5P content, so possibly the path within the H5P content(.json) is broken already, the file went missing, ...

Finally, `ERROR(RSC-008)` then should be related to `ERROR(RSC-006)`.

arzola:
Hello :) I just added some improvements:

* Handle the base64 images in EPUB exports 
* Extracted all `<style>` tags and include them in the main stylesheet
* Images are being rendered properly on EPUB
* Custom fonts fonts included by H5P libraries are being downloaded and referenced locally on the CSS file. 
* Zero validation errors

I included a couple of workarounds to fix CSS issues that doesn't affect the visual representation of the H5P activity, probably we should double check if those attributes can be fixed/removed in the Extractor library.

🚀 EPUB and PDF are being exported with NO warnings.

otacke:
Anything I could be of help with? I think I could spend some time on this next weekend.

arzola:
@otacke hey, thank you It could be helpful if you do some testing with this branch and test EPUB exports to see if they look as expected

richard015ar:
I pulled the latest changes on this branch and updated the dependencies.
After re-exporting the same book I used for testing to EPUB and PDF, I got the same validation issues and image rendering problems persisted in EPUB. In addition, now PDF are not rendering the activities, when previously it did. It might be something in my environment? I'll send it to you @arzola for an inspection/debugging.

richard015ar:
@otacke when exporting a Crossword H5P activity to PDF it seems that I cannot be rendered as expected. It looks good on PDF, tho. See:
Screenshot 2025-04-29 at 1 08 58 PM

Do you think this is something easily fixable?

richard015ar:
Another question for you @otacke or @arzola : would it be possible to remove (or add a hook to do so) the  `No HTML renderer for <type> <version> available.` message? I think the textbox displayed below each activity `The interactive version of this H5P content is available at: <h5p URL link>` would be enough, according to our team's suggestion.

otacke:
@richard015ar That's "Find the Words". This one was a little harder to do, because in it's original version, it uses a Canvas to render the grid - nothing we can do without a browser engine running. I implemented a new render method that interprets the underlying parameters and returns the HTML for common DOM elements. It's well possible that something went wrong while doing that. Will take a look next weekend.

Skipping the "No renderer message" should not be a problem at all.

richard015ar (approved):
LGTM, all the issues mentioned in this PR was solved. Thank you @arzola and @otacke for your contributions!

richard015ar:
only pending issue is the "Find the Words" rendering mentioned earlier which will be tackled in @otacke 's lib soon.

otacke:
@richard015ar I changed the generator for Find The Words to now use the "column" view it would normally have when there's not enough horizontal space, so the list of words to find will be displayed underneath the grid. That should keep everything in bounds no matter where the output is used. Have not checked, because I just wrecked my Pressbooks VM and would need to set it up from scratch :sweat_smile:

otacke:
@richard015ar Anything I can do?

arzola:
@otacke we are working in other branch a full refactor of the export routines to be processed as background processes this week we are on process of testing it and hopefully release it next week

otacke:
@arzola Aye!

PR #3979, opened 2025-04-04
fix: display Thema qualifiers in Subjects metaboxes

SteelWagstaff:
Display Thema qualifiers in the Subject selection metaboxes on the book info page. In the future we may want to update to the latest version of Thema (1.6) and provide a separate select box for qualifiers.


fdalcin (approved):
LGTM. Thank you @SteelWagstaff. I took the liberty to add friendly commits fixing linting on other parts but also including a `named argument` to add context to the parameter you've changed.

PR #3502, opened 2023-10-28
feat: replace custom-metadata-manager

greatislander:
- Resolves #2436
- Resolves #2749
- Partial resolution of #2976 

### Features
 
- [x] Metabox class
- [x] Field classes
	- [x] Field
	- [x] Date
	- [x] Text
	- [x] URL
	- [x] Textarea
	- [x] Wysiwyg
	- [x] Select
	- [x] Taxonomy Select
	- [x] Checkbox
- [x] Book Information page
	- [x] General Information
	- [x] Subjects
	- [x] BISAC Subjects
	- [x] Institutions
	- [x] Copyright
	- [x] About
	- [x] Additional Catalog Information
- [x] Front Matter/Chapter/Back Matter Metadata
- [x] Part Visibility

PR #3982, opened 2025-04-06
feat: ensure checkbox groups use fieldset and legend

greatislander:
This resolves an accessibility issue where grouped radio buttons lacked an appropriately-named grouping element by wrapping all such groups in `<fieldset>` with a visually-hidden `<legend>`. Resolves https://github.com/pressbooks/private/issues/1639.

One note is that it is probably worth refactoring settings pages to use the new blade-based fields templates created in https://github.com/pressbooks/pressbooks/pull/3502.

E.g. instead of https://github.com/greatislander/pressbooks/blob/51c5fc7df2b7a5c544dc1e9491935a5d814e09de/inc/modules/themeoptions/class-ebookoptions.php#L309-L319, we could create a new class in `Pressbooks\Admin\Fields` to handle radio buttons using a blade template. This would also let us replace all the functions in `Pressbooks\Admin\Laf` (e.g. https://github.com/greatislander/pressbooks/blob/51c5fc7df2b7a5c544dc1e9491935a5d814e09de/inc/admin/laf/namespace.php#L1391-L1398) with calls to the fields class. To discuss.


richard015ar (commented):
LGTM, thank you Ned. Also, I like your proposal of using the new blade (fields based) to make it re-usable.

richard015ar (approved):

PR #3983, opened 2025-04-06
fix: apply accessible name to custom styles editor

greatislander:
Resolves https://github.com/pressbooks/private/issues/1641.

See: https://github.com/WordPress/WordPress/blob/36f146497211a7d451dcfa768be4941ae2f06914/wp-admin/js/theme-plugin-editor.js#L402-L409


richard015ar (approved):
LGTM

PR #3986, opened 2025-04-08
feat: Allow network managers to upload custom fonts

SteelWagstaff:
Introduce a comprehensive custom fonts feature for Pressbooks, allowing network administrators to upload, register, and use custom font files across web, PDF, and EPUB exports. The implementation includes backend logic for handling font uploads, frontend admin UI, integration with the SCSS compilation pipeline, and test shims for file uploads. It also updates the font selection logic to include custom fonts in relevant UI and export contexts. Addresses https://github.com/pressbooks/ideas/issues/523. 

**Custom Fonts Management and Integration:**

* Add a new admin submenu page for uploading and managing custom fonts, including a Blade template for the UI and backend handlers for form submission, file validation, and storage. Uploaded fonts are registered as site options and their CSS is generated automatically. Also supports removal of font variants and deleting font families. (`inc/admin/customfonts/namespace.php`, `templates/admin/custom-fonts.blade.php`, `hooks-admin.php`, `inc/admin/laf/namespace.php`, `requires-admin.php`) [[1]](diffhunk://#diff-693213a353bbc9e4550824feefbffa207f1bb4061398e0f9905540e8357d8121R1-R179) [[2]](diffhunk://#diff-c1e32cdbc03a3f29becc20fcdbaece07aaa7b8da0f249a95cbfa556915cb21dcR1-R86) [[3]](diffhunk://#diff-22bd10ffa96e759d116f39767554f68c8559aa9f379e8b6dc6f132854e6a0729R59-R62) [[4]](diffhunk://#diff-f22cdc03da7e566f78087bda7488a4de9313b2fd11f2045771ebaf381ad8283eR635-R643) [[5]](diffhunk://#diff-dfd5946767e93695717fc3f7cc64917df3f34457c644a75eb357112da75b7db9R15)

* Integrate custom fonts into the font selection logic for the ShapeShifter UI, ensuring user-uploaded fonts appear alongside built-in options and are sorted alphabetically. (`inc/class-styles.php`) Checks for deleted fonts and automatically removes them from Shapeshifter selector the next time the relevant them setting page loads if font families are deleted.

* Prepend custom font `@font-face` CSS declarations to the SCSS before compilation for web, PDF, and EPUB exports, ensuring custom fonts are available in all output formats. (`inc/class-globaltypography.php`, `inc/modules/themeoptions/class-weboptions.php`, `inc/modules/themeoptions/class-pdfoptions.php`, `inc/modules/themeoptions/class-ebookoptions.php`) [[1]](diffhunk://#diff-5ff5ba45956033a2f33cf70928056d55757f8ad79c3cc244619f7ab8254c9834R517-R536) [[2]](diffhunk://#diff-f57f4c99a1e1b8e1ba2f76820a5da63e6e36a4c6a00fd67158297ce08a754815R12) [[3]](diffhunk://#diff-f57f4c99a1e1b8e1ba2f76820a5da63e6e36a4c6a00fd67158297ce08a754815R639-R641) [[4]](diffhunk://#diff-3efbd46e49eae8dcf6b262eaf47258e9eff416b2bcfdf466ac425e26b3d95115R14) [[5]](diffhunk://#diff-3efbd46e49eae8dcf6b262eaf47258e9eff416b2bcfdf466ac425e26b3d95115R2240-R2242) [[6]](diffhunk://#diff-f398dd033d11b1a344770234dd9652af9945a41cba95d9018c9bdb041e14b819R11-R12) [[7]](diffhunk://#diff-f398dd033d11b1a344770234dd9652af9945a41cba95d9018c9bdb041e14b819R633-R635)

* Add test-only shims for file upload functions in the custom fonts namespace to facilitate PHPUnit tests, and update the test bootstrap to load these shims. (`tests/bootstrap-customfonts-shim.php`, `tests/bootstrap.php`) [[1]](diffhunk://#diff-c55fc7029451c10e1fc89c2f85e55f8bc00b4124260578239a360179f00440edR1-R29) [[2]](diffhunk://#diff-96a82592013e5f4b91e682332583b7a4a6ca1ff0431d267c61179b1a21b167daR30-R34)

* Add minor updates for Sass handling, including a new `getVariables` method for debugging and a typo fix in a comment. (`inc/class-sass.php`) [[1]](diffhunk://#diff-8bdbba8d5cba56b923b3e62374cac2e44485c781d5f186e0e0c7552b9a87aed2R37-R43) [[2]](diffhunk://#diff-8bdbba8d5cba56b923b3e62374cac2e44485c781d5f186e0e0c7552b9a87aed2L313-R320)
 
* Update PHPUnit configuration to include polyfills and remove unnecessary exclusions. (`phpunit.xml`)


inc/class-sass.php
```diff
@@ -34,6 +34,13 @@ public function setVariables( array $vars ) {
 		$this->vars = array_merge( $this->vars, $vars );
 	}
 
+    /**
+     * Get variables. Helpful for debugging.
+     */
+    public function getVariables() {
```
SteelWagstaff: This class had a setter but not a getter, so I created one.

inc/class-sass.php
```diff
@@ -310,7 +317,7 @@ protected function logException( \Exception $e ) {
 	}
 
 	/**
-	 * Write CSS to a a debug dir
+	 * Write CSS to a debug dir
```
SteelWagstaff, line 320: Fixed a typo

inc/admin/customfonts/namespace.php
```diff
@@ -0,0 +1,163 @@
+<?php
+/**
+ * @author Steel Wagstaff
+ * @license  GPLv3 (or any later version)
+ */
+
+namespace Pressbooks\Admin\CustomFonts;
```
SteelWagstaff, line 7: I'm not super experienced with creating namespaces, so I might be doing this wrong, but I'm trying new things!

hooks-admin.php
```diff
@@ -54,6 +54,10 @@
 add_action( 'admin_menu', '\Pressbooks\Admin\Diagnostics\add_menu', 30 );
 add_action( 'init', [ '\Pressbooks\Admin\SiteMap', 'init' ] );
 add_action( 'init', '\Pressbooks\Admin\Laf\remove_emoji' );
+
+// Custom Font Actions
+add_action( 'admin_post_pb_save_custom_fonts', '\Pressbooks\Admin\CustomFonts\handle_form_submission' );
```
SteelWagstaff, line 61: Add form submission handler action. I think this is the right place for that, but not 100% sure.

templates/admin/custom-fonts.blade.php
```diff
@@ -0,0 +1,81 @@
+{!! $notices ?? '' !!}
+
+<div class="wrap">
+    <h1>{{ __('Upload Custom Font', 'pressbooks') }}</h1>
+    <p>Upload custom font files for any additional font families you want to make available for books on your network. Permitted file types: .otf, .ttf, .woff, .woff2.</p>
```
greatislander: Localize this.

SteelWagstaff:
@cmurtagh I discussed this PR with @greatislander today. He will perform initial review and add any comments/suggestions.

SteelWagstaff:
@codecov

templates/admin/custom-fonts.blade.php
SteelWagstaff: Done

SteelWagstaff:
@greatislander my tests could definitely use some work. I'm very open to suggestions/recommendations here. I'm still a novice when it comes to unit tests.

inc/class-styles.php
```diff
@@ -733,32 +733,46 @@ public function isShapeShifterCompatible() {
...
-			'Libre Franklin' => __( 'Libre Franklin', 'pressbooks' ),
-			'Montserrat' => __( 'Montserrat', 'pressbooks' ),
-			'Noto Sans' => __( 'Noto Sans', 'pressbooks' ),
-			'Open Sans' => __( 'Open Sans', 'pressbooks' ),
-			'Raleway' => __( 'Raleway', 'pressbooks' ),
-			'Roboto' => __( 'Roboto', 'pressbooks' ),
-			'Rubik' => __( 'Rubik', 'pressbooks' ),
-			'Source Sans Pro' => __( 'Source Sans Pro', 'pressbooks' ),
+			'Alegreya Sans' => 'Alegreya Sans',
+			'Barlow' => 'Barlow',
```
greatislander, line 748: It looks like Atkinson Hyperlegible Next was removed from this list.

templates/admin/custom-fonts.blade.php
```diff
@@ -0,0 +1,81 @@
...
+            </tr>
+
+            <!-- Font Fallback Options -->
+            <tr>
+                <th><label for="font_fallback">{{ __('Font Fallback', 'pressbooks') }}</label></th>
+                <td>
+                    <select name="font_fallback" id="font_fallback" required>
+                        <option value="sans-serif">{{ __('Sans-serif', 'pressbooks') }}</option>
+                        <option value="serif">{{ __('Serif', 'pressbooks') }}</option>
+                    </select>
```
greatislander: I'd suggest using radio buttons here instead.

templates/admin/custom-fonts.blade.php
```diff
@@ -0,0 +1,81 @@
...
+                <th>{{ __('Font Fallback', 'pressbooks') }}</th>
+            </tr>
+            </thead>
+            <tbody>
+            @foreach ($fonts as $font)
+                <tr>
+                    <td>{{ esc_html( $font['name'] ?? '' ) }}</td>
+                    <td>
+                        @foreach (($font['files'] ?? []) as $variant => $file)
+                            <a href="{{ esc_url( $file['file'] ?? '' ) }}" target="_blank">{{ esc_html( ucwords(str_replace('_', ' ', $variant)) ) }}</a><br>
```
greatislander: You'll need to add `(opens in new window)` here as visually hidden text if you're using `target="_blank"`.

greatislander (changes requested):
Some initial feedback. Doing a deeper dive now.

greatislander (commented):
I'm not sure that the download links in the Registered Fonts table are that useful. If you want to keep them they need to have visually hidden text with the font name otherwise you might end up with a number of rows with links that all say "Regular", "Italic", etc. I would also suggest adding the `download` attribute to the link: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#download

greatislander (changes requested):
It looks like you can modify existing fonts by uploading with the same name. I don't think this is a good approach as it is not intuitive and risks data loss (someone accidentally overwriting a font which is already in use). I think you should build in some sort of editor for an existing font record. We could discuss different approaches to that, @SteelWagstaff.

greatislander (commented):
I assume there's no ability to delete fonts because they might be used somewhere? I think the implications of this feature raise some questions particularly when there are multiple administrators managing the custom fonts. It would be good to discuss the expectations for the feature and possible edge cases.

inc/admin/customfonts/namespace.php
```diff
@@ -0,0 +1,165 @@
...
+			$format = isset( $format_map[ $ext ] ) ? $format_map[ $ext ] : 'truetype';
+			$custom_css .= "@font-face {
+			font-family: '{$family}';
+            font-style: {$style};
+            font-weight: {$weight};
+            src: url('{$url}') format('{$format}');
+        }\n\n";
+		}
+
+		$css_file_path = WP_CONTENT_DIR . '/uploads/assets/custom-fonts/_custom-fonts.scss';
```
greatislander: This doesn't need to be SCSS; I'd suggest writing a `.css` file and it can be imported as such.

greatislander (dismissed):

inc/admin/customfonts/namespace.php
SteelWagstaff: Ok -- I'll need to adjust the accompanying PR as well then

SteelWagstaff:
@greatislander I made several small changes to address your initial review comments. I'd be happy to discuss options for managing/deleting existing file uploads, but am wary of over-engineering a solution instead of giving people something that at least meets their minimal needs sooner rather than later. The users of this feature will all be network managers who generally 'know what they're doing' or can be given simple instructions for how to use the available feature (including how to replace font files).

inc/admin/customfonts/namespace.php
SteelWagstaff: Done

cmurtagh:
@greatislander When you have a moment, could you have a look at this and see if there's anything else needed for this. Happy to talk about it tomorrow too.

SteelWagstaff:
@greatislander occasionally Google or others will release a new version of a typeface, sometimes with extra glyphs, etc. The current behavior would allow a network manager to easily replace an existing font with its replacement, I believe

arzola (changes requested):
@MichelleWeremczuk and I think that's important to being able to delete uploaded fonts, specially when they were uploaded by mistake, I know we can "override" the uploaded font with a new one but the previous font remains in the uploads directory.

Right now is like

image

SteelWagstaff:
Allowing network manager to delete typeface will add some complications for users at the book level, especially if a book has already chosen to use an uploaded font in the Shapeshifter array. What would you expect the fallback to return to if a font a book user had selected has been deleted? How/when would the fallback be implemented and saved?

inc/admin/customfonts/namespace.php
```diff
@@ -0,0 +1,323 @@
...
+			}
+		}
+	}
+
+	unset( $fonts[ $slug ] );
+	update_site_option( 'pressbooks_custom_fonts', $fonts );
+	generate_custom_font_css();
+
+	// Reset theme options for books that were using this font
+	reset_books_using_font( $font_name );
```
SteelWagstaff: @arzola this is the piece I worry most about. it's going to be an expensive, time consuming operation, especially for large networks.

inc/admin/customfonts/namespace.php
```diff
@@ -0,0 +1,323 @@
...
+	return;
+}
+
+/**
+ * Iterate over all sites in the network and clear any shapeshifter font selections
+ * that reference a deleted font, resetting them to the theme default (empty string).
+ *
+ * @param string $font_name The human‑readable font family name that was deleted.
+ */
+function reset_books_using_font( string $font_name ) {
```
SteelWagstaff: @arzola I'm not sure if there's a better way than this. Open to ideas.

SteelWagstaff: Ideally, we'd make this lazy -- leave the setting as is, make sure the font-family stack contains stable fallbacks and simply update the setting the next time a book admin touches the theme settings. Not sure if that's feasible though.

SteelWagstaff: I think this could be better approach:

Instead of proactively resetting every book, *do nothing at deletion time and let the font gracefully fall back at render time*. Font values only matter when SCSS is compiled. If the font name in the option doesn't match any available font, it simply won't render — the browser/PDF engine will use the CSS fallback stack. We could then add a lightweight check to clean up broken selections whenever a book's theme options page loads

Proposed approach:

1. Remove `reset_books_using_font()` entirely
2. In `getShapeShifterFonts()`, the deleted font will no longer appear in the dropdown — so no new books can select it
3. In the SCSS compilation path (e.g. `class-weboptions.php`), the font name is just set as a SCSS variable. If the `@font-face` declaration no longer exists (because `generate_custom_font_css()` removed it), the browser simply falls back to the next font in the CSS font-family stack (which will typically be system `serif` or `sans-serif` fonts).
4. Add a lightweight check when a book's theme options page loads to clean up stale references one book at a time. Key improvement would be that this is ONLY run when someone actually visits an affected book's theme settings.

SteelWagstaff:
I pushed a change that iterates through every book to reset theme options whenver a font is deleted. I think that's an expensive and unneccesary approach upon more careful consideration. Instead of proactively resetting every book, I think it would be wiser to do nothing at deletion time and let the font gracefully fall back at render time. Font values only matter when SCSS is compiled. If the font name in the option doesn't match any available font, it simply won't render; the browser/PDF engine will use the CSS fallback stack. We could then add a lightweight check to clean up broken selections whenever a book's theme options page loads

Proposed approach:

1. Remove `reset_books_using_font()` entirely
2. In `getShapeShifterFonts()`, the deleted font will no longer appear in the dropdown — so no new books can select it
3. In the SCSS compilation path (e.g. `class-weboptions.php`), the font name is just set as a SCSS variable. If the @font-face declaration no longer exists (because `generate_custom_font_css()` removed it), the browser simply falls back to the next font in the CSS font-family stack. We just need to make sure that the font-family stack includes theme defaults (after any shapeshifter-specific values)
4. Add a lightweight check when a book's theme options page loads to clean up stale references one book at a time. Key improvement would be that this is ONLY run when someone actually visits an affected book's theme settings.

SteelWagstaff:
If we go this route, we have a slightly different problem. Suppose you uploaded a custom font and used it via Shapeshifter in a book. Once you delete this font family, the book will display our fallback stack (which includes default system serif/sans-serif fonts, instead of the theme default), unless/until a book admin visits & saves the relevant theme options page.

We could do one of two things:
1. Accept this brief degradation to serif/sans-serif only during the window between deletion and the next admin visit/save of theme options (when lazy cleanup fires and CSS recompiles). 

2. Always load the theme defaults alongside the custom font in [_fonts.scss] and use CSS variables. Better fallback, small permanent overhead, since we'd be loading the theme's default fonts even when we don't actually need them (i.e. when the custom font is NOT deleted).

If we wanted to do option 2, we'd need to make this change in pressbooks-book (and child themes) by updating the font-family stack login in three places:

https://github.com/pressbooks/pressbooks-book/blob/128361c4a3d6bbd61077c7ff4932df42b6ab58f4/assets/styles/web/_fonts.scss#L18-L31

https://github.com/pressbooks/pressbooks-book/blob/128361c4a3d6bbd61077c7ff4932df42b6ab58f4/assets/styles/prince/_fonts.scss#L18-L36

https://github.com/pressbooks/pressbooks-book/blob/128361c4a3d6bbd61077c7ff4932df42b6ab58f4/assets/styles/epub/_fonts.scss#L18-L32

I think it's fine going the simpler route in step 1. At present, I have javascript warning that displays when a network manager attempts to delete a font family that reads: `Are you sure you want to delete this font family? Any books currently using it will need to have their theme settings updated to replace this font with a desired replacement.`

assets/src/styles/admin.scss
```diff
@@ -400,3 +400,16 @@ html #wpadminbar {
     }
   }
 }
+
```
SteelWagstaff, line 403: Styles for new font display table.

SteelWagstaff:
Several updates pushed. You should now be able to remove font variants and delete font families with a brief javascript confirmation that must be consented to before the form is submitted and the fonts are deleted/removed. Table layout and styling now matches @MichelleWeremczuk recommendations and are responsive.

arzola (approved):
LGTM and works as expected

PR #4004, opened 2025-04-16
fix: remove unnecessary alt text

fdalcin:
Partial fix for pressbooks/private#1581

This PR removes all alt text from the images in the dashboard since all of them are decorative.
In addition, I opted to make the `<a>` and ` tags single line since there was no reason to have attributes spread on multiple lines anymore.


richard015ar (approved):
LGTM

PR #4008, opened 2025-04-16
fix: make PHP compatibility error handler filterable

SteelWagstaff:
Remove hardcoded paths from PHP deprecation notice error handling. Adjust env variable name to make generic. Add a function that filters user provided paths to handle future deprecation notices.


inc/utility/class-errorhandler.php
```diff
@@ -16,39 +16,50 @@ class ErrorHandler {
...
+                return true;
+            }
+        }
 
-		return E_DEPRECATED === $errorno && count( $is_errfile_in_silenced_path ) > 0;
+		return false;
 	}
 
+    protected function getPathsToSilence(): array {
+        $paths = [
```
SteelWagstaff: there are no known paths that produce current PHP deprecation notices, so I emptied out the array and provide a filter to allow anyone to add their own paths

inc/utility/class-errorhandler.php
```diff
@@ -16,39 +16,49 @@ class ErrorHandler {
 	static public function init() {
 		if ( is_null( self::$instance ) ) {
 			self::$instance = new self();
-			set_error_handler( [ self::$instance, 'silenceDeprecationsNotices' ] );
+			set_error_handler( [ self::$instance, 'silenceDeprecationNotices' ] );
```
SteelWagstaff, line 19: fix typo

arzola (approved):
LGTM I added tests and improved the logic

inc/utility/class-errorhandler.php
```diff
@@ -16,39 +16,47 @@ class ErrorHandler {
...
-	 * @return bool
+	 * @param int $errorno Only E_DEPRECATED errors are considered.
+	 * @param string $errfile Used to check source path.
+	 * @return bool True to suppress, false to allow.
 	 */
-	public function silenceDeprecationsNotices( int $errorno, string $errstr, string $errfile, int $errline ): bool {
-		if ( env( 'WP_ENV' ) === 'development' && env( 'DISPLAY_PHP8_1_DEPRECATIONS' ) === true ) {
+	public function silenceDeprecationNotices( int $errorno, string $errfile ): bool {
+		if ( env( 'WP_ENV' ) === 'development' && env( 'DISPLAY_PHP_DEPRECATIONS' ) !== true ) {
+			return true;
```
arzola, line 35: This is the right logic, we want to return true (silence errors) if 'development' and DISPLAY_PHP_DEPRECATIONS is not true (meaning not defined or false)

arzola: Please test @SteelWagstaff this on your local I think this is the expected behaviour

SteelWagstaff: Well, let me think about this. Where do we want to silence deprecation notices? I think we always want to silence them except for 
development environments (condition 1) 
where DISPLAY_PHP_DEPRECATIONS is explicitly set to true (condition 2). 

If we're in staging or production environments, we never want to show deprecation notices to users. In dev environments, we only want to show them if the developer sets an environment variable to show them. 

Otherwise the function should return true and notices should be silenced. I think we had it right originally. 

The reason why this is confusing is that the function is named 'silence' notices, not 'show' notices.

arzola (changes requested):
Let's discuss about this first

SteelWagstaff:
@arzola can we merge this?

arzola (approved):

PR #4010, opened 2025-04-18
fix: prevent horizontal overflow of admin top bar

greatislander:
Resolves pressbooks/private#1585.


richard015ar (approved):
It works as expected, thank you @greatislander 
Screenshot 2025-04-22 at 8 22 52 AM

richard015ar:
I found an unexpected bug after this change. If you hover on the "Administer Network" top bar menu item or the "My Books" one and you have books, a vertical scroll appears at the top and the sub-items menu are not visible"
Screenshot 2025-04-22 at 8 48 53 AM

@arzola is it ok if I revert this PR? Or do we need to do something before this?

arzola:
I think we should revert that in order to unblock all other items ready to release @richard015ar

greatislander:
Ah darn. Okay, I'll look into this further.

fdalcin:
> Ah darn. Okay, I'll look into this further.

Hey @greatislander I've opened this PR which I think fixes it https://github.com/pressbooks/pressbooks/pull/4015. It does some other things like fixing linting and bumping dependencies. But it should fix the overflow issue as well.

greatislander:
@fdalcin can you request my review on that one?

fdalcin:
Done

fdalcin:
@greatislander done!

PR #4015, opened 2025-04-22
fix: prevent horizontal overflow of admin top bar

fdalcin:
Resolves pressbooks/private#1585.

This ticket resolves pressbooks/private#1585 by applying a auto scroll behavior on mobile screens.


greatislander (commented):
@fdalcin I'm working on a different approach—I think the best bet is to change the icons to 24px on small screens, as this will fit and also meets [WCAG SC 2.5.8](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html). PR incoming.

fdalcin:
@greatislander  interesting approach. Thanks for checking that.

fdalcin:
Closing in favor of #4048

PR #4016, opened 2025-04-22
fix: missing aria-describedby on date inputs

arzola:
This pull request introduces accessibility improvements for date picker components by ensuring proper ARIA attributes are set. The changes enhance support for assistive technologies by linking the `aria-describedby` attribute of the input field to its corresponding description.

### Accessibility Improvements:

* [`assets/src/scripts/book-information.js`](diffhunk://#diff-3380be9a21d4863b939dbe34f39810b90eec866b505b7f9704b174e2249b682eR4-R13): Added an event listener to `duet-date-picker` elements to dynamically set the `aria-describedby` attribute on the associated input field, ensuring it references the correct description.
* [`templates/metaboxes/fields/date.blade.php`](diffhunk://#diff-f6972895cbf9855016670681b02fdb7aec672a4951df8127b92654589a1ce2f8L2-R5): Updated the `duet-date-picker` component to include an `aria-describedby` attribute pointing to the description element, and added an `id` to the label for better accessibility.

### How to test.

Navigate book info inputs with voice over activated when datepicker inputs are focused It should say the input's name + the description below the input


richard015ar (approved):
it works as expected, and the screen reader announces the description, for example the "This is added to the metadata in your ebook” text in the publication date field in the Book Info form.
I thought that we'd need to forward the `aria-describedby` to the `<input>` inside the component as well by adding `input-aria-describedby` to make the screen reader announce it, but it seems that it is not necessary. 

A note (out of the scope of this change): there is a duet-date picker in the network catalog filter.

It LGTM.

PR #4017, opened 2025-04-22
feat: collect book admin metadata

fdalcin:
Issue pressbooks/bi-data-manager#604

This PR addresses the first point of this comment https://github.com/pressbooks/bi-data-manager/issues/604#issuecomment-2815836039. It collects/updates the ids of the book admins as a comma separated string.

Related PRs: https://github.com/pressbooks/pressbooks-network-analytics/pull/399 + https://github.com/pressbooks/pb-cli/pull/74


inc/datacollector/class-book.php
```diff
@@ -143,12 +144,68 @@ public static function init() {
...
+			callback: fn () => $obj->updateBookAdmins( get_current_blog_id() ),
+			priority: 999
+		);
+
+		add_action(
+			hook_name: 'remove_user_role',
+			callback: fn () => $obj->updateBookAdmins( get_current_blog_id() ),
+			accepted_args: 999
+		);
+
```
cmurtagh: Do we even need these if the collector runs daily and re-creates the list from scratch?

cmurtagh: Ok, Steel explained why this is useful, since it is also visible in real time in the book list.

richard015ar (changes requested):
It is possible to delete a user directly, which won't update the `pb_book_admins` blog metadata.
Maybe we could add:
```php
add_action(
    'deleted_user',
    fn ( int $user_id ) => $obj->updateAllBooksAdmins( get_current_blog_id(), $user_id ),
    999
);
```
to cover that case.

richard015ar (approved):
LGTM.

PR #4018, opened 2025-04-22
fix: navigate tinyMCE toolbar with shift + tab

richard015ar:
Issues:
- https://github.com/pressbooks/private/issues/1587
- https://github.com/pressbooks/private/issues/1706

This pull request introduces a new function to customize TinyMCE editor settings and custom the navigation behaviour. The changes enhance the editor's behavior by adding a keyboard navigation feature for improved accessibility.

### TinyMCE customization and accessibility improvements:

* **Added `mce_settings` function**: A new function `mce_settings` was introduced in `inc/editor/namespace.php` to modify TinyMCE settings. It includes a JavaScript callback that enables Shift+Tab navigation to focus on the last button in the toolbar, improving accessibility.
* **Integrated `mce_settings` into WordPress filters**: The `mce_settings` function was registered with the `tiny_mce_before_init` filter in `hooks-admin.php`, ensuring it is applied when initializing the TinyMCE editor.

### Testing case
- Go to a chapter edition page or Book info page (or whenever you find a tinyMCE editor in the book admin area)
- Navigate to the tinyMCE content using TAB
- Once you are in the content, press Shift + TAB key to access to the tinyMCE toolbar
- You should be able to navigate and select/use any of the options available, including the "?" helper
- as the helper indicates, you can press `ESC` to go back to the content section.


richard015ar:
tagging @greatislander in case you have time to take a look at this. Would you please review this?

inc/editor/namespace.php
```diff
@@ -515,6 +515,27 @@ function show_kitchen_sink( $args ) {
 	return $args;
 }
 
+function mce_settings( $settings, $editor_id ) {
+	$settings['init_instance_callback'] = <<<JS
```
richard015ar, line 523: I tried to make this in a separate file, but babel will complain when compiling:
```
Module build failed (from ./node_modules/babel-loader/lib/index.js):
SyntaxError: /app/web/app/plugins/pressbooks/assets/src/scripts/tinymce-settings.js: Unexpected token (1:8)

> 1 | function(editor) {
    |         ^
```
and that because it needs a function name (or export it). I tried some things like exporting the method and overriding `tinyMCE.init` to use this new exported function instead of the WP way (`tiny_mce_before_init` hook).
I also tried to avoid webpack compilation and read the raw file. But at the end, given the size and low complexity of this JS lines, I decided to leave it as it is.

greatislander (commented):
This looks good to me, @richard015ar. My one question is if this works on Windows? I only have access to a Mac to test.

SteelWagstaff:
@cmurtagh Is Windows testing still needed for this? I have access to a Windows machine and can test locally if desired.

SteelWagstaff:
Tested locally on machine running Windows. 
**Good news:** Shift + Tab does send focus from WYSIWYG editor into the toolbar. ESC does return focus to the WYSIWYG editor. **Bad news:** However, there's no way to `Shift` or `Shift + Tab` OUT of the toolbar. Pressing Shift / Shift + Tab inside the toolbar just loops endlessly through the toolbar options. Not sure if this is expected or desired behaviour.

Here's what I see when accessing from a Windows machine:

https://github.com/user-attachments/assets/9fbebdf5-2ab3-42e0-a469-039ef4040632

If I want to navigate to a link that occurs prior to the toolbar, like the chapter title, I have to ESC and then Shift all the way through all options in page and cycle back to the element I want to focus on that occurs prior to the toolbar. If I keep Shifting and wind up in the toolbar, I get caught back in this loop. 

**Suggestion:** Could we improve this so that pressing `Shift + Tab` when focus is on the first element in the toolbar (the paragraph/Heading dropdown) sends focus to the Visual/Code toggle selector? We might also consider making it so that pressing `Shift` when focus is on the last element in the toolbar (insert Glossary Term' sends focus to the WYSIWYG editor -- though the user can achieve the same results by clicking ESC, so maybe this second suggestion is not as important.

PR #4023, opened 2025-04-23
Change Touro College to Touro University

MichSpieler:
Change name for existing institution from "Touro College" to "Touro University"


SteelWagstaff (approved):
LGTM

PR #4049, opened 2025-05-18
feat: switch to Coloris colour picker

greatislander:
Replaces instances of WordPress' built-in Iris colour picker with [Coloris](https://coloris.js.org) which is much more accessible.

- [x] Implement Coloris
- [x] Support internationalization of colour picker labels, [see here](https://coloris.js.org#accessibility-and-internationalization)
- [x] Adjust styling of colour pickers to match Pressbooks UI

### Testing

Using VoiceOver or another screen reader, ensure that all colour picker UI elements are labelled and that each colour picker has a unique label to open it corresponding with the name of the field.

### Other Notes

Some assets which were no longer used have been removed from the `assets/dist` directory.

### Rated Issues

Resolves https://github.com/pressbooks/private/issues/1705, resolves https://github.com/pressbooks/private/issues/1582.


fdalcin (approved):
This PR looks good to me and the labels were properly announced. I tested it locally and it works as expected on all color pickers I could find. I pushed a friendly commit to resolve conflicts with previously merged PRs

PR #4003, opened 2025-04-15
fix: convert pressbooks table class to trait

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1631

This pull request includes several changes aimed at improving the sorting functionality and code structure within the project. The most important changes include adding sorting buttons, updating the `PressbooksTable` class to a trait, and adjusting styles for better user experience, including the sorting action for the contributors table (book admin).

### Improvements to sorting functionality:

* [`assets/src/scripts/contributors.js`](diffhunk://#diff-d32533293436ca4842065cc7ad39b393f588b2a33792ffcdf060a47f4b71da62R270-R285): Added a sorting button to the name header, allowing users to sort by name through a button click.
* [`assets/src/scripts/pressbooks-table.js`](diffhunk://#diff-70712e28334da6c3f55d1767e4a63ada12a9044268a31802f994776a70a5ca31R1-R9): Added an event listener for sorting buttons to handle sorting functionality when the document is ready.
* [`templates/admin/pressbooks-table/column-headers.blade.php`](diffhunk://#diff-5300f391a3ace2bf88527adac6279cbef7c8592a531c9d864beb76c6dd9b7643L4-R13): Updated column headers to use `data-href` for sorting URLs and adjusted the button label rendering.

### Code structure improvements:

* [`inc/admin/class-network-managers-list-table.php`](diffhunk://#diff-2d5a6bcf31af19f849b1fc61b055d222b4d9630987e1c60eb921fe9cd4fb3a9fL19-R21): Changed `Network_Managers_List_Table` to extend `\WP_List_Table` and use the `PressbooksTable` trait.
* [`inc/admin/class-pressbookstable.php`](diffhunk://#diff-e805a8d194105ca3a8f5cb9483cce4a75ec1ef18e34af7f9f9c8c69bd92726d1L6-R7): Converted `PressbooksTable` from a class to a trait, allowing it to be used more flexibly across different classes.
* [`webpack.mix.js`](diffhunk://#diff-974472aeeb3c0d4802bb4087cf02e9cf5591004b99361954801b10b64d7a1f0bR69): Added the new `pressbooks-table.js` script to the build process.

### Style adjustments:

[...]

PR #3990, opened 2025-04-08
feat: add accessible wp table headers

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1631

This pull request includes significant changes to the table sorting functionality and the addition of a new `PressbooksTable` class. This class is intended to be use whatever we use WP_List_Table, and it replaces links from the column headers with buttons, and also adds screen reader text for a11y improvements. The main changes involve the styling of table sorting indicators, the introduction of a new class for handling table rendering, and the corresponding updates in tests and templates.

For this PR and just as an example, it is being used only in the Network Managers table (Settings > Network Managers).

### Table Sorting and Styling:

* [`assets/src/styles/pressbooks-table.scss`](diffhunk://#diff-e4d4fa95d4c193d97226eb1fa4212c0128427712eca8977cb265db00d074b526R1-R45): Added new styles for sorting indicators and buttons to enhance the visual appearance and accessibility of sortable table columns.

### Class and Functionality Enhancements:

* [`inc/admin/class-network-managers-list-table.php`](diffhunk://#diff-2d5a6bcf31af19f849b1fc61b055d222b4d9630987e1c60eb921fe9cd4fb3a9fL19-R19): Changed the `Network_Managers_List_Table` class to extend the new `PressbooksTable` class instead of `WP_List_Table`.
* [`inc/admin/class-pressbookstable.php`](diffhunk://#diff-e805a8d194105ca3a8f5cb9483cce4a75ec1ef18e34af7f9f9c8c69bd92726d1R1-R81): Introduced the `PressbooksTable` class, which extends `WP_List_Table` and includes methods for rendering column headers with sorting functionality.

### Template Updates:

* [`templates/admin/pressbooks-table/column-headers.blade.php`](diffhunk://#diff-5300f391a3ace2bf88527adac6279cbef7c8592a531c9d864beb76c6dd9b7643R1-R16): Updated the template to render column headers as buttons with sorting indicators and screen reader text for accessibility.

### Enqueueing Styles:

* [`inc/admin/networkmanagers/namespace.php`](diffhunk://#diff-54c18814f014ed6672848bc473a8db7b8360e3d13b169aeb939a9a516e234245R36): Enqueued the new `pressbooks-table.css` stylesheet to ensure the new styles are applied.

### Testing:

[...]

PR #4056, opened 2025-05-23
fix: revert unneccessary changes to table sorting

greatislander:
In #4003, changes were made to use buttons for sorting tables instead of links (see also #3990 and pressbooks/private#1631). However, in a call to review the audit, I noted that this was a mistake in the audit (confirmed with Lisa at IDRC):

> The elements that sort the table function as buttons, but have link roles. As a result users may be lead to believe that the table headers will redirect to other pages.

Reading this closely it's important to note the crucial detail: _users may be lead to believe that the table headers will redirect to other pages_. The table headers ARE links which redirect to other pages— for example, if I click the "Username" header on the Network Managers page, I am following a link to: https://pressbooks.test/wp-admin/network/settings.php?page=pb_network_managers&orderby=user_login&order=asc

If I click it again on that page, I am following a link to: https://pressbooks.test/wp-admin/network/settings.php?page=pb_network_managers&orderby=user_login&order=desc

Changing these links to buttons is not an accessibility improvement and in fact is a regression. I'm sorry I didn't catch this earlier when my review was requested. The key here is: if it's a link to a different URL, it should be an `<a>` element; if it changes the page without visiting a different URL, it should be a `<button>` element.


cmurtagh:
Adding @arzola @richard015ar @fdalcin as reviewers to this, and they'll have better feedback than I.

arzola (approved):
LGTM I added a PR https://github.com/greatislander/pressbooks/pull/1 to merge dev and solve conflicts

arzola (changes requested):
We need to fix the tests I opened a friendly PR against @greatislander branch https://github.com/greatislander/pressbooks/pull/1

arzola (approved):

PR #4064, opened 2025-05-28
feat: robots privacy setting

fdalcin:
Issue pressbooks/private#1745

This PR exposes a new Sharing & Privacy setting aiming to allow clients to better control how they want AI bots our search engines to ingesting/indexing their books.


richard015ar (changes requested):
After enabling the AI I can see the meta is correct:
Screenshot 2025-05-29 at 2 59 06 PM

However, if I check in the first checkbox " Discourage AI from ingesting this book.", the another one ("Discourage crawlers and search engines from indexing this book.") if checked on by default, see:
[robots-settings.webm](https://github.com/user-attachments/assets/a9e9f448-a574-437c-819d-47a3a77d70c5)

fdalcin:
Ok, this is interesting. Thank you for catching that @richard015ar. I updated the code to fix the unexpected behavior.

richard015ar (approved):
LGTM

PR #4070, opened 2025-05-30
feat: allow allowfullscreen attribute on iframes (resolves #2223)

### LINKED ISSUE #2223: Change iframe parsing rules to permit allow="fullscreen" for embedded videos
Pressbooks users sometimes embed videos and other iframe resources inside of their books. We have strict parsing rules that require the src to be allowlisted: https://github.com/pressbooks/pressbooks/blob/61bcbadcf4716e1a584089e828217e49896e0677/inc/interactive/class-content.php#L145-L168

we also define a set of allowed properties which can be included in the iframe: 
https://github.com/pressbooks/pressbooks/blob/61bcbadcf4716e1a584089e828217e49896e0677/inc/interactive/class-content.php#L232-L241 

Clients would like to be able to allow Kaltura videos embedded via iframe to allow full screen. Here's a sample iframe:
```
<iframe id="kmsembed-1_7aybe0lu" width="608" height="402" src="https://www.macvideo.ca/embed/secure/iframe/entryId/1_7aybe0lu/uiConfId/39241881" class="kmsembed" allowfullscreen webkitallowfullscreen mozAllowFullScreen allow="autoplay *; fullscreen *; encrypted-media *" referrerPolicy="no-referrer-when-downgrade" sandbox="allow-forms allow-same-origin allow-scripts allow-top-navigation allow-pointer-lock allow-popups allow-modals allow-orientation-lock allow-popups-to-escape-sandbox allow-presentation allow-top-navigation-by-user-activation" frameborder="0" title="Kaltura Player"></iframe>
```
When Pressbooks parses this, it removes the allow="fullscreen" and allowfullscreen properties from the iframe, which breaks the player's ability to make the video fullscreen. See example at https://integrations.pressbooks.network/shorttitle/chapter/chapter-1/. Resulting html is `<iframe width="608" height="402" src="https://www.macvideo.ca/embed/secure/iframe/entryId/1_7aybe0lu/uiConfId/39241881" frameborder="0" title="Kaltura Player"></iframe>`

We should let users who have permission to embed iframes for Kaltura videos embed iframes which allow the user to make the video full screen.

## Notes
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy/Using_Feature_Policy#the_iframe_allow_attribute and https://caniuse.com/mdn-html_elements_iframe_allowfullscreen

greatislander:


fdalcin (approved):
LGTM

PR #4073, opened 2025-06-04
fix: conditionally register robots settings

fdalcin:
Issue: pressbooks/private#1745

This PR updates the robots settings to be conditionally registered based on a hook


richard015ar (approved):
LGTM

Issue #3484, opened 2023-10-05
Use constant for DocRaptor pipeline version


SteelWagstaff:
We currently hardcode the DocRaptor version here: https://github.com/pressbooks/pressbooks/blob/9658b9915439a6be68d4df9f753d2e7b67d6816e/inc/modules/export/prince/class-docraptor.php#L103

It would be better to check for a new constant called `DOCRAPTOR_PIPELINE`. If set, use it, if not fall back to hardcoded value.

We could then set this constant in our bedrocks or allow open source users to define it in their wp-config file.

SteelWagstaff:
Still relevant.

cmurtagh:
It makes sense to check for a variable, and default back to a hard-coded value. The crappy part is we have to do that check in the code rather than in the bedrock .env, otherwise it will break OSS users out in the wild.

SteelWagstaff:
@cmurtagh I opened a simple PR that will check for a env variable called 'DOCRAPTOR_PIPELINE' and fall back to a default value if not found: https://github.com/pressbooks/pressbooks/pull/4081. This should let us more easily control pipeline version with env variables without breaking anything for OSS users who don't have these in place.

SteelWagstaff:
Once merged, we may also want to add `DOCRAPTOR_PIPELINE = '10.1'` to the dev and staging branches of our bedrocks so we can test this pipeline (which uses PrinceXML 15.1 instead of 14.3). Let me know if these PRs would be useful?

cmurtagh:
> Once merged, we may also want to add `DOCRAPTOR_PIPELINE = '10.1'` to the dev and staging branches of our bedrocks so we can test this pipeline (which uses PrinceXML 15.1 instead of 14.3). Let me know if these PRs would be useful?

I'm wondering whether we would ever want to be able to configure this at the network layer instead of the bedrock. It just means adding env('DOCRAPTOR_PIPELINE') in the config/application.php or not.

tw77:
Ready for release, has worked as expected on dev

PR #4082, opened 2025-06-10
fix: add support for DocRaptor pipeline env variable to cover generator

SteelWagstaff:
Adds support for setting DocRaptor pipeline version for the cover generator tool. Partial fix for https://github.com/pressbooks/pressbooks/issues/3484.

This will make it easier for us to test updating the pipeline in various environments before bumping in production.


fdalcin (approved):
LGTM

cmurtagh:
Actually, we should roll this back. It should just use the variable, and we run the env() command in config/application.php in the bedrock. The only issue is, what happens if people aren't doing bedrock installs.

PR #4101, opened 2025-06-30
fix: allow spans with lang attributes in section titles

### LINKED ISSUE #493: Make it possible to add language tags in chapter titles
### Feature Description

As a user I would like to be able to assign a different language than the book language to a chapter title or to part of a chapter title.

### Feature Use Case

Many language learning books have chapter titles that are in more than one language. For instance, a German language book for English speakers might have a title like: "Lesson 13: Wetter". When a reader uses a screenreader, that screenreader will output the text in the book's language, unless the text has a language tag. It is possible to assign language tags in chapters, but not in chapter titles. This makes the part of the chapter title that is not in the book's language unintelligible when a screenreader outputs it.

Related: https://github.com/pressbooks/pressbooks/issues/3936

### LINKED ISSUE #3936: A11y Issue: lang Attributes stripped out of chapter titles
### Description

When adding a language tag to words in a chapter title, the `lang` attribute is stripped out upon saving. 

This is an accessibility issue because screen readers use language tags to help direct content to the correct pronunciation engine. Without language tags content can become incomprehensible and inaccessible for people using screen readers. 

### Steps to Reproduce

1. Add a span with language tag to a chapter title, eg: "A Central Asian Trilogy: the Legacy of Aitmatov in Ismailov’s <span lang='kk'><em>Vunderkind Erzhan</em></span>"
2. Save the changes

**Expected behavior:** 
The lang attribute should persist

**Actual behavior:** 
The lang attribute is stripped out

Tagging @angwatters who reported the issue

Related: https://github.com/pressbooks/ideas/issues/493

greatislander:
Resolves #3936, resolves pressbooks/ideas#493.

### DIFF
--- a/inc/sanitize/namespace.php
+++ b/inc/sanitize/namespace.php
@@ -251,6 +251,7 @@ function filter_title( $title ) {
 		'br' => [],
 		'span' => [
 			'class' => [],
+			'lang' => [],
 		],
 		'em' => [],
 		'strong' => [],
--- a/tests/test-sanitize.php
+++ b/tests/test-sanitize.php
@@ -186,6 +186,10 @@ class SanitizeTest extends \WP_UnitTestCase {
 		$test = '<del><strike>Keep me</strike></del>';
 		$test = \Pressbooks\Sanitize\filter_title( $test );
 		$this->assertEquals( '<del>Keep me</del>', $test );
+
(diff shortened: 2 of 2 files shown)


richard015ar (approved):
LGTM

Issue #4119, opened 2025-07-17
PDF Theme Options and some Global Options ignored in PDF Export Files


MichSpieler:
When creating a PDF export (print or digital), newly set PDF Theme Options are being ignored.

First signaled by a network manager, I was able to reproduce the issue on university.pressbooks.pub. https://university.pressbooks.pub/october24testbookchange/

my testing showed the following:

- PDF theme options are not applied in the PDF export file. they are applied and visible in the PDF Preview
- Global theme options are mostly applied in the PDF export file. an example of a setting that is not applied in the PDF export file is chapter numbering. it is applied in the PDF Preview.
- Custom CSS is applied both in the PDF export file and in the PDF Preview

### Steps to Reproduce

1. Go to any book
2. Make changes to the PDF Theme Options and click save
3. Create a PDF export (digital or print)
4. Compare with PDF preview

**Expected behavior:** Both the export file and the PDF preview apply the PDF Theme Options changes.

**Actual behavior:** Only the PDF preview applies the PDF Theme Options changes, the PDF export files do not apply the PDF Theme Options changes.

tw77:
Fix deployed to production networks, client who reported the issue has been notified

PR #4123, opened 2025-07-17
fix: theme options hooks on background

arzola:
See #4119 

This pull request adds functionality to override ebook and PDF stylesheets during the export process in `hooks.php`. 

This is required on background exports to properly attach the Theme Options during exports

Key changes:

* Added a `pre_export` action to apply filters for overriding ebook (`pb_epub_css_override`) and PDF (`pb_pdf_css_override`) stylesheets using the respective `scssOverrides` methods from `EbookOptions` and `PDFOptions` classes.

### How to test

Follow instructions on the ticket options should be displayed on preview and PDFs

### DIFF
--- a/hooks.php
+++ b/hooks.php
@@ -317,8 +317,10 @@ if ( $is_book ) {
 	// Overrides (sometimes a web stylesheet update will be triggered by a visitor so this filter needs to be active outside of the admin)
 	add_filter( 'pb_web_css_override', [ '\Pressbooks\Modules\ThemeOptions\WebOptions', 'scssOverrides' ] );
 	// Overrides for ebook and PDF stylesheets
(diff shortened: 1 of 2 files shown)


richard015ar (approved):
LGTM, Theme options are applied as expected

hooks.php
```diff
@@ -316,6 +316,11 @@
 if ( $is_book ) {
 	// Overrides (sometimes a web stylesheet update will be triggered by a visitor so this filter needs to be active outside of the admin)
 	add_filter( 'pb_web_css_override', [ '\Pressbooks\Modules\ThemeOptions\WebOptions', 'scssOverrides' ] );
+	// Overrides for ebook and PDF stylesheets
+	if ( ! defined( 'WP_TESTS_DOMAIN' ) ) {
```
arzola, line 320: Not happy about this conditional but I think we can later improve this

PR #4130, opened 2025-07-21
feat: header link color in customizer

arzola:
Required customizer variable for https://github.com/pressbooks/pressbooks-aldine/pull/506

This also solves slowness issues on tests


fdalcin (approved):
LGTM

PR #4134, opened 2025-07-22
fix: tablepress not rendering on exports

arzola:
With the new background export processing, we now attach hooks differently. Previously, these hooks were registered only when an export submission occurred. However, since export submission and processing are now decoupled, we need to ensure the hooks are triggered whenever TablePress runs.

This change does not introduce performance overhead, as the hooks will only be triggered when a TablePress table is rendered.

### Note

Tablepress tables on exports appears **unstyled**, this was a previous behavior too I tested it with the previous export system.

### How to test

1. Create a TablePress Table
2. Include it in your chapter to be exported using the shortcode for ie: `[table id=1 /]`
3. Create an export including that chapter
4. An unstyled table should be rendered **without an Edit button** below the content.
5. TablePress **should display an Edit button in webview** below the table

### DIFF
--- a/inc/shortcodes/class-tablepress.php
+++ b/inc/shortcodes/class-tablepress.php
@@ -41,20 +41,22 @@ class TablePress {
 	}
 
 	/**
-	 * Add actions and filters to TablePress to load shortcodes in the admin context.
+	 * Add actions and filters to TablePress to load shortcodes in the admin or exports context.
 	 */
 	public function loadShortcodes() {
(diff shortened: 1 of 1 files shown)


richard015ar (approved):
LGTM and works as expected.
PDF and EPUB versions:
Screenshot 2025-07-22 at 6 14 42 PM
Screenshot 2025-07-22 at 6 14 33 PM

PR #4177, opened 2025-08-20
feat: intgrate editoria11y to tinyMCE editor

richard015ar:
Related issue: https://github.com/pressbooks/private/issues/1806

> [!NOTE]
> This is a POC, we need some additional checks and improvements in the JS, as well as some validation from stakeholders and el viejo.

This PR integrates the existing Editoria11y accessibility checker into the classic TinyMCE editor (non‑Gutenberg) used in Pressbooks. It reuses the upstream editoria11y.min.js without modifying the library, instantiates it inside the editor iframe, and now relies on the library’s own change-watching to keep results up to date automatically.

## Goals

- Reuse upstream Editoria11y assets.
- Scope checking to only the editor content (not the whole admin page).
- Remain resilient to WordPress/TinyMCE timing and iframe loading quirks.
- Keep the integration self‑contained and low risk.

## Core Challenges & Solutions

Challenge | Solution
-- | --
The class `Ed11y` is defined lexically and not exposed on `window`, so the iframe can’t “see” it. | Inject a tiny export shim inside the iframe (and a final fallback in the parent) to set `window.Ed11y = Ed11y` once available.
Race conditions between TinyMCE readiness, library availability, and iframe DOM. | Bounded polling: wait for TinyMCE, then iframe body, then the Ed11y constructor; graceful timeouts at each step.

## Life Cycle Flow

1. Early hook: wrap `tinyMCEPreInit.mceInit.content.setup to attach our setup while preserving original behavior.
2. DOM ready → wait for at least one TinyMCE editor.
3. On TinyMCE `init`:
    - Locate the iframe; ensure `body#tinymce`.
    - Export `Ed11y` into the iframe (shim) and instantiate with tailored options.
4. Final parent-scope safety shim assigns `window.Ed11y` if still not set.

## Why This Approach (Trade-offs)

- Simpler, faster, and more maintainable: defers change detection to the library.
- Keeps the upstream library pristine for easy updates.
- Minimizes Pressbooks-specific code to a small, isolated adapter plus enqueue.


assets/src/scripts/editoria11y-tinymce.js
```diff
@@ -0,0 +1,298 @@
...
+			return done();
+		}
+		const parentScript = document.querySelector( 'script[src*="editoria11y.min.js"]' );
+		if ( ! parentScript ) {
+			return done();
+		}
+		if ( iframeWin.document.querySelector( 'script[src*="editoria11y.min.js"]' ) ) {
+			let loops = 0;
+			( function waitCtor() {
+				try {
```
richard015ar, line 84: Inject a tiny export shim inside the iframe to allow it to get the `Ed11y` object definition within it

assets/src/scripts/editoria11y-tinymce.js
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	/**
+	 * Ensure stylesheet is present in iframe.
+	 *
+	 * @param {Document} iframeDoc
+	 */
+	function ensureIframeStyles( iframeDoc ) {
+		const parentCss = document.querySelector( 'link[href*="editoria11y"][href$=".css"], link[href*="editoria11y.min.css"]' );
+		if ( parentCss && ! iframeDoc.querySelector( 'link[href="' + parentCss.href + '"]' ) ) {
```
richard015ar: mirror the Editoria11y CSS if missing and inject it

assets/src/scripts/editoria11y-tinymce.js
```diff
@@ -0,0 +1,298 @@
...
+			iframeDoc.head.appendChild( l );
+		}
+	}
+
+	/**
+	 * Create Ed11y instance for an editor iframe.
+	 *
+	 * @param {object} editor
+	 */
+	function initEd11yInstance( editor ) {
```
richard015ar: wait for TinyMCE editors, then for iframe body, then for the Ed11y constructor with bounded retry loops and timeouts

assets/src/scripts/editoria11y-tinymce.js
```diff
@@ -0,0 +1,298 @@
...
+		} );
+	}
+
+	/**
+	 * Run full or incremental recheck.
+	 *
+	 * @param {object} editor
+	 * @param {boolean} forceFull
+	 */
+	function runEd11yRecheck( editor, forceFull ) {
```
richard015ar: Track last full scan timestamp. Do checkAll() every 12s (or forced) else attempt incrementalCheck(); fallback to checkAll() if incremental not present

assets/src/scripts/editoria11y-tinymce.js
```diff
@@ -0,0 +1,298 @@
...
+		}
+	}
+
+	/**
+	 * Hook lifecycle for a TinyMCE editor.
+	 *
+	 * @param {object} editor
+	 */
+	function attachEditorLifecycle( editor ) {
+		if ( initializedEditors.has( editor ) ) {
```
richard015ar: Use a WeakSet of initialized editors plus a MutationObserver to reapply hooks only when needed.

assets/src/scripts/editoria11y-tinymce.js
richard015ar: After a conversation with John he suggested I take a look at `editableContents` parameter (see [here](https://github.com/pressbooks/pressbooks/pull/4177)) to allow Ed11y to watch for changes. Try with `mce-content-body` element or `editableContent: 'body'`

itmaybejj:
(also sent in email)

For CKEditor in Drupal, the following are the overridden defaults; some seem relevant to the challenges we discussed this morning. You’ll want to set editableContent as well, as discussed:

{
  "autoDetectShadowComponents": false, // setting this notably improves performance; and these functions are not needed in TinyMCE 
  "inlineAlerts": false,
  "watchForChanges": true,
  "alertMode": "active”, // There’s config to pass here, but the default “assertive" is confusing; I think it is better when editing to start open or stay shut, rather than pop open dynamically on first detected error.
  "editorHeadingLevel": [
    {
      "selector": "form[id^=\"node-\"] #edit-body-wrapper .ck-content",
      "previousHeading": 1
    },
    {
      "selector": "*",
      "previousHeading": 0
    }
  ],
  "cssUrls": [  // this could replace your loader; WordPress attaches the route in my variables array
    "/modules/custom/editoria11y/library/dist/editoria11y.min.css"
  ],
  "currentPage": "/realurl”, // this is one of several [params the WordPress plugin passes](https://github.com/itmaybejj/editoria11y-wp/blob/main/src/functions.php#L173) to allow syncing dismissals; we’ll want to make sure they are all injected into the iframe DOM or exported to it via JS
  "reportsURL": "/admin/reports/editoria11y”  // WordPress  will provide the route
  "syncedDismissals": [],  // the WordPress plugin provides
  "theme": "sleekTheme”,  // the WordPress plugin will provide
  
  "checkRoots": ".gutenberg__editor .is-root-container, [contenteditable=\"true\"]:not(.gutenberg__editor [contenteditable], [contenteditable=\"true\"] [contenteditable])",

...
}

richard015ar:
> (also sent in email)
> 
> For CKEditor in Drupal, the following are the overridden defaults; some seem relevant to the challenges we discussed this morning. You’ll want to set editableContent as well, as discussed:
> 
> { "autoDetectShadowComponents": false, // setting this notably improves performance; and these functions are not needed in TinyMCE "inlineAlerts": false, "watchForChanges": true, "alertMode": "active”, // There’s config to pass here, but the default “assertive" is confusing; I think it is better when editing to start open or stay shut, rather than pop open dynamically on first detected error. "editorHeadingLevel": [ { "selector": "form[id^="node-"] #edit-body-wrapper .ck-content", "previousHeading": 1 }, { "selector": "*", "previousHeading": 0 } ], "cssUrls": [ // this could replace your loader; WordPress attaches the route in my variables array "/modules/custom/editoria11y/library/dist/editoria11y.min.css" ], "currentPage": "/realurl”, // this is one of several [params the WordPress plugin passes](https://github.com/itmaybejj/editoria11y-wp/blob/main/src/functions.php#L173) to allow syncing dismissals; we’ll want to make sure they are all injected into the iframe DOM or exported to it via JS "reportsURL": "/admin/reports/editoria11y” // WordPress will provide the route "syncedDismissals": [], // the WordPress plugin provides "theme": "sleekTheme”, // the WordPress plugin will provide
> 
> "checkRoots": ".gutenberg__editor .is-root-container, [contenteditable="true"]:not(.gutenberg__editor [contenteditable], [contenteditable="true"] [contenteditable])",
> 
> ... }

Thanks @itmaybejj ! I was very useful. I removed a significant portion of the code to use the Ed11y parameters already in place. I did some tests and it seems to work pretty well. 
More improvements to come!

assets/src/scripts/editoria11y-tinymce.js
```diff
@@ -0,0 +1,235 @@
...
+		scriptElement.src = parentScript.src;
+		/**
+		 *
+		 */
+		scriptElement.onload = function () {
+			let postLoadPollCount = 0;
+			( function confirmCtor() {
+				try {
+					if ( ! iframeWindow.Ed11y ) {
+						const shimScriptEl3 = iframeWindow.document.createElement( 'script' );
```
itmaybejj, line 120: Have you worked with [mce_external_plugins](https://developer.wordpress.org/reference/hooks/mce_external_plugins/) before? I'm attempting to inject the assets directly, which should simplify this and remove the race conditions. But if you have already attempted that and it did not work, I'll look more closely at this.

itmaybejj:
@richard015ar I have a proof of concept adaptation of this working in the plugin; I'll polish that up a bit this weekend before sharing. It _should_ import and automate most of your custom code.

My biggest remaining challenge is exactly what I feared -- The inserted Editoria11y elements **_are_** corrupting the user input. The Web components are getting converted to paragraphs (extra carriage returns), and the hidden `div` with inline styles is not being stripped. I think I am going to need to:

1. Experiment with declaring [invalid elements](https://github.com/tinymce/tinymce-docs-4x/blob/main/_includes/configuration/invalid-elements.md) -- **_hopefully_** that will delete the elements rather than convert them to paragraphs.
2. Modify the base library to insert the CSS into the `<head>` or a custom tag other than a `div` that can be stripped, since the invalid elements filter cannot detect attributes.

If that doesn't work we may need to create some sort of post-save filter.
Screenshot 2025-08-27 at 9 00 39 PM

I'm calling it a night; feel free to try to tackle the invalid elements puzzle.

richard015ar:
Great to know you are planning to centralize things in the lib, @itmaybejj ! I think that's going to be very helpful and clean.

Thinking of stripping undesired inserted elements, that's a good catch. I think option 1 would be great; however, option 2 is also something interesting and sounds easier to explore.

Another issue I encountered today. I clicked on the "Code" tab in the TinyMCE option to explore the inserted elements:
Screenshot 2025-08-28 at 9 31 57 AM
Then, when I go back to the "Visual" tab, the Ed11y checker disappears. I have to refresh to page to get it back again. 

Since we work Mon to Thu, and today is a end-of-the-sprint day I don't think I'll be able to explore this challenge till next week. But I'll be on the lookout to see if I can help on Monday afternoon.

itmaybejj:
Ahh; I bet switching to Code view drops and then recreates the DOM, which would destroy the Editoria11y mutation observer and the toggle. I could recreate those, but........The heading structure and alt text visualizers aren't working well either; if we just _drop_ them we'd only really need two buttons -- one to toggle it on and off, and one to step to the next issue. The fastest way forward _might_ be to create those in the TinyMCE toolbar as a simple plugin, and hide the Editoria11y toggle altogether. Then we just need to figure which event comes from a return from code view, and we can reinitialize the mutation observers.

itmaybejj:
OK @richard015ar -- I think I have it working. Could you please clone the branch attached to this [plugin pull request](https://github.com/itmaybejj/editoria11y-wp/pull/43) and see if it works out of the box with no custom code?

I am afraid I was not able to get the dev environment running; I got mostly there but composer couldn't find a required module to finish the install. Lando is up but WordPress fails to bootstrap; I decided not to troubleshoot it further on the clock. But that means I'll need to lean a bit harder on you to test.

This PR scans the page for iframes and (theoretically) checks the content in all of them. I don't have any plugins that let me get more than one on the page, so there's a bit of hand-waving "I think this will work" I am hoping you can help with that.

On my list of things that might still be needed:

- Exposed configuration in the plugin for which Tiny instances to check. I imagine there are some you don't want checked? If not I won't bother with that work.
- Exposed configuration in the plugin for heading roots, if you want to specify which instances start at heading 2 or 3 or don't-flag-the-first-one-at-all. If you don't need it, I won't bother with it.
- My code does not account for scrolling within the iframe. The WP iframes i am looking at all expand the container to fit all of the content, so they never scroll. Assuming that is true of all site configurations may be a mistake, but if that is how your platform also works, I see no reason to revisit it on your dime.

For what it's worth -- the data corruption issue was BAD and turned out to be quite unsolvable with Editoria11y running inside the iframe. I spent Saturday afternoon [rewriting various bits of the library ](https://github.com/itmaybejj/editoria11y/pull/38/files#diff-bd88837547fb1cf8cb34adac30513bc524630788f295bafde7395c8e89074353) to allow it to accept direct node references as check roots, so that I can have the Editoria11y instance _outside_ the editor scan _into_ the iframes, but draw the elements outside the frames. A hat tip to your globalizing Editoria11y code -- I needed that lead to figure out how to pass variables back and forth across the frames.

_I'm going to spend a few more minutes working on some small positioning bugs I still see_ (edit: done) and then leave it alone until we're all back at our desks and can tell if you want me to do any more, or just tag a release. I'm at ~14.5 hours.

richard015ar:
I am closing this in favour of an ongoing built-in integration performed by @itmaybejj in Editoria11y WP plugin (and the main library).

PR #4193, opened 2025-09-03
fix: increase ed11y widget for mobile view

richard015ar:
Issue: https://github.com/pressbooks/private/issues/1807#issuecomment-3250128438

This PR increase the Ed11y widget size in mobile view for easier manipulation.

### Testing notes
- With Editoria11y plugin active to a chapter edition page
- Switch to mobile view and refresh the page 
- Make sure the widget is slightly bigger.


richard015ar:
We decided to hide Ed11y widget for mobile views, so we don't need this anymore.

Issue #4220, opened 2025-09-24
Pressbooks logo in admin top nav bar has incorrect markup


SteelWagstaff:
The admin top nav bar includes a menu with several items. The first item in the menu is a Pressbooks logo that used to be a link but is no longer. It has markup that is confusing for users of assistive technology:

```<div class="ab-item ab-empty-item" role="menuitem"><span class="ab-icon"></span><span class="screen-reader-text">Pressbooks logo</span></div>```

We should either communicate that this is an image with a good accessible name:
```
<div class="ab-item ab-empty-item" role="img" aria-label="Pressbooks logo">
  <span class="ab-icon" aria-hidden="true"></span>
</div>
```
Or simply hide from screen-readers entirely:
```
<div class="ab-item ab-empty-item" aria-hidden="true">
  <span class="ab-icon"></span>
</div>
```

This was reported by Indiana University: https://axe.deque.com/issues/8ecfb259-3791-4140-99bd-c3955ae5f99c

tw77:
Revised markup appears right on dev, ready for release

Issue #4219, opened 2025-09-24
Clone a book page does not indicate which fields are required


SteelWagstaff:
The clone a book page includes two required fields, but it provides no visual indicator to users that these fields are required before form submission.
![Image](https://github.com/user-attachments/assets/4c871dc9-bfb2-48e2-9435-285e008b9647)

We should add aria-required="true" to both inputs and add a * for required along with relevant screen reader text to provide a visual indicator that both fields are required before submission.

Something like:
`<p><span aria-hidden="true">*</span> Required fields</p>`
and
```
<h2>
  <label class="pb-label" for="source-book-url">
    Source Book URL <span aria-hidden="true">*</span>
    <span class="screen-reader-text">required</span>
  </label>
</h2>
```

A similar implementation would be welcome for the second item in https://github.com/pressbooks/pressbooks/issues/4218

tw77:
Fixed on dev, ready for release

Issue #4218, opened 2025-09-24
A11y issues on 'Create a Book' page


SteelWagstaff:
There are three accessibility issues on the 'Create a new book' page:
1. The title element does not identify the contents or purpose of this page (example: https://axe.deque.com/issues/30ead635-549f-4201-8bdb-8757177d8cd5)
2. There are no indicators of which fields are required before user submission (see #4219)
3. If there is an error in user input, the form fields with errors are not identified in text, text alternative, or via programmatic association. The tool tip is not announced in any way. (example: https://axe.deque.com/issues/dad1b448-3415-47c6-ae4a-c8fb18fcf8ec)

These issues were reported by Indiana University

SteelWagstaff:
Rather than fixing on our current page, I recommend making sure we address if/when we implement https://github.com/pressbooks/pressbooks/issues/2856

cmurtagh:
Yeah, I think I like that option too. I'll submit it for next sprint, and fall back to this if we can't do it for some reason.

cmurtagh:
Closing in favour of https://github.com/pressbooks/pressbooks/issues/2856

Issue #2856, opened 2022-06-28
Feature: Allow users to 'create a book' with standalone page


SteelWagstaff:
As a Pressbooks user, I'd like to be able to create a new book from a simple form which resembles the rest of the UI.

At present, new books are created using the wp-signup.php page and make use of the default WordPress site creation process. Ideally, we'd create a standalone page used for book creation, and use this page to allow users to create new books, similar to how we handle cloning books. By default 'lowly users' should be able to create new books with this form, unless network managers disable the 'Allow registered users to create and clone new books' option in pressbooks-network-analytics 
![Screenshot from 2022-06-28 16-13-56](https://user-images.githubusercontent.com/13485451/176319456-38026e62-5d04-4b0d-aa7f-6686b2de072f.png)
(compare with the 'Allow new registrations' option WordPress' default 'Registration Settings')
![Screenshot from 2022-06-28 16-13-37](https://user-images.githubusercontent.com/13485451/176319459-c705b62b-d454-4418-b12b-eb73fc4110ae.png)

SteelWagstaff:
FYI @MichelleWeremczuk

MichelleWeremczuk:
Sounds good to me. I’m not sure what additional UX/UI is needed here since it seems we’d just be recreating this page. I assume we just need to refine/remove some content that won't be relevant. 
![screencapture-university-pressbooks-pub-wp-signup-php-2025-01-07-09_19_03](https://github.com/user-attachments/assets/cb97954c-4071-4a8b-95c9-24648440163a)

cmurtagh:
If we do this, it should also resolve https://github.com/pressbooks/pressbooks/issues/4218. I think this is the better option.

arzola:
This issue has been automatically closed as part of an organization-wide backlog review.

**Reason:** This issue was created 3y 8m ago with no updates in the last 5m.

If this is still relevant, please feel free to reopen it with updated context. We appreciate your contributions and want to ensure our issue tracker stays actionable.

**Learn more:** https://pressbooks.community/t/announcement-automatic-closure-of-old-github-tickets/4542

PR #4221, opened 2025-09-24
fix: a11y issues for PB logo and clone a book page

SteelWagstaff:
Improve markup and styles for PB logo in admin nav top bar. Removes misleading role via JS and stops logo from changing state upon hover. Addresses https://github.com/pressbooks/pressbooks/issues/4220

Improve markup for clone a book page by adding clear visual indicators to help users know which form fields are required. Addresses https://github.com/pressbooks/pressbooks/issues/4219


assets/src/scripts/a11y.js
```diff
@@ -81,3 +81,16 @@ document.addEventListener( 'DOMContentLoaded', function () {
 		} );
 	} );
 } );
+
+const pbLogo = document.querySelector('#wp-admin-bar-pb-logo > .ab-item');
+if (pbLogo) {
+	pbLogo.removeAttribute('role');
```
SteelWagstaff, line 87: Removes the confusing role attribute (resolves #4220)

assets/src/styles/admin.scss
```diff
@@ -43,6 +43,18 @@ html #wpadminbar {
   height: var(--pb-toolbar-item-height);
 }
 
+#wp-admin-bar-pb-logo > .ab-item {
+  pointer-events: none;
```
SteelWagstaff, line 47: Adjusts CSS styling so that the logo no longer changes colour when the list item containing it receives focus.

SteelWagstaff:
@cmurtagh this should be ready for review by Ned or one of the devs on our team. I didn't attempt to fix the issues described in #4218 -- our solution for that page is really hacky. We probably should consider making our own 'create a book' page in place of the strange overrides we're making to the default wp-signup.php page. See https://github.com/pressbooks/pressbooks/issues/2856 @greatislander probably has some good ideas for this.

cmurtagh (commented):
Looks good to me. I'll ask a dev for a quick review too.

fdalcin (approved):
LGTM

Issue #4226, opened 2018-01-02
Paragraphs: change default hyphenation setting to "on"


greatislander:

tw77:
Setting enabled by default for new books on dev.pressbooks.network, ready to release

PR #4227, opened 2025-10-03
fix: change default pdf_hyphens value

SteelWagstaff:
Change default PDF hyphenation value to true in PDF theme options. Fix for https://github.com/pressbooks/pressbooks/issues/4226


fdalcin (approved):
LGTM

PR #4231, opened 2025-10-03
feat: add Contributors as an allowed front matter type

### LINKED ISSUE #3756: Allow users to create 'contributors' page as a front matter type
As an author, I want to create the dynamically generated 'contributors' page for my book, but I want to include it as front matter rather than back matter.

When we built the feature, we only built support for ‘contributors’ as a back matter type (I think because that’s where Chicago Manual of Style said contributors were supposed to be displayed): #2336. Some authors would like to use the existing feature, with the only difference being they'd like to page to live in their front matter. Initial request from the University of Oregon here: https://pressbooks.community/t/using-back-matter-contributors-page-type-as-a-front-matter-element/2826

I've looked up the reference for Chicago Manual of Style and it says:
> A list of contributors may be appropriate for a work by many authors in which only the volume editor’s name appears on the title page. The list (usually headed Contributors) may appear in the front matter, but the preferred location is in the back matter, immediately be­fore the index.

It would be nice if authors could choose to place this in either the front or back matter.

SteelWagstaff:
Add Contributors as a permitted front matter type. Resolves #3756

To test:
1. Create a new front matter and give it a title
2. Select the front matter type 'Contributors' from the drop down list and click 'update'
3. Observe the message 'To display a list of contributors, leave the content blank.' at the top of the page
4. Confirm that when the content is left blank, the full list of contributors is displayed on the published front matter content page
5. Confirm that the existing Contributors back matter type (and Glossary type) are unaffected by this change.


fdalcin (approved):
LGTM

Issue #3818, opened 2024-10-22
Missing aria-labels in text editor toolbar


MichSpieler:
Three buttons on the Text Editor Toolbar (`close tags`, `fn` and `Distraction free writing mode`) do not have aria-labels (all other buttons do). `close tags` and `Distraction free writing mode` have a title instead. `fn` does have neither a title nor an aria-label. 

`close tags` should have an `aria-label="Close tags"` (or Close all open tags) and `fn` should have an `aria-label="Footnote"`.

Distraction free writing mode does have an aria-label in the Visual editor (`Distraction-free writing mode (^⌥W)`).

![image](https://github.com/user-attachments/assets/55f59d94-77a5-408b-8922-f4bc8ebd27c0)

cmurtagh:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pressbooks-Product-Backlog-60ae7d807f6e370010b3d5be/issues/pressbooks/pressbooks/3818?planning-poker) with Zenhub @arzola @fdalcin @MichelleWeremczuk @richard015ar @tw77

SteelWagstaff:
@cmurtagh @arzola I've added a PR to address this: #4233

tw77:
aria-labels added as needed on dev (state change / changing aria-label in place for footnote button too), ready for release

PR #4233, opened 2025-10-04
fix: add aria-labels to code editor buttons which lack them

SteelWagstaff:
Add aria-labels to the footnote, close tags, and distraction free mode buttons in the Code (text) editor. Ensure that the aria-label changes when the footnote button state changes. Fix for #3818

To test:
1. Open the Code/text editor and inspect the markup for buttons. Observe that aria-labels have been added for the three relevant buttons
2. Click the footnote button. Observe the state change and updated aria-label
3. Click the footnote button again & observe the state change & updated aria-label.


fdalcin (approved):
LGTM. Tested and works as expected locally.

PR #4250, opened 2025-10-17
fix: HTML decoding error when cloning Custom CSS within a network (resolves #4209)

### LINKED ISSUE #4209: '>' in Custom CSS transformed into '&gt;' when cloning the book which causes CSS to break
Cloning the book "[Introduction to Criminal Justice](https://openwa.pressbooks.pub/crimjust1/)" onto a [different network](https://uml.pressbooks.pub/criminaljusticeunlocked) has caused issues with the Custom CSS.

We've seen the same thing happening when the book "[Introduction to Welding](https://openwa.pressbooks.pub/welding1)" was cloned within the same network.

All `>` characters in the Custom CSS for Web panel were transformed into `&gt` which leads to CSS compilation errors.

Image

Image

Cloning the book to [our Demo network](https://university.pressbooks.pub/crimjustclonecss) did not produce the same issue.

SteelWagstaff:
Decode HTML entities when performing in network clones. Fix for https://github.com/pressbooks/pressbooks/issues/4209. I don't know what code coverage for the patch is at 0%, especially as I added new tests specifically to test that `>` are properly decoded for both types of clone (internal and external).

### DIFF
--- a/inc/cloner/class-cloner.php
+++ b/inc/cloner/class-cloner.php
@@ -1456,9 +1456,17 @@ class Cloner {
 		foreach ( $styles_container->getSupported() as $slug => $style_type ) {
 			if ( isset( $this->sourceStyles[ $slug ] ) ) {
 				$post = $styles_container->getPost( $slug );
+				$content = $this->sourceStyles[ $slug ];
+
+				// For in-network clones, the REST API may have already encoded HTML entities
+				// Decode them to prevent double-encoding when wp_update_post() applies its own sanitization
+				if ( ! empty( $this->sourceBookId ) ) {
+					$content = html_entity_decode( $content, ENT_QUOTES | ENT_HTML5, 'UTF-8' );
+				}
+
 				$post_params = [
 					'ID' => $post->ID,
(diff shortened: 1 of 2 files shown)


cmurtagh (approved):

richard015ar:
I cannot reproduce the bug to test this solution. My test cases:
I tried with the follosing custom style rule:
```css
.front-matter h1 > .something {
  color: blue;
}
```
Screenshot 2025-10-28 at 1 07 39 PM

- Internal cloning worked as expected; the cloned book has the expected custom css, no error was displayed.
- External cloning also worked. The cloned book is [here](https://dev.pressbooks.network/customcss1/wp-admin/themes.php?page=pb_custom_styles).

richard015ar (approved):
This change doesn't seem to be a problem, I tested it locally and new problem was identified. I couldn't replicate the bug that this solutoins intends to fix, even with the books reported in the related issue. I'll merge this, and hopefully it fixes it. Futher testing is recommended to confirm this changes doesn't cause any problem in the cloning routine (which I don't think it would).

PR #4251, opened 2025-10-17
fix: improve redirect logic after login (resolves #1794)

### LINKED ISSUE #1794: Improve redirection logic after login
**Expected:** 
After login from network root page, the user sees their dashboard page (regular dashboard if regular user, network manager dashboard if they're a network manager).
After login from a particular book, the user sees the book dashboard page, unless they have no role in the book or are a subscriber in that book, in which case they're taken back to the page they logged in from.

SteelWagstaff:
Improve redirection logic. Fix for https://github.com/pressbooks/pressbooks/issues/1794. Desired logic:
1. When a user logs in from within a book, return them to the page they logged in from if they don't have a role in the book, or have the role of subscriber
2. If the user has any other role in the book (or is a super admin), return them to the book dashboard
3. If the user logs in from the root site, check if they are a super admin. if so, return the user to the administer network dashboard
If not, return them to the standard user dashboard.


inc/redirect/namespace.php
```diff
@@ -496,13 +497,53 @@ function handle_dashboard_redirect( string $redirect_to, string $requested_redir
...
 	}
 
-	if ( ! is_super_admin( $user->ID ) ) {
+	// Only redirect if user is being redirected to admin areas, dashboard pages, or in book context where we want to customize behavior
+	$is_admin_redirect = (
+		$redirect_to === admin_url() ||
+		str_contains( $redirect_to, '/wp-admin/' ) ||
+		str_contains( $redirect_to, 'page=pb_home_page' ) ||
+		str_contains( $redirect_to, 'page=pb_network_page' ) ||
+		str_contains( $redirect_to, 'page=book_dashboard' ) ||
```
richard015ar: `str_contains` could bring false positives if the URL contains those substrings but they aren't from the admin area.  We could use `wp_parse_url` here, and maybe create a new method to avoid this long check within this already long method:
```php
function is_admin_area_redirect( string $redirect_to, array $admin_pages ): bool {
    $parsed = wp_parse_url( $redirect_to );
    $path   = isset( $parsed['path'] ) ? $parsed['path'] : '';
    $query  = isset( $parsed['query'] ) ? $parsed['query'] : '';

    $is_wp_admin_path = preg_match( '#^/wp-admin(?:/|$)#', $path ) === 1;

    parse_str( $query, $qs );
    $is_special_page = isset( $qs['page'] ) && in_array( $qs['page'], $admin_pages, true );

    $admin_path = wp_parse_url( admin_url(), PHP_URL_PATH );
    $is_admin_root = $path === $admin_path || rtrim( $redirect_to, '/' ) === rtrim( admin_url(), '/' );
 
    $is_same_as_requested = rtrim( $redirect_to, '/' ) === rtrim( $requested_redirect_to, '/' );

    return $is_wp_admin_path || $is_special_page || $is_admin_root || $is_same_as_requested;
}

// we could use it as:
is_admin_area_redirect( $redirect_to, [ 'pb_home_page', 'pb_network_page', 'book_dashboard' ] );
```

inc/redirect/namespace.php
```diff
@@ -496,13 +497,53 @@ function handle_dashboard_redirect( string $redirect_to, string $requested_redir
...
+			return admin_url( 'index.php?page=book_dashboard' );
+		} else {
+			// Check regular user's role in this book
+			$user_roles_in_book = $user->roles; // This gives roles for current blog context
+
+			// If user has no role in book, or only subscriber role, return to original page
+			if ( empty( $user_roles_in_book ) || ( count( $user_roles_in_book ) === 1 && $user_roles_in_book[0] === 'subscriber' ) ) {
+				// Users with no role or only subscriber role go back to the page they logged in from
+				if ( $requested_redirect_to && $requested_redirect_to !== admin_url() ) {
+					return $requested_redirect_to;
```
richard015ar: Here maybe we could use [wp_validate_redirect](https://developer.wordpress.org/reference/functions/wp_validate_redirect/), the url can come from the user.

richard015ar (dismissed):

richard015ar:
I am working in some improvements. I'll push them as soon as I get answer from the question I asked earlier.

richard015ar:
Some how the question I left here in a previous comment dissapeared.

With the Pressbooks Shared Network enabled, and as a IM, I:
- Log in from a book page where I have no role, and I am redirected to the Book Dashboard page with a message similar to: "You don't have permissions to access to this page".
- Log in from a network page, and I am redirected to the IM network dashboard.

I guess the first case is incorrect and we should redirect the IM to the book page I was trying to navigate to? and if the NM has some role in the book maybe we should redirect to the book dashboard as it is now (except subscriber?) and for the second case, it is ok.
I want to confirm this with @cmurtagh

cmurtagh:
> I guess the first case is incorrect and we should redirect the IM to the book page I was trying to navigate to? and if the NM has some role in the book maybe we should redirect to the book dashboard as it is now (except subscriber?) and for the second case, it is ok. I want to confirm this with @cmurtagh

We talked about this in standup, and we're not sure what the right behaviour should be. My intuition is that if you logged in from a location and you have rights to it, you should return to that location after logging in. Only in contexts where you will be given a "you do not have rights here" type of message, should you be redirected to somewhere else.

richard015ar:
We'll put this on hold, till we discuss it in the SC.

tw77:
@MichelleWeremczuk the desired logic @SteelWagstaff outlines in the initial summary here makes sense, imo. To summarize my understanding of it (and then I'll reply to Ricardo's comment above):

1. Logging in from the network home page / root site is always done with the intention of reaching the dashboard appropriate for a given user. (There isn't really a reason to be returned to the root site here; being returned would just introduce an extra step as the user would then have to click 'Admin' to reach the dashboard as intended)

2. When a Subscriber user or a user with no role in a book logs into a book, it makes sense to return them to the same book page they were on, rather than bringing them to a book dashboard that they can't meaningfully interact with. (There are many books where Subscriber users need to sign in to read private content, and this is best facilitated by keeping them in the book upon login)

3. When a user with a higher-than-Subscriber role in a book chooses to login from that book, the user has edit permissions and is likely to be logging in with the intention of reaching the book dashboard and editing from there.

re: what @richard015ar notes:

> With the Pressbooks Shared Network enabled, and as a IM, I:
> 
> Log in from a book page where I have no role, and I am redirected to the Book Dashboard page with a message similar to: "You don't have permissions to access to this page".
> Log in from a network page, and I am redirected to the IM network dashboard.
> I guess the first case is incorrect

The second case (login from a network page) seems right to me.

When I think about it more, the first case seems fine to me too. If an IM is attempting to login to the dashboard of an unassigned book or a book from another institution, the 'You don't have permissions' message is the appropriate one. This is very much an edge case, but it seems unlikely that an IM would be logging into another institution's book solely with the intention of reading it. Even if we were to modify the logic to return the IM-from-another-institution to the book upon login, they would still encounter the 'You don't have permissions' message upon clicking the 'Admin' link in the top-right of the webbook interface.

fdalcin (approved):
LGTM. Works as expected locally.

PR #4252, opened 2025-10-17
fix: improve file export time precision (resolves #1869)

### LINKED ISSUE #1869: Newly exported files don't always display at the top of the export table by default
### Description

By default, files in the export table should be arranged by the timestamp of the export, descending in order of recency. Users can _then_ change the export table to arrange files by other qualities like file name, format, size, or pinned status, or reverse the order of the time stamps. 

On some occasions, the most recently exported file will not be the file at the top of the table. Generally this occurs when files are exported not at the same time but fairly closely together. See the image below of the export table with none of its defaults changed: 

Screen Shot 2019-12-04 at 1 55 33 PM

Notice the export numbers. We can tell that 15754855**41**  is the newest only because it has the largest number in order of export sequence, followed by 15754855**32** and 15754855**23**. These files are displaying in the table in the reverse order of what they should be. 

### Steps to Reproduce

1. Go to the **Export** page of a book in Pressbooks
2. Export a file
3. Export the file again
4. Export the file again
5. Check to make sure that the most recent file is at the top of the export page. 

**Expected behavior:** All export files are ordered by recency of export by default. 

**Actual behavior:** Files may not appear in reverse order under some circumstances.

SteelWagstaff:
Ensure that sort order by date exported works correctly for file exports by increasing precision to include seconds. Addresses https://github.com/pressbooks/pressbooks/issues/1869

### DIFF
--- a/inc/modules/export/class-table.php
+++ b/inc/modules/export/class-table.php
@@ -344,7 +344,7 @@ class Table extends \WP_List_Table {
 				'format' => $this->getFormat( $file ),
 				'size' => $stat['size'],
 				'pin' => $this->hasPin( $id ) ? 1 : 0,
-				'exported' => date_i18n( 'Y-m-d H:i', $stat['mtime'] ),
+				'exported' => date_i18n( 'Y-m-d H:i:s', $stat['mtime'] ),
 			];
 		}
 		return $files;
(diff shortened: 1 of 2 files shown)


fdalcin (approved):
LGTM

PR #4253, opened 2025-10-18
fix: hide 'add new book information' button (resolves #3964)

### LINKED ISSUE #3964: Remove unnecessary Add New Book Information button
The Book Info page contains an  **Add New Book Information** button with the `class="page-title-action"` to the right of the **Edit Book Information** `<h1>` text:

Image

Clicking this **Add New Book Information** button loads a new Book Info page with all fields cleared. Changes saved from here override fields that were filled out in the previous Book Info page. The previous Book Info page becomes unreachable at this point.

Pressbooks only supports one Book Info page per book, with no way for a book admin to navigate between different Book Info pages, so the existence of an **Add New Book Information** button is confusing and pointless. One can easily add or edit new information from the default Book Info page without using this button and without clearing existing book metadata.

The inclusion of this button seems unintentional since an equivalent, but actually useful button exists in the Edit page for front / back matter, a chapter, or a Part (Book Info is also a type of post). It would be good to remove this **Add New...** button in the context of the Book Info page without getting rid of the equivalent button in other Edit pages.

SteelWagstaff:
Add more precise selector to hide 'Add New Book Information' button. Create redirect rule in case a user manually accesses the link on accident. Fix for #3964


SteelWagstaff:
@arzola @richard015ar this is ready for review. The failing test on 8.4 is not related to any changes I made and happens for all PRs opened against the new test suite matrix. I don't know why codecov thinks this doesn't have tests -- I did write a test for the new function.

inc/admin/laf/namespace.php
```diff
@@ -1864,3 +1864,19 @@ function filter_media_list_for_contributors( $query ) {
...
+/**
+ * Redirect users from the Add Book Information page to the allowed book info page
+ */
+function block_metadata_add_new_page() {
+	global $pagenow;
+
+	if ( 'post-new.php' === $pagenow && isset( $_GET['post_type'] ) && 'metadata' === $_GET['post_type'] ) {
+
+		// Redirect to the allowed book info page instead
+		wp_redirect( admin_url( 'post.php?post=12&action=edit' ) );
```
richard015ar: @SteelWagstaff why `post=12`? Is this a fixed Post ID?
I tried with my local network, which is a DB copy of UEN network, and the `<ADMIN_URL>/post.php?post=12&action=edit` URL shows me:
Screenshot 2025-11-19 at 10 27 23 AM

richard015ar (changes requested):

inc/admin/laf/namespace.php
richard015ar: I see what's happening. I was on the main site instead of a particular book.
I think we could use a more reliable way to ensure the post ID. I looked how we do it in other parts of PB plugin, and we use the `Pressbooks\Metadata` class:
```php
$metadataPostId = ( new Metadata )->getMetaPostId();

wp_redirect( admin_url( "post.php?post={$metadataPostId}&action=edit" ) );
```

SteelWagstaff: Nice catch. Just pushed changes to use this approach.

SteelWagstaff:
@richard015ar made a change addressing your concern. ready for re-review

hooks-admin.php
```diff
@@ -110,6 +110,7 @@
 add_filter( 'ajax_query_attachments_args', '\Pressbooks\Admin\Laf\filter_media_for_contributors' );
 add_action( 'pre_get_posts', '\Pressbooks\Admin\Laf\filter_media_list_for_contributors' );
 add_action( 'init', '\Pressbooks\Admin\Laf\enable_media_buttons_for_contributors' );
+add_action( 'admin_init', '\Pressbooks\Admin\Laf\block_metadata_add_new_page' );
```
richard015ar: It is not mandatory, but it is more of an idea to see what you think.
We could use a more specific hook instead `admin_init`. I was looking an I think `load-post-new.php ` could make it.
```php
add_action( 'load-post-new.php', '\Pressbooks\Admin\Laf\block_metadata_add_new_page' );
```

And then in the function we shouldn't need to check the global $page:
```php
function block_metadata_add_new_page() {
	if ( ! isset( $_GET['post_type'] ) || 'metadata' !== $_GET['post_type'] ) {
        return;
    }

   $metadata_post_id = ( new Metadata )->getMetaPostId();
    if ( ! $metadata_post_id ) {
        return;
    }

    wp_safe_redirect( admin_url( 'post.php?post=' . intval( $metadata_post_id ) . '&action=edit' ) );
    exit;
}
```

richard015ar:
@SteelWagstaff What do you think about the idea I shared?

SteelWagstaff:
@richard015ar yeah that makes sense to me. Right now we don't have any redirect rule in place, so anything would be better. I don't it's very likely to ever be triggered, since a user would have to literally type the URL in to try to load the page we're suppressing (we're hiding the button that has the link), but happy to implement whatever you think is best. Do you want me to make that change or will you send a friendly commit?

SteelWagstaff:
@richard015ar I pushed changes similar to what you proposed. Adding the action on load-post-new is a good idea. I added early returns to the redirect function and decided to use wp_safe_redirect and absint (probably not necessary, but just in case meta post ID returned a weird value).

richard015ar (approved):
LGTM. I just saw your notes, but I appreciate the change! Thank you @SteelWagstaff

PR #4256, opened 2025-10-22
fix: remove privacy option on signup

arzola:
This pull request introduces a new privacy-related feature that removes the public/private site option from the WordPress signup page, ensuring users cannot select site visibility during signup. The main logic is implemented in the `Privacy` class and is now covered by a dedicated test.

Privacy option removal from signup page:

* Added the `removePublicOptionFromSignup` method to the `Privacy` class, which uses output buffering and a regex to remove the privacy option HTML from the signup page (`wp-signup.php`).
* Called `removePublicOptionFromSignup` inside the `hooks` method to ensure it's triggered during initialization.

Testing:

* Added a new test `it_removes_privacy_options_from_signup` in `tests/test-privacy.php` to verify that the privacy options are removed from the signup page output.

### DIFF
--- a/inc/class-privacy.php
+++ b/inc/class-privacy.php
@@ -31,6 +31,7 @@ class Privacy {
 	 */
 	static public function hooks( Privacy $obj ) {
 		add_filter( 'schedule_event', [ $obj, 'reschedulePrivacyCron' ] );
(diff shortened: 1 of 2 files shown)


richard015ar (approved):
LGTM

Issue #4263, opened 2025-10-27
Ensure that language tags are available in PDF export (digital)


MichSpieler:
A [forum user has reported](https://pressbooks.community/t/language-tags-missing-from-pdf-exports/4456/2) that language tags don't appear consistently in PDF exports.

> We are using the Prince pdf export routine, and, while producing a series of books with a mix of languages, our digital publishing specialist discovered that language tags from the web book do not appear in the pdf exports. These are a required accessibility feature, and because there are thousands of them in these texts, this is not a remediation that we can manage by hand.

> I notice that, oddly, footnotes with `<span lang=”fr”>` are added in the pdf correctly (e.g., the span shows up as a tag and that tag has the language as a property) Pressbooks uses the XHTML for the source document for Prince. I notice that in the XHTML document, footnotes have `<span lang=”fr”>` while other places in the document have `<span xml:lang=”fr”>` or `<p xml:lang=”fr”>` and, for some reason, Prince is not recognizing the xml:lang property but is recognizing the lang property.
>
>After consulting the Prince documentation, I get a few leads. To cover my bases, I add @namespace xml “``http://www.w3.org/XML/1998/namespace”``; to the pdf css, with no effect, and then tell prince that the source of the digital pdf is xml by adding $prince->setInputType(‘xml’); to the Prince wrapper for digital pdfs. This does cause the PDF to generate with the appropriate language property in all of the places I had `<span lang=”fr”>` as well as other places like `<p xml:lang=”fr”>` that were rendering as xml:lang in the xhtml.

When reproducing this issue I found the following in the PDF preview ([PDF preview of chapter on staging](https://staging.pressbooks.network/bugreprobook/format/xhtml?debug=prince#chapter-language-tags-in-pdf-en-francais))
- Language tags in chapter titles are present in the format `<span lang="&quot;fr&quot;" data-ref="8cb37424-3b45-4775-a7eb-f76b904d276f">en français</span>` (there's an additional set of "")
- Language tags in regular text are present in the format `<span xml:lang="fr" data-ref="99ea880a-d58e-4896-9ffb-8f688cfb1992">je change pour le français</span>`
- Language tags in footnotes are present in the format `<span lang="fr" data-ref="0936044d-ca1e-4a67-8bd4-ed044d07eebb">je change pour le français</span>`
[...]

PR #4265, opened 2025-10-28
fix: pass input type to Prince

alexdryden:
This addresses #4263 by adding parsing for xml:lang properties in Prince pdf exports. Because Prince can't infer the file type by the extension, it needs to be told explicitly that this is xml, otherwise it won't parse xml:* attributes, like xml:lang.

### DIFF
--- a/inc/modules/export/prince/class-pdf.php
+++ b/inc/modules/export/prince/class-pdf.php
@@ -254,6 +254,7 @@ class Pdf extends Export {
 		$prince->setHTML( true );
 		$prince->setCompress( true );
 		$prince->setHttpTimeout( defined( 'WP_TESTS_MULTISITE' ) ? 5 : 600 ); // 5 seconds for tests, 10 minutes for production
+		$prince->setInputType('xml');
 		if ( defined( 'WP_ENV' ) && ( WP_ENV === 'development' ) ) {
 			$prince->setInsecure( true );
 		}
@@ -265,6 +266,7 @@ class Pdf extends Export {
 			$prince->setPDFOutputIntent( $this->pdfOutputIntent );
 		} elseif ( stripos( get_class( $this ), 'print' ) === false && empty( $this->pdfProfile ) ) {
 			$prince->setPDFProfile( 'PDF/UA-1' );
+			
 		}
 
 		yield 60 => __( 'Adding stylesheets and scripts...', 'pressbooks' );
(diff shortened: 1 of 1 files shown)


arzola:
@alexdryden thank you for your contribution, could you run please `composer fix` on your branch to fix lint issues?

alexdryden:
@arzola composer fix reports no fixable errors were found.

alexdryden:
converting to draft to address failing test without triggering gh action

alexdryden:
@arzola I'm looking at the failure on  Modules_Export_ExportTest::test_sanityChecks, and the errors that Prince is putting out sound reasonable. It looks like the test is using an html output as the input for the pdf conversion, but not something that was generated live from the application according to the pattern the application uses, i.e. the xhtml version of a book. I've not dug too deep into how this mock was created, but it looks like the test document used for the Prince input needs to be updated to match the behavior of the application for the test to pass. (When I sub in an xhtml export for the input, the test passes.) For an example of the failures Prince is reporting on the failing test, I'm getting:
```
 [0] => Array
        (
            [0] => err
            [1] => /tmp/wordpress//wp-content/uploads/sites/3/pressbooks/exports/Site-0000024-1765391184.html:143
            [2] => Opening and ending tag mismatch: img line 143 and div
        )

    [1] => Array
        (
            [0] => err
            [1] => /tmp/wordpress//wp-content/uploads/sites/3/pressbooks/exports/Site-0000024-1765391184.html:143
            [2] => Opening and ending tag mismatch: br line 143 and p
        )

    [2] => Array
        (
            [0] => err
            [1] => /tmp/wordpress//wp-content/uploads/sites/3/pressbooks/exports/Site-0000024-1765391184.html:143
            [2] => Opening and ending tag mismatch: p line 143 and div
        )

    [3] => Array
        (
            [0] => err
            [1] => /tmp/wordpress//wp-content/uploads/sites/3/pressbooks/exports/Site-0000024-1765391184.html:185
            [2] => Opening and ending tag mismatch: img line 185 and div
        )
```
For most of these errors, the test document uses img or br tags that aren't self closing, but the application uses self closing tags for the prince input (i.e., via my-pb-instance.example.com/my-book/format/xhtml?timestamp=<time_stamp>&hashkey=<hash>).

So, in this case Prince is correctly reporting errors you would expect to see from an xml parser given the input and then failing. I'd be happy to update the test to support this specific change, but given the TODOs in the test, this might be a case where the tests need broader attention than this one issue. Let me know!

alexdryden:
A few observations about the failing test:

This is a little hard to debug because the $xhtml_path passed to the exporter is set and overwritten inside the loop, and its value will depend on the order of the modules. I'm seeing a few possible states: If PDF is first, then its value will be null. If it is second and XHTML/XHTML11 instance comes before it, it will be the value of the Xhtml11 exporter's `$exporter->getOutputPath()` method. 

It seems like it really *has* to be the xhtml output from the book. So, if there is an explicit way to do that, I think that would be the clearest. Before hitting that loop, the value is, http://example.org/testpath0000024/format/xhtml?timestamp=<timestamp>&hashkey=<hash>. 

So, I think the solution is either to ship a sample xhtml document to test against, or to more reliably get the xhtml from book that is being mocked up.

PR #4273, opened 2025-11-06
fix: disable wptexturize on mathjax pages

arzola:
This pull request updates the MathJax integration to ensure that the `wptexturize` filter is removed from post content when MathJax scripts are loaded. This prevents unwanted text transformations that could interfere with LaTeX/math rendering. Corresponding tests have also been added to verify this behavior.

**MathJax Integration Improvements:**

* Updated the `addScripts` method in `class-mathjax.php` to remove the `wptexturize` filter from `the_content` when MathJax is enabled, preventing unwanted text transformations in math content.

**Testing Enhancements:**

* Modified the `testAddHeaders` test to call `addScripts`, and added assertions to verify that `wptexturize` is no longer attached to `the_content` when MathJax is active.
* Minor cleanup in `testSectionHasMath` by removing an unnecessary blank line.


fdalcin (approved):
LGTM

Issue #2431, opened 2021-10-12
Table header cells created as `<td>` elements


SteelWagstaff:
### Description

When a user creates a new table in Pressbooks and designates a row as the table header, the cells within that row remain `<td>` elements. They should be `<th>`. See https://www.w3.org/MarkUp/html3/tablecells.html#:~:text=TH%20is%20used%20for%20table,heavier%20font%20for%20header%20cells.&text=The%20CLASS%20attribute%20can%20be,instance%20into%20heads%20and%20subheads.

### Steps to Reproduce

1. Create a new table in your book
2. Select the top row and designate it as the table header
3. View the HTML output for your book and observe that cells inside the `<thead>` element are rendered as `<td>` elements rather than the preferred `<th>` elements.

https://user-images.githubusercontent.com/13485451/137010335-0e50c881-6072-43ad-a1dc-d10467551c67.mp4

**Expected behavior:** 
When a user designates a row as a table header, the cells in that row should be rendered as `<th>` elements.

**Actual behavior:** 
When a user designates a row as a table header, the cells in that row are rendered as `<td>` elements.

## Notes
We use the build in TinyMCE table editor (with some customizations). See https://github.com/pressbooks/pressbooks/blob/16e57f7f97e88ba5c74b2121d7e07960b14cbe10/inc/editor/namespace.php#L305-L395

SteelWagstaff:
Check https://www.tiny.cloud/docs/plugins/opensource/table/ and do further research.

SteelWagstaff:
It appears that accessible headings can be created using TinyMCE 5.4 and later: https://www.tiny.cloud/docs/plugins/opensource/table/#table_header_type

SteelWagstaff:
[...]

PR #4304, opened 2025-12-11
fix: convert all td elements to th in thead elements before save

SteelWagstaff:
Automatically convert `<td>` elements within table header rows to `<th>`. Fix for  https://github.com/pressbooks/pressbooks/issues/2431. 

Previously, when users created tables with header rows using the TinyMCE editor, the cells were incorrectly created as `<td>` elements instead of the semantically correct `<th>` elements. This PR checks the content before save and safely converts td elements inside of thead rows to the correct th element. It will remove the Editoria11y checks/flags for malformed thead content upon content save for preexisting and new tables alike.

## How It Works
When a user saves a post with a table that has a header row:

1. TinyMCE creates the table with `<thead>` but uses `<td>` elements
2. WordPress fires the content_save_pre filter before saving
3. The newly created function processes the HTML and converts `<td>` cells within `<thead>` to `<th>` elements.
4. The corrected HTML is saved to the database

## Changes Made
1. Added a new fix_table_header_cells() function to the Presbooks/Editor namespace

This function processes HTML content using DOMDocument to find all `<thead>` elements. It converts all `<td>` elements within `<thead>` to `<th>` elements and preserves all attributes (class, colspan, rowspan, etc.).

It only processes content when `<thead>` is present (performance optimization) and leaves `<tbody>` cells untouched.

2. Hooked the function to content_save_pre filter (hooks-admin.php)

This filter runs when content is saved, automatically converting the HTML before it's stored in the database. 

3. Added tests.


fdalcin (approved):
LGTM!

PR #4322, opened 2026-01-13
feat: add filter to adjust book menu order

SteelWagstaff:
Add filter to allow book menu order to be altered and styles for 'disabled' menu links. Accompanies https://github.com/pressbooks/pressbooks-plugins-config/pull/130. Needed for https://github.com/pressbooks/private/issues/2027


fdalcin (approved):
LGTM

Issue #4329, opened 2026-01-22
<td> to <th> on save creates accessibility issues in tables with an empty cell in the first row


MichSpieler:
Recently a change was introduced that modifies table cell markup upon saving the chapter. If a table has a `thead` element, all `td` cells inside it will be transformed to `th`. There are [valid use cases](https://www.w3.org/WAI/tutorials/tables/two-headers/) where this should not be applied to all cells in a row inside `thead`. Some Pressbooks users have such tables in their chapters and they have manually marked up the tables correctly and in an accessible way. If they save such a chapter now, their correct markup gets replaced with incorrect markup (an empty cell with `th` instead of `td`). There currently is no way for authors to fix this without changing the table's structure.

There is a case to be made to revert the recent change until a better solution can be found.

tw77:
Reverting the PR had the expected result on dev.pressbooks.network, ready for release

SteelWagstaff:
Discussed this with Mich today. Better solution is to check inside thead elements that contain tr elements. Use the same approach as previous, except check to see if the td element is empty before turning it into a th. If the td is empty, keep it a td. If it has content inside of it, make it a th. Will open PR to implement this improvement shortly.

tw77:
Works as expected on dev, ready to release.

tw77:
Actually, I just noticed one thing and have to write a quick question here, moving this back to Testing QA. One second...

tw77:
An interesting side effect of the latest code on dev is that empty header cells are automatically converted to `<td>` on save. The Editoria11y alert for an empty header cell is therefore immediately cancelled out by pressing save.

Is this desirable from an accessibility standpoint? I thought `<td>` was only used in the rare scenario where an empty header cell is intentional. In most cases, the better practice is to not have empty header cells. If we release exactly what we have on dev.pressbooks.network, we'd be automatically resolving the empty header cell alert before it can be acted on by authors.

tw77:
If possible, I think a better approach would be:

- if a `<td>` in a header row is empty, keep it as a `<td>` to preserve intentional formatting from the Code editor.
- if a `<th>` in a header row is empty, keep it as a `<th>`so that Editoria11y can flag the empty header cell.

SteelWagstaff:
@tw77 Yes, I understand the case you're describing. For me, the current functionality is desirable for two reasons:
1. We already shipped code that broke tables for some users -- converting intentionally blank <td> elements into <th> elements. This change would fix this for them automatically.
[...]

PR #4330, opened 2026-01-22
chore: revert "fix: convert all td elements to th in thead elements before save"

arzola:
Reverting pressbooks/pressbooks#4304

https://github.com/pressbooks/pressbooks/issues/4329

Until we have a more reliable solution


richard015ar (approved):
LGTM

PR #4333, opened 2026-01-23
fix: correctly handle th and td elements in thead upon save

SteelWagstaff:
Fix for https://github.com/pressbooks/pressbooks/issues/4329. See https://github.com/pressbooks/pressbooks/issues/4329#issuecomment-3789346497

Previously, when users created tables with header rows using the TinyMCE editor, the cells were incorrectly created as <td> elements instead of the semantically correct <th> elements. This PR checks the content before save and safely converts td elements inside of thead rows to the correct th element IF AND ONLY IF they contain content. It also converts emtpy <th> elements to <td> and leaves empty td elements as is (to avoid the problem reported by users after the last fix was released). This will result in accessible markup for thead row content and will remove the Editoria11y checks/flags for malformed thead content upon content save for preexisting and new tables alike.

## How It Works
When a user saves a post with a table that has a header row:

1. TinyMCE creates the table with <thead> but uses <td> elements
2. WordPress fires the content_save_pre filter before saving
3. The newly created function processes the HTML and converts <td> cells within <thead> to <th> elements if they contain content and <th> to <td> if they are empty
4. The corrected HTML is saved to the database

## Changes Made
1. Added a new fix_table_header_cells() function to the Presbooks/Editor namespace
This function processes HTML content using DOMDocument to find all <thead> elements. It converts <td> elements within <thead> to <th> elements IF THEY CONTAIN CONTENT and preserves all attributes (class, colspan, rowspan, etc.). It converts empty <th> elements to <td> elements fixing any problems introduced by the previous release.

It only processes content when <thead> is present. It leaves <tbody> cells untouched.

2. Hooked the function to content_save_pre filter (hooks-admin.php)
This filter runs when content is saved, automatically converting the HTML before it's stored in the database.

3. Added tests.

## To test
1. Add a new table to a book
2. Select a row, click table row properties and choose row type header
3. Add content to one or more of the cells in that row. Leave at least one cell blank
4. Manually convert an empty cell to <th>
5. Save the chapter
6. Observe that cells in the header row with content have been converted to th elements. Observe that cells without content become (or remain) td elements. Observe that there are no Editoria11y warnings after save

https://github.com/user-attachments/assets/61f6f308-30b2-448f-bfdb-87b380823282

### DIFF
--- a/inc/editor/namespace.php
+++ b/inc/editor/namespace.php
@@ -16,6 +16,7 @@ use PressbooksMix\Assets;
 use Pressbooks\Container;
 use Pressbooks\HtmlParser;
 use Pressbooks\Shortcodes\Glossary\Glossary;
+use Masterminds\HTML5;
 
 /**
  * Ensure that Word formatting that we like doesn't get filtered out.
@@ -528,45 +529,28 @@ function fix_table_header_cells( string $content ): string {
 		return $content;
 	}
 
-	return preg_replace_callback(
-		pattern: '/<thead\b[^>]*>.*?<\/thead>/is',
-		callback: function( $matches ) {
-			$thead = $matches[0];
+	$parser = new HTML5();
+	$dom = $parser->loadHTML( $content );
 
-			// Convert <td> with content to <th>
-			$thead = preg_replace_callback(
-				pattern: '/<td\b([^>]*)>(.*?)<\/td>/is',
-				callback: function( $td_matches ) {
-					$attributes = $td_matches[1];
-					$content = $td_matches[2];
+	foreach ($dom->getElementsByTagName('thead') as $thead) {
+		/** Replace <td> elements with content with <th> elements */
+		foreach ($thead->getElementsByTagName('td') as $td) {
+			if (trim($td->textContent) !== '') {
+	    		$th = $dom->createElement("th", $td->nodeValue);
+	      		$td->replaceWith($th);
+			}
+		}
 
-					if ( trim( $content ) !== '' ) {
-						return '<th' . $attributes . '>' . $content . '</th>';
-					}
-					return $td_matches[0];
-				},
-				subject: $thead
-			);
+		/** Replace empty <th> elements with <td> elements */
+		foreach ($thead->getElementsByTagName('th') as $th) {
+			if (trim($th->textContent) === '') {
+	    		$td = $dom->createElement("td", '');
+     			$th->replaceWith($td);
+			}
+		}
+	}
 
-			// Convert empty <th> to <td>
-			$thead = preg_replace_callback(
-				pattern: '/<th\b([^>]*)>(.*?)<\/th>/is',
-				callback: function( $th_matches ) {
-					$attributes = $th_matches[1];
-					$content = $th_matches[2];
-
-					if ( trim( $content ) === '' ) {
-						return '<td' . $attributes . '>' . $content . '</td>';
-					}
-					return $th_matches[0];
-				},
-				subject: $thead
-			);
-
-			return $thead;
-		},
-		subject: $content
-	);
+	return $dom->saveHTML();
 }
 
 /**
--- a/inc/editor/namespace.php
+++ b/inc/editor/namespace.php
@@ -528,25 +528,16 @@ function fix_table_header_cells( string $content ): string {
 		return $content;
 	}
 
-	// Protect LaTeX backslashes before unslashing by replacing with unique placeholder
+	// Protect LaTeX backslashes through WordPress slashing/unslashing
 	$latex_placeholder = '!@#LATEXBACKSLASH#@!';
 	$content = str_replace( '\\\\', $latex_placeholder, $content );
-
-	// Unslash content to get actual HTML (WordPress adds slashes for database protection)
 	$content = wp_unslash( $content );
-
-	// Restore LaTeX backslashes
 	$content = str_replace( $latex_placeholder, '\\', $content );
 
+	// Parse HTML and transform table headers
 	libxml_use_internal_errors( true );
-
 	$dom = new \DOMDocument( '1.0', 'UTF-8' );
-	// Wrap content in div to preserve structure, load as UTF-8
-	$dom->loadHTML(
-		'<html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head><body>' . $content . '</body></html>',
-		LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD
-	);
-
+	$dom->loadHTML( '<?xml encoding="UTF-8">' . $content, LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD );
 	libxml_clear_errors();
 
 	$theads = $dom->getElementsByTagName( 'thead' );
@@ -588,22 +579,10 @@ function fix_table_header_cells( string $content ): string {
 		}
 	}
 
-	// Extract the body content (DOMDocument wraps content in html/body tags)
-	$html = '';
-	$body = $dom->getElementsByTagName( 'body' )->item( 0 );
-	if ( $body ) {
-		foreach ( $body->childNodes as $node ) {
-			$html .= $dom->saveHTML( $node );
-		}
-	}
-
-	// Protect LaTeX backslashes again before slashing (DOMDocument may have single backslashes)
+	// Save HTML and restore LaTeX backslashes through slashing
+	$html = $dom->saveHTML();
 	$html = str_replace( '\\', $latex_placeholder, $html );
-
-	// Re-slash for WordPress (filter expects slashed content)
 	$html = wp_slash( $html );
-
-	// Restore LaTeX backslashes as double-backslashes (for WordPress slashed output)
 	$html = str_replace( $latex_placeholder, '\\\\', $html );
 
 	return $html;
--- a/inc/editor/namespace.php
+++ b/inc/editor/namespace.php
@@ -12,11 +12,11 @@
 namespace Pressbooks\Editor;
 
 use function Pressbooks\Sanitize\normalize_css_urls;
+use Masterminds\HTML5;
 use PressbooksMix\Assets;
 use Pressbooks\Container;
 use Pressbooks\HtmlParser;
 use Pressbooks\Shortcodes\Glossary\Glossary;
-use Masterminds\HTML5;
 
 /**
  * Ensure that Word formatting that we like doesn't get filtered out.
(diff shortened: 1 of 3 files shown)


tw77:
@greatislander The tech team is wondering if the approach taken in this PR seems good to you or if there's a different approach you might suggest?

I'd note that it's already possible, in production networks, to define a header row with th elements for selected cells; however, the only way to do that in the visual editor is to modify Table Cell Properties as described at https://guide.pressbooks.com/chapter/create-tables#chapter-802-section-4, which may be cumbersome / not the most intuitive.

The change in this open PR seeks to make it easier to define a header row with th elements. The previous attempt at this change, https://github.com/pressbooks/pressbooks/pull/4304, was reverted because it led to the issue Steel linked above, https://github.com/pressbooks/pressbooks/issues/4329.

tw77:
After talking with Steel and Mich earlier today I understand this PR as a safe, worthwhile improvement and feel like it would be good to merge to dev, test and release as soon we have time. This change takes effect when the user presses Save, which will already be a helpful improvement in many cases.

A future iteration would be one where changing a row to a header row, from Table Row Properties, instantly handles the code as needed (if an Editoria11y alert is present, this would be resolved _before_ the user saves changes).

greatislander:
@tw77 I'll review this today or tomorrow and have any feedback ready for Monday!

greatislander (changes requested):
Functionally this seems okay. Practically, I don't think it's a good approach as it's unwise to use regex to parse HTML (see: https://stackoverflow.com/a/590789)

I would recommend using [`masterminds/html5`](https://github.com/Masterminds/html5-php) which is already a Pressbooks dependency to parse the content into a [DomDocument](https://www.php.net/manual/en/class.domdocument.php) instance, manipulating that instance, and then saving the result.

Bonus: when Pressbooks bumps the PHP requirement to 8.4, the `masterminds/html5` library will no longer be needed as the built-in Dom utilities are [HTML5-compatible](https://www.php.net/manual/en/migration84.new-features.php#migration84.new-features.dom).

SteelWagstaff:
Thank you Ned! I think I can redo this functionality using the recommended approach. Will read up on masterminds and look at other places we are using it and revise this PR soon.

greatislander:
@SteelWagstaff I took a crack at it in 249be47. Hope that's okay! (Still WIP.)

SteelWagstaff:
Yeah works for me -- all I care about is the end result, how we get there could be any number of ways. Thanks for the improvement.

SteelWagstaff:
Three notes on your solution:
1. I think `loadHTMLFragment()` might be better for fragments (like we have here) instead of `loadHTML()` and `saveHTML()`. 
2. I think we might have a live collection problem with the foreach loop. I believe getElementsByTagName(`td`) will return a _live_ HTMLCollection that updates dynamically as the DOM changes.  In a table with `<td>A</td><td>B</td><td>C</td>` the iterator will move to position 1 after the first element gets converted, but the third element will be in that position at that point. The end result will be something like `<th>A</th><td>B</td><th>C</th>`. 
3. I think `createElement('th', $td->nodeValue)` only gets plain text. We'll lose nexted HTML inside the <td> elements (like `<strong>` or `<a>`). 

What if we did something like this:
```
function fix_table_header_cells( string $content ): string {
	// Only process if content contains tables with thead
	if ( empty( $content ) || ! str_contains( $content, '<thead' ) ) {
		return $content;
	}

	$html5 = new \Masterminds\HTML5( [ 'disable_html_ns' => true ] );
	
	$dom_fragment = $html5->loadHTMLFragment( $content );
	if ( ! $dom_fragment ) {
		return $content;
	}

	$dom = $dom_fragment->ownerDocument;
	$xpath = new \DOMXPath( $dom );
	$theads = $xpath->query( './/thead', $dom_fragment );

	foreach ( $theads as $thead ) {
		// Get all td and th elements within rows
		$rows = $thead->getElementsByTagName( 'tr' );
		foreach ( $rows as $row ) {
			// Convert NodeList to array to avoid issues with live collections during DOM modification
			$cells = [];
			foreach ( $row->childNodes as $node ) {
				if ( $node->nodeType === XML_ELEMENT_NODE && ( $node->nodeName === 'td' || $node->nodeName === 'th' ) ) {
					$cells[] = $node;
				}
			}

			foreach ( $cells as $cell ) {
				$has_content = trim( $cell->textContent ) !== '';
				$is_td = $cell->nodeName === 'td';

				// Convert td with content to th, or th without content to td
				if ( ( $is_td && $has_content ) || ( ! $is_td && ! $has_content ) ) {
					$new_tag = $is_td ? 'th' : 'td';
					$new_cell = $dom->createElement( $new_tag );
					
					// Copy attributes
					if ( $cell->hasAttributes() ) {
						foreach ( $cell->attributes as $attr ) {
							$new_cell->setAttribute( $attr->name, $attr->value );
						}
					}
					
					// Copy child nodes
					while ( $cell->firstChild ) {
						$new_cell->appendChild( $cell->firstChild );
					}
					
					$cell->parentNode->replaceChild( $new_cell, $cell );
				}
			}
		}
	}

	// Serialize the fragment back to HTML
	$result = '';
	foreach ( $dom_fragment->childNodes as $node ) {
		$result .= $html5->saveHTML( $node );
	}

	return $result ?: $content;
}
```

SteelWagstaff:
Pushed the changes I just proposed to make sure test suite passed (they did locally).

SteelWagstaff:
I think this one is ready for new review, Ned.

greatislander:
@SteelWagstaff I think this is good. The one thing I noticed is that when I test it in the editor inline styles (e.g. `<table style="border-collapse: collapse; width: 100%;" border="0">` which are added by default when creating a new table) are stripped out. I think this behaviour is unrelated to this change, though.

greatislander (approved):

SteelWagstaff:
Thanks @greatislander. We observed a couple of preexisting bugs with the way the visual editor creates and handles these rows (for example, if you select a row and mark it as a header and then select it later and remove the header property, the row and its content is moved to the _bottom_ of the table instead of remaining in place at the top of the table. Our plan was to note and handle those separately.

inc/editor/namespace.php
```diff
@@ -515,6 +515,78 @@ function show_kitchen_sink( $args ) {
...
+		}
+	}
+
+	// Serialize the fragment back to HTML
+	$result = '';
+	foreach ( $dom_fragment->childNodes as $node ) {
+		$result .= $html5->saveHTML( $node );
+	}
+
+	return $result ?: $content;
```
fdalcin: There is an issue with this way of saving the HTML.

Saving the following table structure:

```HTML
<table class="grid" border="0">
<thead>
<tr>
<th colspan="1" class="some-class">This is a Header</th>
<th colspan="1">This is another Header</th>
<td colspan="1"></td>
<td colspan="1"></td>
</tr>
</thead>
<tbody>
<tr>
<td>This is a column</td>
<td>This is another column</td>
<td>This is another column</td>
<td></td>
</tr>
</tbody>
</table>
```

Will result in the following output:

```HTML
<table class="&quot;&quot;grid&quot;&quot;" border="&quot;&quot;0&quot;&quot;">
<thead>
<tr>
<th colspan="&quot;&quot;1&quot;&quot;" class="&quot;&quot;some-class&quot;&quot;">This is a Header</th>
<th colspan="&quot;&quot;1&quot;&quot;">This is another Header</th>
<td colspan="&quot;&quot;1&quot;&quot;"></td>
<td colspan="&quot;&quot;1&quot;&quot;"></td>
</tr>
</thead>
<tbody>
<tr>
<td>This is a column</td>
<td>This is another column</td>
<td>This is another column</td>
<td></td>
</tr>
</tbody>
</table>
```

This is not expected and will cause issues with attributes added to those tables.

fdalcin (changes requested):
The code works as expected for the `td` with `th` replacement. However it fails to update the quotes in the HTML.

inc/editor/namespace.php
SteelWagstaff: Thanks for catching, Dalcin. I did not notice this in my testing, but can reproduce. Will attempt to fix before requesting re-review

SteelWagstaff: I think I found the problem: my function was calling saveHTML() inside the foreach loop (while manually iterating through nodes). This caused double-encoding when I ran saveHTML() again at the end outside the loop. I can use the HtmlParser wrapper class to better handle HTML serialization.

fdalcin (dismissed):
I have reviewed this carefully now and while it is now working as expected for the tables, it comes with unexpected behaviors to the HTML.

This is the chapter content I have locally:

```HTML
There is a bunch "texts" here that we should address with 'quotes'. But what does it happen when we have two single quotes '' in place like ''here''?

<span class="strong">Some additional attribute</span>

\[ f(x) = f(a) + f'(a)(x-a) + frac{f''(a)}{2!}(x-a)^2 + frac{f'''(a)}{3!}(x-a)^3 + cdots \]

\[ e^{i pi} + 1 = 0 \]

 
<table class="grid" border="0">
<thead>
<tr>
<td colspan="1" class="some-class">This is a Header</td>
<td colspan="1">This is another Header</td>
<th colspan="1"></th>
<th colspan="1"></th>
</tr>
</thead>
<tbody>
<tr>
<td>This is a column</td>
<td>This is another column</td>
<td>This is another column</td>
<td></td>
</tr>
</tbody>
</table>
```

I would expect it to correctly fix the table header with `th` and `td` elements but keep the rest unchanged, like so:

```HTML
There is a bunch "texts" here that we should address with 'quotes'. But what does it happen when we have two single quotes '' in place like ''here''?

<span class="strong">Some additional attribute</span>

\[ f(x) = f(a) + f'(a)(x-a) + frac{f''(a)}{2!}(x-a)^2 + frac{f'''(a)}{3!}(x-a)^3 + cdots \]

\[ e^{i pi} + 1 = 0 \]

 
<table class="grid" border="0">
<thead>
<tr>
<th colspan="1" class="some-class">This is a Header</th>
<th colspan="1">This is another Header</th>
<td colspan="1"></td>
<td colspan="1"></td>
</tr>
</thead>
<tbody>
<tr>
<td>This is a column</td>
<td>This is another column</td>
<td>This is another column</td>
<td></td>
</tr>
</tbody>
</table>
```

However, it currently breaks math expressions by removing the `\`.

```HTML
There is a bunch "texts" here that we should address with 'quotes'. But what does it happen when we have two single quotes '' in place like ''here''?

<span class="strong">Some additional attribute</span>

[ f(x) = f(a) + f'(a)(x-a) + frac{f''(a)}{2!}(x-a)^2 + frac{f'''(a)}{3!}(x-a)^3 + cdots ]

[ e^{i pi} + 1 = 0 ]

 
<table class="grid" border="0">
<thead>
<tr>
<th colspan="1" class="some-class">This is a Header</th>
<th colspan="1">This is another Header</th>
<td colspan="1"></td>
<td colspan="1"></td>
</tr>
</thead>
<tbody>
<tr>
<td>This is a column</td>
<td>This is another column</td>
<td>This is another column</td>
<td></td>
</tr>
</tbody>
</table>
```

We will need to work this a bit more before merging it as it has implications outside of the table elements.

SteelWagstaff:
Rushing around between meetings today, but believe the issue you found is due to not re-slashing before saving the adjusted HTML (I unslashed to get rid of double encoding but forgot to reslash before returning). Made a change and added new tests. Will continue to test locally before marking for re-review

SteelWagstaff:
Finally got it working -- had to do some creative str_replace backslash protection, but this should now be ready for review. If you want some sample HTML to test it with in your chapter, try:
```
<h2>Test 1: LaTeX in Table Headers</h2>
<p>This paragraph has LaTeX: \( f(x) = \frac{1}{2} \)</p>
<table class="grid">
<thead>
<tr>
<td>\[ x = y \]</td>
<td>Normal Héader</td>
<td>\sum_{i=1}^{n} i</td>
</tr>
</thead>
<tbody>
<tr>
<td>Data cell 1</td>
<td>Data cell 2</td>
<td>Data cell 3</td>
</tr>
</tbody>
</table>
<h2>Test 2: Empty and Mixed Cells</h2>
<table>
<thead>
<tr>
<td>"Header" 1: 6\7 people</td>
<th></th>
<td>Another Niño Header</td>
<td></td>
<th></th>
</tr>
</thead>
<tbody>
<tr>
<td>Data</td>
<td>More data</td>
<td>Even more</td>
<td>Last cell</td>
<td>Final</td>
</tr>
</tbody>
</table>
<h2>Test 3: Attributes and Special Characters</h2>
There are "double quotes" and 'single quotes' [pb_glossary id="66"]here[/pb_glossary].

<span class="strong">Bold text</span>
<table class="grid" border="0">
<thead>
<tr>
<td class="some-class" colspan="1">This is a Header—with an em dash</td>
<td colspan="1">中文</td>
<th colspan="1"></th>
</tr>
</thead>
<tbody>
<tr>
<td>This is a column</td>
<td>This is another column</td>
<td>This is another column</td>
<td></td>
</tr>
</tbody>
</table>
<p>Paragraph after the table. There's some display math next:</p>
\[ e = mc^2\]
```

SteelWagstaff:
@fdalcin @arzola @greatislander I don't know what you have planned for the rest of your sprint, but I believe this is now actually ready for review.

greatislander (commented):
Looks good to me, but I'll leave it to the tech team to approve for merge.

arzola (approved):
LGTM reviewed and test but I would prefer to have proper Wiz's testing

arzola (approved):

PR #4347, opened 2026-02-13
feat: add hooks when network managers are added or removed

SteelWagstaff:
Add action hooks which are fired when network managers are added or removed, allowing for more granular event handling. Partial fix for https://github.com/pressbooks/private/issues/1516. Accompanies https://github.com/pressbooks/pressbooks-multi-institution/pull/298 and https://github.com/pressbooks/pressbooks-plugins-config/pull/148.

* Add tracking variables (`$was_added`, `$was_removed`) to determine if a network manager was added or removed during the function execution.
* Introduce new action hooks: `pressbooks_network_manager_added` and `pressbooks_network_manager_removed`, fired after a manager is added or removed.

### DIFF
--- a/inc/admin/networkmanagers/namespace.php
+++ b/inc/admin/networkmanagers/namespace.php
@@ -95,7 +95,7 @@ function update_admin_status() {
 		} else {
 			delete_site_option( 'pressbooks_network_managers' );
 		}
-		
+
 		// Fire hooks after successful manager changes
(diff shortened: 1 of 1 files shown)


arzola (approved):

arzola (approved):
LGTM I was thinking that we should just call the action before without creating those flags but I think it's safer to run this after the `update_site_option` call

PR #4348, opened 2026-02-13
chore: Update Louisiana Institutions and add East Los Angeles College to list of institutions

MichSpieler:
Update Louisiana Institutions and add East Los Angeles College to Institution list.

Addresses https://github.com/pressbooks/private/issues/2271


SteelWagstaff (approved):
LGTM. Thanks Mich!

PR #4355, opened 2026-02-24
td content improvement

arzola:
This pull request makes a targeted adjustment to the logic for fixing table header cells in the `fix_table_header_cells` function. The change ensures that only `<td>` elements with content are converted to `<th>`, improving the accuracy of table header detection.

- Table header cell logic improvement:
  * Updated the conditional in `fix_table_header_cells` so that only `<td>` elements with content are converted to `<th>`, instead of also converting `<th>` elements without content to `<td>`. (`inc/editor/namespace.php`)


fdalcin (dismissed):
LGTM

fdalcin (commented):
LGTM

fdalcin (approved):

Issue #4352, opened 2026-02-19
Delete book email


richard015ar:
As a book administrator, I want to receive a confirmation email when I delete a book, whether I am the admin_email user or not.

Right now, the delete action email is handled by WP core, and it sends an email to the `admin_email` user of the book by default, which is not necessarily the user who performed the action.

For more context see: https://github.com/pressbooks/private/issues/2302

arzola:
PR here: https://github.com/pressbooks/pressbooks/pull/4357

I read the following comment:

> Before a final refinement we need to discuss if we want to send a copy of the email to each administrator, or just the one who request the delete action, for example.

My assumption is regularly book admins could request a book deletion and then the book admin will approve the deletion request that's why the requester that could be a regular book admin doesn't receive the deletion email but if we want we can send a copy to the requester **if their role is super admin** or **network administrators** as they call them, so two mails will be send one for the requester and one for the original admin (we should honor that default behavior).

tw77:
Works as expected on dev: confirmation email is received by the user who initiates the "delete book" action from the book dashboard, with the book owner / creator / admin user cc'd if that is a different user. As this action is only reachable by book admins and super admins, this is safe to release.

For future reference, this action, completed by clicking the link in the confirmation email, only deactivates the book. Actually permanent book deletion, by contrast, is only available for super admins (in the broad sense of the term, including network and institutional managers) from the network book list. With permanent deletion of a book from the network book list, no email step is involved.

PR #4357, opened 2026-02-25
fix: enable book deletion email for requesting super admins

arzola:
See #4352 

This pull request enhances the site deletion process by ensuring that super admins who request a book deletion also receive the confirmation email, even if their email differs from the book's admin email. It introduces a new method to add the current user to the email recipients and provides comprehensive tests to verify this behavior.

Improvements to email recipient logic for site deletion:

* Added the `addCurrentUserToDeleteEmail` method to `class-book.php`, which checks if the current super admin's email differs from the book's admin email and, if so, adds them as a recipient of the delete confirmation email.
* Registered the new method as a filter on `wp_mail` to ensure it is applied when sending delete confirmation emails.

Testing enhancements:

* Added multiple tests in `test-admin-delete-book.php` to verify the new email recipient logic, including cases for ignoring non-delete emails, adding the current user, skipping when emails match, and skipping non-super admins.

Minor corrections:

* Fixed a placeholder typo in the delete email content from `SITE_NAME` to `SITENAME`.


ho-man-chan (approved):
LGTM

PR #4375, opened 2026-03-11
fix: language mapping

arzola:
This pull request makes a small change to the `crowdin.yml` configuration file, removing the `ta-IN` language mapping.


fdalcin (approved):
LGTM

PR #4383, opened 2026-03-17
fix: restore pin toggle functionality

arzola:
This pull request updates the pin-saving functionality in the export script to improve how pin data is sent to the server and clarifies the server action being triggered.

**Export pin data handling:**

* Changed the AJAX action from `'pb_export_pins'` to `'pb_update_pins'` to better reflect the operation being performed.
* Updated the way pin data is sent to the server by serializing the `pins` array with `JSON.stringify` before sending it in the AJAX request.


fdalcin (approved):
LGTM

Issue #4390, opened 2026-03-18
MS Word Import does not work


MichSpieler:
When a user tries to upload a Word document to import into Pressbooks, the result is an error message:

"No chapters were selected for import."

As the screen capture shows, a chapter has been selected for import.

<img width="1147" height="437" alt="Image" src="https://github.com/user-attachments/assets/c63b8b39-e646-4580-bcab-9090b4ac4d9d" />

This was reported by a user and I have reproduced it with two different files.

SteelWagstaff:
@MichSpieler this happens to me often. Usually I can resolve it by clicking 'import selection' a second time (which makes me think it's a caching problem). Does that work in this case as well? I do agree the bug should still be fixed, but it's less urgent if this workaround still works.

MichSpieler:
No, that's what used to work. But the user reported it does not.

MichSpieler:
I tried 7 times

<img width="1149" height="729" alt="Image" src="https://github.com/user-attachments/assets/36804ff5-766d-4c04-9c9d-45236be2b792" />

PR #4391, opened 2026-03-18
fix: missing jquery dependency in import

arzola:
Issue https://github.com/pressbooks/pressbooks/issues/4390

This pull request makes a minor update to the script enqueuing logic for the import page in the admin interface. The change ensures that the `pb-import` script explicitly lists its dependencies, which helps guarantee that required scripts like `jquery` and `jquery-form` are loaded beforehand.

* Added explicit `dependencies` (`jquery`, `jquery-form`) to the `pb-import` script when enqueuing it in `replace_book_admin_menu()` in `inc/admin/laf/namespace.php`


fdalcin (approved):
LGTM

Issue #4395, opened 2026-03-19
WordPress' native adjustSubmenu function gets broken in Pressbooks


SteelWagstaff:
Core WordPress includes a javascript function that is designed to ensure that an admin submenu is displayed within the visual viewport. It works by adding a negative margin-top value to the submenu if it detects that it will display too close to the bottom of the page: https://github.com/WordPress/WordPress/blob/8745a4388b2e5d3ace9f13629d51d33b8faa0b0b/wp-admin/js/common.js#L911-L946

<img width="347" height="248" alt="Image" src="https://github.com/user-attachments/assets/4e5bda2e-47fc-4468-812e-6aa33c833637" />

Pressbooks does something that breaks this. When I hover over an admin item, the parent li receives the opensub class via javascript, but the child submenu never receives the margin-top adjustment. I don't yet know what but want to figure it out.

<img width="347" height="132" alt="Image" src="https://github.com/user-attachments/assets/5f9ace96-a7a6-4ba3-be65-108dff62a59d" />

SteelWagstaff:
Found the problem! In admin.scss we set #wp-toolbar > ul > li { display: list-item; } which makes _all_ toolbar items visible,  including `li#wp-admin-bar-menu-toggle`. We set display: none; for the the child `.ab-item`, but the li itself remained visible

WP core's `adjustSubmenu()` checks $('#wp-admin-bar-menu-toggle').is(':hidden'). It queries the li element, _not its children_. Since the li was visible (just empty), jQuery returns false, and margin-top is never applied. I fix this with #4396

MichSpieler:
Is this the same issue as https://github.com/pressbooks/pressbooks/issues/4187?

SteelWagstaff:
Yes! Sorry I didn't see that one.

SteelWagstaff:
We can close either one as duplicate, IMO. Sorry for extra work. Did a cursory search but didn't immediately discover your ticket.

tw77:
Closing in favour of https://github.com/pressbooks/pressbooks/issues/4187

Issue #4187, opened 2025-08-26
Menu subitems overflow the bottom of the screen


MichSpieler:
When a window is below a certain height or a large menu item is expanded to show all sub-items, the menu sub-items towards the bottom of the screen become inaccessible. Clicking on the main menu item is a viable work around, as it displays sub-items in a scrollable window.

<img width="445" height="638" alt="Image" src="https://github.com/user-attachments/assets/dfc7520d-acba-4692-a29d-72933684879d" />

MichSpieler:
From Bug Scrub: could be nice to add a scroll to the submenu.

HVbajoria:
Hi @MichSpieler ,

Tried fixing it: https://github.com/pressbooks/pressbooks/pull/4236

Could you please check once and help me with it?

arzola:
Review and test PR.

arzola:
I reviewed PR and needs some changes

tw77:
This is solved on dev.pressbooks.network, ready for release

PR #4396, opened 2026-03-19
fix: restore native WordPress jquery functionality for submenu items (#4187)

SteelWagstaff:
Fix for #4187 and #4395

Make minor adjustments to the admin interface styles to simplify the CSS rules for toggling menu display so that a native WordPress submenu repositioning feature works again. 

* Change the selector for hiding the menu toggle in `admin.scss` to target the entire list item instead of just the `.ab-item`, ensuring the menu toggle is hidden more reliably.
* Update the selector for displaying the menu toggle in `admin.scss`, making the list item visible and setting its child `.ab-item` to use flex layout for improved alignment.
* Remove the `transform: translateZ(0);` property from `#adminmenu` in `pressbooks.scss`. This was added to address a problem with Chrome that was fixed a decade ago (see https://core.trac.wordpress.org/ticket/33199)

Also includes two small fixes to get rid of lint nags during the Vite build process:
* Remove empty `dashboard.js` file from src and build step
* Replace `eval` for template string interpolation in `glossary.js` with a safer and more direct string replacement method


assets/src/scripts/glossary.js
```diff
@@ -63,8 +63,8 @@
 					} else {
 						myActiveTab = 0;
 						if ( mySelection ) {
-							let templateString1 = mySelection.trim(); // eslint-disable-line no-unused-vars
-							termExistsMessage = eval( '`' + PB_GlossaryToken.not_found.replace( /`/g, '' ) + '`' ); // eslint-disable-line no-eval
+							let templateString1 = mySelection.trim();
+							termExistsMessage = PB_GlossaryToken.not_found.replace( '${templateString1}', templateString1 );
```
SteelWagstaff, line 67: Rewrote these two lines so they don't throw lint warnings anymore. If you want to test it doesn't affect functionality
1. Open the editor (TinyMCE)
2. Select some text that doesn't already have an existing Glossary term (e.g., "supported")
3. Click the Insert Glossary Term button
4. You should see the message with the template string interpolated: "Glossary term _supported_ not found. Please create it."

Image

inc/admin/dashboard/class-dashboard.php
```diff
@@ -64,7 +64,6 @@ public function enqueueStyles( string $page ): void {
 		add_action( "admin_print_styles-{$page}", function() {
 			/** @var \PressbooksFrontendTools\Assets $assets */
 			$assets = app( 'Assets' );
-			$assets->enqueue( 'assets/src/scripts/dashboard.js', 'pb-book-dashboard' );
```
SteelWagstaff, line 67: No longer need to enqueue the dashboard.js file as it has been removed from build assets (it was empty)

assets/dist/manifest.json
```diff
@@ -131,12 +131,6 @@
       "assets/covergenerator-BPTG2Ppo.css"
     ]
   },
-  "assets/src/scripts/dashboard.js": {
```
SteelWagstaff, line 134: Remove empty .js file

assets/src/scripts/dashboard.js
SteelWagstaff, line 1: Delete empty file

arzola (approved):
LGTM works as expected

Issue #4397, opened 2026-03-19
Analytics submenu item invisible when active


MichSpieler:
The "Analytics" submenu item in the network admin view has the same colour as the background when it's active.

<img width="507" height="414" alt="Image" src="https://github.com/user-attachments/assets/d0a8eb1f-ff9e-489e-9266-328ef4fd6f00" />

tw77:
We may be losing the sidebar menu organization context on this network-level Analytics page (as the URL changes to no longer include `/network/` here)

arzola:
I opened a PR https://github.com/pressbooks/pressbooks/pull/4408

tw77:
This is fixed on dev.pressbooks.network, ready for release

PR #4408, opened 2026-03-30
fix: stats analytics submenu highlight

arzola:
This PR address #4397 

Sidebar has been a nightmare since it was implemented for now to "fix" this we should apply a tweak but I hope soon we tackle this in a better way.

Pasted_Image_2026-03-30__6_49 PM


inc/admin/menus/class-sidebar.php
```diff
@@ -540,100 +601,157 @@ private function addSuperAdminMenuItems(): void {
...
-					66
+					66,
 				);
 			}
 		}
 	}
+
+	/**
+	 * Fix the parent file for Koko Analytics so the Stats menu is highlighted.
+	 *
```
arzola, line 624: This is the important part, the other part is just `lando composer fix`

richard015ar (approved):
LGTM

PR #4411, opened 2026-03-31
fix: store export pins as options instead of transients

arzola:
I just noticed that pins are stored in a transient entry those are always whipped out when we clear the cache

    With Object Cache Pro (Redis/Memcached), WordPress transients are stored in the object cache rather than the database. Every time we clear the network cache whether during deployments or routine maintenance all transients get wiped, including pb_export_pins.

When we use transients without a caching provider (object cache pro) the settings are stored in the database, but our current setup is set to be a volatile cache (redis).

I think we should store those as wp_options instead

I checked out the version before the background exports refactor and yes those pins were removed after clearing the cache ONLY when using Object Cache Pro, before object cache pro integration was working as expected pins were not cleared even after clearing the cache.

This pull request updates the pinning logic for export formats to improve reliability and enforce a limit on the number of pinned exports per format. The main changes include switching from WordPress transients to persistent options for storing pin data, introducing a per-format pin limit, improving error messaging, and updating related tests.

**Pin storage and management:**
- Replaced usage of `get_transient`/`set_transient` with `get_option`/`update_option` for storing and retrieving export pins, ensuring pin data persists reliably and is not subject to transient expiration. [[1]](diffhunk://#diff-672361dc1fc6ea96c9cbb679673ce69f151f161179e65d579c01f47569e48b29L230-R230) [[2]](diffhunk://#diff-672361dc1fc6ea96c9cbb679673ce69f151f161179e65d579c01f47569e48b29L271-R271) [[3]](diffhunk://#diff-3bbc455a78900441581d45e9a9dab3a58f55934e0462a20ab0343f55e8291598R286-R311) [[4]](diffhunk://#diff-0cd07c8739dfc04dd222dfee6606ed2669cda91861a34c1324d074d35dc12664L159-R160) [[5]](diffhunk://#diff-0cd07c8739dfc04dd222dfee6606ed2669cda91861a34c1324d074d35dc12664L175-R175) [[6]](diffhunk://#diff-0cd07c8739dfc04dd222dfee6606ed2669cda91861a34c1324d074d35dc12664L198-R199)

**Pin limit enforcement:**
- Added a new function `is_format_pin_limited` to check if a specific export format has reached the maximum allowed number of pins (default: 3). This prevents users from pinning more than the allowed number of exports per format.
- Updated the export job queuing logic to block new pinned exports if the limit is reached, returning a clear, user-friendly error message.

**Error handling improvements:**
- Enhanced error messaging when no export jobs are successfully queued, now listing specific error details if available, rather than a generic message.

**Other minor changes:**
- Minor style adjustment in the admin table for custom fonts.
- Streamlined JSON response handling in the export JavaScript to always return parsed data.

### How to test

1. Pin 3 versions of the same export type
2. Go to network settings and flush cache
3. Go back to your book and your pins should be still there
4. Try to produce another export of the same kind 
5. You should get an alert displaying that you can't produce another export of that kind until you unpin one.


assets/src/scripts/export.js
```diff
@@ -310,11 +310,13 @@ document.addEventListener( 'DOMContentLoaded', function () {
 			body: serverFormData,
 		} );
 
+		const data = await response.json();
+
 		if ( ! response.ok ) {
-			throw new Error( `HTTP error! status: ${ response.status }` );
+			return data;
 		}
```
richard015ar: We could delete this condition, since it doesn't throws an error anymore. Do we want to return data anyway?

richard015ar (changes requested):
Feature works and code LGTM, I just left a small comment.

assets/src/scripts/export.js
arzola: Yes you're right that conditional is not required anymore we are handling the response in the caller, I just removed it

richard015ar (approved):
LGTM

PR #4413, opened 2026-03-31
chore: remove user activation keys menu (plugin deprecation)

arzola:
This pull request removes the integration of the "User Activation Keys" plugin from the super admin menu in the admin sidebar. The code that conditionally added the "User Activation Keys" submenu item has been deleted.

Menu cleanup:

* Removed the check for the activation of the `user-activation-keys/ds_wp3_user_activation_keys.php` plugin and the related submenu page for "User Activation Keys" in the `addSuperAdminMenuItems` method of `class-sidebar.php`.


fdalcin (approved):
LGTM

PR #4414, opened 2026-04-01
fix: language attributes display on PDF

arzola:
This pull request focuses on improving XHTML and XML language attribute handling, as well as refining entity decoding in the sanitization process. It also adds a unit test to ensure correct decoding of double quotes in XML attributes. The most important changes are grouped below:

**Language Attribute Handling Improvements:**

* Added the `xml:lang` attribute to the configuration arrays in both the EPUB and XHTML export modules to ensure proper language tagging in generated files. [[1]](diffhunk://#diff-4d36e712f4455de8592afc2af3953c30f7cf23921703a5c4ce5980930b1ed895R740) [[2]](diffhunk://#diff-4d36e712f4455de8592afc2af3953c30f7cf23921703a5c4ce5980930b1ed895R870) [[3]](diffhunk://#diff-d114cd98af204df5d88c00b9c9e8cede1cb667d34a520b04a3016b1869717e46R1151) [[4]](diffhunk://#diff-d114cd98af204df5d88c00b9c9e8cede1cb667d34a520b04a3016b1869717e46R1174)
* Updated the XHTML export's `echoDocType` method to include both `xml:lang` and `lang` attributes on the root `<html>` element for better standards compliance and compatibility.

**Sanitization and Decoding Enhancements:**

* Changed the `decode` function in `namespace.php` to use `ENT_COMPAT` instead of `ENT_NOQUOTES` for `html_entity_decode`, ensuring that double quotes are correctly decoded and improving round-trip handling of XML attributes.
* Added a new unit test to verify that double quotes are handled correctly when decoding sanitized XML attributes, specifically checking that `&quot;` does not appear in the decoded output and that language spans are preserved.

### How to test

1. Checkout this branch
2. Add some chapter titles or texts with language attributes for instance

```
<span lang="fr">en français</span>
<span lang="es">Ejemplo</span>
```

3. Titles should have the attribute with `lang="es"` single doubles quote not double.
4. PDF accessibility sidebar should display the expected language in the node was applied

Pasted_Image_2026-03-31__8_41 PM

### DIFF
--- a/inc/modules/export/epub/class-epub.php
+++ b/inc/modules/export/epub/class-epub.php
@@ -737,6 +737,7 @@ class Epub extends Export {
 
 		$config = [
 			'valid_xhtml' => 1,
+			'xml:lang' => 1,
 			'no_deprecated_attr' => 2,
 			'unique_ids' => 'fixme-',
 			'hook' => '\Pressbooks\Sanitize\html5_to_epub3',
@@ -866,6 +867,7 @@ class Epub extends Export {
 	protected function html5ToXhtml( string $html ): string {
 		$config = [
 			'valid_xhtml' => 1,
+			'xml:lang' => 1,
 			'unique_ids' => 0,
 		];
 
--- a/inc/modules/export/xhtml/class-xhtml11.php
+++ b/inc/modules/export/xhtml/class-xhtml11.php
@@ -1148,6 +1148,7 @@ class Xhtml11 extends Export {
 
 		$config = [
 			'valid_xhtml' => 1,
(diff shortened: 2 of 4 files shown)


richard015ar (approved):
change LGTM. However, I couldn't test it since my adobe acrobat is not PRO, so I don't have the accessibility checker available to inspect. I've tried other online tools but no one offers dom inspection. There are other free tools described [here](https://www.w3.org/WAI/test-evaluate/tools/list/), but those are for windows. I'll merge this and let Thomas to test, code looks good, thanks bro.

PR #4419, opened 2026-04-09
chore: improve single delimiter settings page instructions

arzola:
This pull request makes a small improvement to the MathJax admin template by clarifying the note about how dollar signs are interpreted in LaTeX math mode.

* Updated the instructional text in `templates/admin/mathjax.blade.php` to combine and clarify the explanation about spaces around dollar signs, ensuring users understand how to avoid unintended math formatting for currency values like "$ 5,000".


richard015ar (approved):
LGTM

PR #4423, opened 2026-04-13
fix: mathjax \div regression

### LINKED ISSUE #117: Address \div producing a new symbol
Following the recent MathJax update,`[latex]\div[/latex]` now produces a [del](https://en.wikipedia.org/wiki/Del) symbol rather than a division symbol like before:

Before:
Image

After:
Image

Using `\divsymbol` produces a division symbol, though.

This seems to be because of how the `physics` package handles the`\div` command, currently overriding the original TeX command. Discussing with devs, a potential solution that was brought up was an automatic check to determine whether or not to enable `physics` in a book, enabling it only when physics notation is used.

Definition of done: for books that don't aim to use the `physics` package, `\div` produces division symbols again.

arzola:
This pull request removes the MathJax "physics" extension from being loaded and adds a test to ensure it is not included in the headers. This change helps prevent unnecessary or potentially problematic MathJax extensions from being loaded.

Fixes https://github.com/pressbooks/pb-mathjax/issues/117

**Removal of the "physics" extension:**

* The "physics" extension is no longer included in the MathJax configuration arrays in `inc/class-mathjax.php`, ensuring it is not loaded on the frontend. [[1]](diffhunk://#diff-75471e261b56521dd46690fedb88783fbf2ce63be3148866d236767dfe25fde2L315) [[2]](diffhunk://#diff-75471e261b56521dd46690fedb88783fbf2ce63be3148866d236767dfe25fde2L340)

**Testing:**

* Added a new test `testAddHeadersDoesNotIncludePhysicsPackage` in `tests/test-mathjax.php` to verify that the "physics" extension is not present in the output of `addHeaders()`.

### Notes for tester

follow instructions here: https://github.com/pressbooks/pressbooks/pull/3973#issue-2961864264


richard015ar (approved):
LGTM

PR #4425, opened 2026-04-13
fix: allow book's users to see archived books

arzola:
This pull request updates the access logic for archived books in `hooks.php` and adds a comprehensive test suite to ensure correct behavior. The main change is to allow access to archived private books for logged-in book members, while maintaining public access for archived public books. The new tests verify these scenarios and edge cases.

**Access control improvements:**

* Updated the archived book access logic in `hooks.php` to allow logged-in book members to access archived private books, while keeping archived public books accessible to everyone.

**Testing enhancements:**

* Added a new test file `tests/test-ms-site-check.php` with unit tests covering access rules for archived public books, archived private books (for both members and non-members), non-archived books, and non-book sites.

### How to test

1. Archive a book that has different kind of users (authors, subscribers, editors)
2. Log in as one of those users and you should be able to navigate book's url and admin pages for that book. (You should know the URL. because archived books are not added into the My books menu nor Catalog)


richard015ar (approved):
LGTM

PR #4435, opened 2026-04-28
chore: remove ip message from reset password email

arzola:
This pull request introduces a new filter to remove the user's IP address from password reset emails, enhancing user privacy. It adds the implementation, hooks it into WordPress, and provides a unit test to ensure the behavior works as expected.

**Password Reset Email Privacy Improvements:**

* Added the `remove_ip_from_password_reset_email` function in `inc/registration/namespace.php` to strip the IP address line from password reset emails.
* Registered the new function with the `retrieve_password_message` filter in `hooks.php`, so it is automatically applied to outgoing password reset emails.

**Testing:**

* Added a new unit test `test_remove_ip_from_password_reset_email` in `tests/test-registration.php` to verify that the IP address is removed and the rest of the email content remains intact.

Address pressbooks/private#2399


fdalcin (approved):
LGTM

PR #4211, opened 2025-09-22
fix: top bar visualization under specific breakpoints

arzola:
This pull request introduces a new `book-context` class to the WordPress admin body, enabling targeted styling for the book admin context. The main changes include adding a filter to append the class, updating admin SCSS for responsive toolbar adjustments based on this class, and adding tests to ensure the class is applied correctly.

**Admin body class enhancements:**

* Added a new function `append_book_admin_context` in `namespace.php` to append the `book-context` class to the admin body classes.
* Registered the new function as a filter on `admin_body_class` in `hooks-admin.php` to ensure the class is added only in the appropriate admin context.

**Styling updates for book context:**

* Updated `admin.scss` to add responsive styling for the admin toolbar when the `book-context` class is present, including width and padding adjustments for toolbar items at different screen sizes.

**Testing:**

* Added a test in `test-admin-laf.php` to verify that the `book-context` class is correctly appended under various scenarios.

image

PR #4456, opened 2026-05-14
fix: resolve display issue with admin bar View link

greatislander:
A regression was introduced in https://github.com/pressbooks/pressbooks/pull/4211 which caused the `View Chapter`, `View Front Matter`, `View Back Matter` and `View Part` links to be improperly wrapped:

Screenshot of the admin bar showing wrapped 'View Chapter' menu item.

This PR addresses the issue and also adjusts the font size for this item to match the rest of the admin bar:

Screenshot of the admin bar showing properly displayed 'View Chapter' menu item.


richard015ar (changes requested):
LGTM and it works as expected. What is missing @greatislander is to compile the assets and push them. Otherwise, it's ready to go. I didn't push it, so I can review and merge (don't want to push and self-approve).

arzola:
@greatislander I know we discussed already about the npm reusable workflow to build assets, it's still WIP we disabled that because it was halting the tests workflows because GH detected the commit coming from an automation, we need to tweak the reusable workflow to amend the PR with assets with a better strategy, meanwhile we are building assets as part of the PR (I also know that this is risky considering these days 0 days exploits during `npm install` scripts), we will be addressing this soon

arzola:
image

I checked out this branch on my local and rebuilt the assets and this is how it looks like on my end

richard015ar:
good point Os, my tests were based on resizing the browser. But after your recommendation I tried with the "Mobile view" in Chrome, and it doesn't work as intended

greatislander:
Updated to use an icon which replaces text at smaller viewports as the text (especially when it's "View Front Matter") was causing the user icon to get pushed onto a second line:

Screenshot of 'View Chapter' icon and menu item

Mobile menu bar with hamburger, network, books, book and view chapter icons

richard015ar (approved):
LGTM, it works on multiple size screens. Thank you, @greatislander

tw77:
I see this is in the Testing QA column, so I tried it out on dev.pressbooks.network and it's working as expected, the new icon is effective as a solution here.

PR #4460, opened 2026-05-15
feat: announce progress bar progress (resolves #3718)

### LINKED ISSUE #3718: Progress bars don't convey status to users of assistive technology
There are a few places in the Pressbooks interface where we have dynamic progress bars that inform sighted users of progress towards a lengthy task (imports, exports, cloning routine). An a11y specialist at Wilfrid Laurier University has reported that there is no indication of progress announced for users of JAWS (screenreader software):  https://pressbooks.zendesk.com/agent/tickets/20515. There should be.

To reproduce:
1. As a sighted user importing a Pressbook.xml file, select the Import Selection button and observe the import progress bar.
    
image001

(This screenshot shows meter depicting current level of progress in uploading a file)
2. Perform this same task using the JAWS screenreader software. Upon importing a Pressbook.xml file and selecting the Import Selection button, you hear no status updates. You wonder whether anything is happening. 

Problem:
The visual interfaces conveys progress, while the aural interface has no indicators of progress.
 
Technical solution:
Use a progress bar with relevant ARIA roles, states, and properties: https://w3c.github.io/aria/#progressbar + https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role#associated_wai-aria_roles_states_and_properties. For further reading, consult https://www.w3.org/TR/wai-aria-1.1/#progressbar and https://pressbooks.library.torontomu.ca/wafd/chapter/progress-bars/

See https://pressbooks.org/blog/2018/11/12/refactoring-slow-forms-using-php-generators-and-event-streams/, https://github.com/pressbooks/pressbooks/issues/1282, and https://github.com/pressbooks/pressbooks/pull/1589 for implementation. It looks to me from the PR that aria updates were a part of the initial implementation, but a user has reported not hearing updates when performing the import action.

greatislander:
Resolves #3718 by injecting percentage of progress bars for import, export, clone and cover generator operations into a live region.

As per https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA25:

> The following example uses an element with a `role="progressbar"` attribute. The `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` values on the `progressbar` component programmatically communicate the progress of the upload, but because the `progressbar` role is not a live region, screen readers won't announce updates to those values as they change. The use of an ARIA progress bar is not actually important for this technique – what matters is the progress being conveyed with a status message.

The initial implementation of these progress bars were semantically correct but extra work is required to announce the progress.

## Definition of Done

Enable VoiceOver or JAWS. Import a file or clone a book (the larger the better), generate a cover, or export a book. Percentages should be announced periodically.

**NOTE:** I had to set throttling to 3G in the Network tab of Chrome Developer Tools in order to hear more than one announcement for most of these as the operations on my local dev instance are quite fast.

### DIFF
--- a/assets/src/scripts/cloner.js
+++ b/assets/src/scripts/cloner.js
@@ -14,6 +14,7 @@ jQuery( function ( $ ) {
 		const button = $( '#pb-cloner-button' );
 		const bar = $( '#pb-sse-progressbar' );
 		const info = $( '#pb-sse-info' );
+		const status = $( '#pb-sse-status' );
 		const notices = $( '.notice' );
 
 		// Init clock
@@ -53,11 +54,12 @@ jQuery( function ( $ ) {
 			switch ( data.action ) {
 				case 'updateStatusBar':
 					bar.val( parseInt( data.percentage, 10 ) );
+					status.html( `${ data.percentage }%` );
 					info.html( data.info );
 					break;
 				case 'complete':
 					evtSource.close();
-					$( window ).unbind( 'beforeunload' );
+					$( window ).off( 'beforeunload' );
 					if ( data.error ) {
 						bar.val( 0 ).hide();
 						button.attr( 'disabled', false ).show();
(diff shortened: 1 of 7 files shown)


arzola (approved):
Nice job, it works as expected

PR #4480, opened 2026-06-04
fix: include src/js assets for distribution

arzola:
This pull request makes a small update to the `.distignore` file. The change refines how files in the `assets/src` directory are ignored during distribution, now ignoring all files except the `assets/src/scripts` directory.


richard015ar (approved):
LGTM

PR #4485, opened 2026-06-09
fix: add_filter/add_action corrections in hooks files

thisismyurl:
## What

Two related hook-registration defects across `hooks-admin.php` and `hooks.php`. Both are provable by reading the surrounding code — no speculation involved.

---

### 1. `hooks-admin.php` — `add_action` on a filter hook (line ~100)

`'gettext'` is a WordPress **filter** fired via `apply_filters()`. The two registrations immediately below it (`gettext_with_context`, `ngettext`) correctly use `add_filter()`. The `sites_to_books` callback explicitly returns `$translated_text`, confirming it is filter-shaped.

With `add_action()`, the return value is discarded and the string substitution never fires.

Also corrects `accepted_args` from `20` to `3`: the `gettext` filter passes exactly three arguments (`$translation`, `$text`, `$domain`). The value `20` was harmless but inaccurate.

**Before:**
```php
add_action( 'gettext', '\Pressbooks\Admin\Laf\sites_to_books', 3, 20 );
```
**After:**
```php
add_filter( 'gettext', '\Pressbooks\Admin\Laf\sites_to_books', 3, 3 );
```

---

### 2. `hooks.php` — `add_filter` on action hooks (7 lines)

Seven callbacks on the `'init'` action were registered with `add_filter()`. All seven are action-shaped: four return nothing, two return object instances whose return value serves no purpose as a hook callback, one is declared `void`. Every other `'init'` registration in the same file uses `add_action()`.

Also removes the spurious `accepted_args=2` from the `BookDirectory::init` registration — the `'init'` action passes no extra arguments.

---

## Testing

1. Verify the `sites_to_books()` string substitution fires correctly on the Plugins page.
2. Confirm all `init` callbacks (rewrite rules, error handler, privacy content) still execute on page load with `WP_DEBUG` enabled and no PHP warnings.

---
*Development and testing assisted by AI. All code reviewed and verified manually.*

### DIFF
--- a/hooks-admin.php
+++ b/hooks-admin.php
@@ -97,7 +97,7 @@ if ( is_main_site() && is_network_admin() ) {
 }
 
 // Replace strings
-add_action( 'gettext', '\Pressbooks\Admin\Laf\sites_to_books', 3, 20 );
+add_filter( 'gettext', '\Pressbooks\Admin\Laf\sites_to_books', 3, 3 );
 add_filter( 'gettext_with_context', '\Pressbooks\Admin\Laf\sites_to_books_with_context', 3, 4 );
 add_filter( 'ngettext', '\Pressbooks\Admin\Laf\sites_to_books_ngettext', 3, 5 );
 
--- a/hooks.php
+++ b/hooks.php
@@ -234,9 +234,9 @@ if ( is_admin() === false ) {
 // Redirects
 // -------------------------------------------------------------------------------------------------------------------
 
-add_filter( 'init', '\Pressbooks\Redirect\rewrite_rules_for_format', 1 );
-add_filter( 'init', '\Pressbooks\Redirect\rewrite_rules_for_catalog', 1 );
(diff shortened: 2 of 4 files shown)


fdalcin (approved):
Thanks for this PR, @thisismyurl. 

Both fixes are good changes. 

A few notes:

**XHTML title escaping**

A raw & in the site name would cause issues with PDF and EPUB exports. I've added a regression test that exercises the full XHTML export pipeline with a site name containing &, <, and > characters.

**Hooks cleanup** 

One factual note: the message states that using `add_action()` on the `gettext` filter "discards the return value and the string substitution never fires." In WordPress, `add_action()` is a thin wrapper around `add_filter()` — both register into the same `$wp_filter` array. This is a semantic correctness improvement, not a functional bug fix. 

The `accepted_args` correction and the removal of the spurious `accepted_args=2` on `BookDirectory::init` are real improvements though.

Approving and merging.

PR #4507, opened 2026-06-29
fix: keep reset URL in password reset email when no IP line is present

### LINKED ISSUE #4469: Password reset email does not contain URL
When a network manager requests a password reset from the user-edit page, that email does not contain the expected reset URL

Image

Image

This was first reported by a network manager and was reproduced on other networks.

thisismyurl:
Hey team,

Fixes #4469.

### Root cause
`remove_ip_from_password_reset_email()` (added in #4435) pops the last line of the message, assuming it is always the IP-address line. WordPress only appends that line when the requester is logged out (the `! is_user_logged_in()` branch of `retrieve_password()` in `wp-includes/user.php`). When a logged-in network manager requests a reset from the user-edit screen there is no IP line, so the reset URL — the real last line — was removed and the email arrived with no link. That matches the report.

### Fix
Remove only the IP-address line, identified by WordPress's own translatable sentence (`This password reset request originated from the IP address %s.`), wherever it appears, instead of blindly popping the last line. When there is no IP line, nothing is removed and the URL is preserved.

### Testing
- Existing `test_remove_ip_from_password_reset_email` still passes (IP line present → removed, URL kept).
- Added `test_remove_ip_from_password_reset_email_preserves_url_when_no_ip_line` as a delete-the-fix guard: a message with no IP line must keep the reset URL — with the previous `array_pop()` it fails.

(full disclosure: AI helped me identify the issue and verify my work)


arzola (approved):

tw77:
Thanks for this fix – it works as expected and is ready for inclusion in the next Pressbooks release.

PR #4508, opened 2026-06-30
fix: fall back to editors for copyright when book has no authors

### LINKED ISSUE #4324: Copyright information in Metadata has no copyright holder when book has no author but only editor
### Description

When a book has an author or authors their name(s) are automatically added as copyright holders in the copyright notice in the metadata. If a book does not have an author but only an editor, no copyright holder information is in the copyright notice in the metadata, but the "by" still appears.

Book metadata with an author:

Image

Book metadata without an author:

Image

### Steps to Reproduce

1. Remove all contributors from the Authors field in a book
2. Leave the copyright holder field in Book info blank
3. View the book's homepage

**Expected behavior:** Pull the copyright holder information from the Editors field?

**Actual behavior:** No copyright holder name appears but the "by" in the text remains.

fdalcin:
Resolves #4324 

### Problem

When a book has authors, their names are automatically used as the copyright holder in the copyright notice. But when a book has **only editors** and the copyright-holder field is left blank, no holder name appears in the notice — while the "by" still renders, leaving a dangling "by" with no name.

### Steps to reproduce

1. Remove all contributors from the **Authors** field of a book
2. Leave the **Copyright holder** field in Book Info blank
3. View the book's homepage

**Before:** `… Copyright © by  is licensed under …` — empty holder, dangling "by".
**After:** the editor's name is used as the copyright holder.

### Changes

- **`inc/class-licensing.php`** — `doLicense()` now adds `pb_editors` as the next fallback in the copyright-holder precedence chain:
  *section author → `pb_copyright_holder` → `pb_authors` → `pb_editors`*.
  The `pb_copyright_holder`/`pb_authors` checks were switched from `isset()` to `! empty()` so a blank-but-present field correctly falls through to the next source.
- Extracted contributor normalization into a new private `contributorsToString()` helper, reused for both authors and editors.

### Tests

Added `test_doLicenseFallsBackToEditors` covering:
- editors used as the holder when there's no author/copyright holder (string and array forms)
- multiple editors rendered together
- authors still taking precedence over editors

### DIFF
--- a/inc/class-licensing.php
+++ b/inc/class-licensing.php
@@ -8,7 +8,7 @@ namespace Pressbooks;
 
 use function \Pressbooks\Utility\debug_error_log;
 use function \Pressbooks\Utility\explode_remove_and;
-use function \Pressbooks\Utility\implode_add_and;
+use function \Pressbooks\Utility\get_contributors_name_imploded;
 
 /**
  * TODO: Refactor
@@ -303,17 +303,7 @@ class Licensing {
 			return $contributors;
 		}
 
-		$names = [];
-
-		foreach ( $contributors as $contributor ) {
-			if ( empty( $contributor['name'] ) ) {
-				continue;
-			}
-
-			$names[] = $contributor['name'];
-		}
-
-		return implode_add_and( ';', $names );
+		return get_contributors_name_imploded( $contributors );
 	}
 
 	/**
--- a/inc/class-licensing.php
+++ b/inc/class-licensing.php
@@ -252,20 +252,15 @@ class Licensing {
 		if ( ! empty( $section_author ) && ! empty( $section_license ) ) {
 			// section author higher priority than book author when there's a custom license
 			$copyright_holder = $section_author;
-		} elseif ( isset( $metadata['pb_copyright_holder'] ) ) {
+		} elseif ( ! empty( $metadata['pb_copyright_holder'] ) ) {
 			// book copyright holder higher priority than book author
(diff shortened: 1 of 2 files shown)


inc/class-licensing.php
```diff
@@ -293,6 +288,34 @@ public function doLicense( $metadata, $post_id = 0, $title = '' ) {
...
+	 * Normalize a contributor metadata value into a display string.
+	 *
+	 * Contributor metadata can be either a pre-formatted string or an array of
+	 * entries, depending on how the book information was loaded.
+	 *
+	 * @param string|array{name: string} $contributors
+	 *
+	 * @return string
+	 */
+	private function contributorsToString( string|array $contributors ): string {
```
richard015ar, line 301: I think we can use this method:
https://github.com/pressbooks/pressbooks/blob/b0b4969deb149844758ddb1960816ef189f97fb1/inc/utility/namespace.php#L1277-L1285

instead to avoid duplication. What do you think?

I was thinking that maybe we should change `isset()` by `empty()` in that method, but I am not sure if that'd satisfy.

richard015ar (changes requested):

inc/class-licensing.php
fdalcin: I think we can be used, yes. We need to tweak it a little bit because this method does not accept strings. I’ll tweak it tomorrow morning :)

Not sure if we need to change the `isset` though. I think it’s fine as is.

richard015ar (approved):
LGTM

PR #4513, opened 2026-07-06
fix: global duet datepicker for proper enqueue

### LINKED ISSUE #4512: Register the Duet Date Picker as a reusable handle in core Pressbooks
We currently make use of an external date-picker package in core Pressbooks (on the book info page) and in several plugins (like LTI and pressbooks-network-analytics). Currently, core imports @duetds/date-picker and defines the duet-date-picker custom element, but only inside the [book-information.js bundle](https://github.com/pressbooks/pressbooks/blob/dev/assets/src/scripts/book-information.js). The element definition is isolated to that one page context and never registered as a reusable WordPress script/style handle.

When downstream plugins (like pressbooks-network-analytics) try to use the date picker on different pages, they have two options, neither of which are ideal:

1. Duplicate the import: add the dependency and define it themselves (wasteful). This is what the LTI plugin does now.
1. Try to enqueue a handle [wp_enqueue_script( 'duet-date-picker' )]. This will fail silently because core never registered that handle, unless another plugin that does defines the element globally (like LTI) is activated. This is what pressbooks-network-analytics does now. 

Ideally, core would register duet-date-picker as a proper WordPress handle at plugin initialization, making it available globally for any plugin to enqueue, regardless of which page context they're on. I think we'll want to create a dedicated duet-date-picker.js entry point, add it to vite.config and register it in PHP. Both the JS and CSS (if needed) should be built and registered as separate handles to give plugins the flexibility to enqueue either or both. We could then clean up all downstream usages, like in LTI, PB Results, pressbooks-network-analytics, etc

arzola:
The Duet date picker was only defined inside the Book Info page bundle, so any other page or plugin that wanted it had to ship its own copy or hope another active plugin had already defined the element. This registers `duet-date-picker` as proper, globally-available WordPress script and style handles at plugin init, so any plugin can enqueue the picker on any page — admin or front end — without duplicating the dependency.

### What changed

- Added a dedicated entry that defines the `duet-date-picker` custom element, plus a standalone theming stylesheet, built and registered as separate, stable handles so consumers can enqueue the script, the styles, or both.
- Registered the handles globally on `init`, so they're available in every context rather than only on the Book Info screen.
- Book Info now depends on the shared handle instead of bundling its own picker, shrinking that bundle to a fraction of its previous size.

### How to test

1. On a book, open Book Info and confirm the publication date pickers still open, are styled, and behave as before.
2. On another page that uses the picker (admin or front end), confirm `<duet-date-picker>` renders and is interactive.
3. With picker-using plugins active, confirm there are no "already defined" errors in the console.

Closes #4512


richard015ar (approved):
LGTM

PR #4517, opened 2026-07-09
chore: consolidate network managers helper

arzola:
I noticed we have this helper function repeated across plugins I think it's better to extract it and make it available from core


fdalcin (approved):
LGTM

PR #4519, opened 2026-07-13
fix: correct activation page login URL

### LINKED ISSUE #4504: Broken link on user activation page
### Description

When a new network user is created by adding them to a book, they receive an email with the link to activate their account. On the page that displays their credentials, there is a link "Log in". That link is broken.

([Video for Pressbooks staff only](https://us02web.zoom.us/clips/share/NegvnVEgTP2RLXUChyVAeg))

### Steps to Reproduce

1. Invite a new network user to a book
2. Have that user click on the URL in the email
3. On the screen with the credentials, have them click on "Log in"

**Expected behavior:** They should be taken to `[networkurl]/[bookurl]/wp-login.php`

**Actual behavior:** They are taken to `[networkurl]/wp/[bookurl]/wp-login.php`

oiahoon:
## Summary

- apply the existing Bedrock login URL normalization to WordPress's `login_url` filter
- add regression coverage for the activation-page login link
- update the callback comment to reflect both login and lost-password URLs

## Root cause

Pressbooks already removes the extra `/wp/` prefix from lost-password URLs, but the callback was registered only on `lostpassword_url`. WordPress's activation page switches to the new book and calls `wp_login_url()`, which uses the separate `login_url` filter.

## User impact

After activating an invited user account, the `Log in` link now points to the book's actual `wp-login.php` URL instead of the broken `/wp/<book>/wp-login.php` path.

## Testing

- `vendor/bin/phpunit --configuration phpunit.xml --group Login` — 2 tests, 4 assertions
- repository PHPCS ruleset across `inc/` and `bin/`, plus the changed top-level hook file
- PHP syntax checks for all three changed files
- `git diff --check`

The non-integration suite was also started locally; its first 136 of 1,053 tests passed before the run was stopped because PHP 8.4 produced extensive deprecation output from existing dependencies. The pull request CI can run the complete supported matrix.

Fixes #4504

### DIFF
--- a/inc/registration/namespace.php
+++ b/inc/registration/namespace.php
@@ -357,7 +357,7 @@ function check_for_strong_password( $pwd ) {
 }
 
 /**
- * Remove unwanted and unnecessary Bedrock's prefix in Lost Url link
+ * Remove unwanted and unnecessary Bedrock's prefix in login-related URLs
  * only if the book name is not wp
  */
 function remove_wp_prefix( $lostpassword_url ) {
--- a/inc/registration/namespace.php
+++ b/inc/registration/namespace.php
@@ -360,8 +360,8 @@ function check_for_strong_password( $pwd ) {
  * Remove unwanted and unnecessary Bedrock's prefix in login-related URLs
  * only if the book name is not wp
  */
-function remove_wp_prefix( $lostpassword_url ) {
-	return preg_replace( '/(.*[^\/])\/wp\/(.*)(\/wp-login.php)(.*)/i', '$1/$2$3$4', $lostpassword_url );
+function remove_wp_prefix( $url ) {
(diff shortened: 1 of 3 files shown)


inc/registration/namespace.php
```diff
@@ -357,7 +357,7 @@ function check_for_strong_password( $pwd ) {
 }
 
 /**
- * Remove unwanted and unnecessary Bedrock's prefix in Lost Url link
+ * Remove unwanted and unnecessary Bedrock's prefix in login-related URLs
```
arzola, line 360: @oiahoon considering we are reusing this callback to not only handle lost password urls could you change the variable name? `$lostpassword_url` to maybe just `$url`

arzola (changes requested):

inc/registration/namespace.php
oiahoon: Addressed in b6770577. Renamed the callback parameter to $url to match its shared login and lost-password use.

arzola (approved):

PR #4522, opened 2026-07-16
feat: cloning hooks and removing candela citations handler

SteelWagstaff:
Check if Candela Citations is active in source book before cloning, if so, attempt to activate in target book. Partial fix for https://github.com/pressbooks/private/issues/2603. It's very difficult to test this locally because it involves changes to the cloning routine. I need help putting this on a dev network (steeldev.pressbooks.network gets shut down after hours and on weekends, which is when I'm most likely to be able to work on this) or setting up something like ngrok for my local.

To test:
1. Check out this branch and make sure that Candela Citation is network deactivated on your test network
2. Go to a book on another network, activate the Candela Citation plugin and add at least one Candela Citation to front matter, back matter, or a chapter
3. Clone that book to your test network
4. Ensure that Candela Citation was activated as part of the clone and that all citations were included in the cloning process 

## Use of AI Tools
AI assistance: Yes
Tools: GitHub Copilot Pro (integrated in IDE), Skills provided by obra/superpowers
Models: Claude Haiku 4.5
Used for: Writing tests. Code changes, comments, PR description were written by me.


inc/cloner/class-cloner.php
```diff
@@ -652,6 +665,17 @@ public function clonePreProcess() {
 		if ( $this->sourceHasH5pApi ) {
 			$this->targetHasH5pApi = $this->h5p->activate();
 		}
+
+		// Candela Citations
+		if ( $this->sourceHasCandelaCitations ) {
+			$candela_citation_plugin = 'candela-citation/candela-citation.php';
+			if ( is_plugin_active_for_network( $candela_citation_plugin ) || is_plugin_active( $candela_citation_plugin ) ) {
```
SteelWagstaff: I'm not sure whether all these checks are strictly necessary. It might be simpler just to check if the target network has the CC plugin and attempt to activate it. I was just being cautious. To be honest, I really don't like checking for a deprecated third party plugin in core at all, so hopefully this doesn't have a long lifespan in our project.

inc/cloner/class-cloner.php
```diff
@@ -786,6 +810,12 @@ public function clonePostProcess() {
 			// Add a notice to the user indicating that the H5P could not be cloned
 			\Pressbooks\add_notice( __( 'The source book contained H5P content that could not be cloned. Please review the cloned version of your book carefully, as missing H5P content will be indicated. You may want to remove or replace these elements.', 'pressbooks' ) );
 		}
+
+		// Candela Citations
+		if ( $this->sourceHasCandelaCitations && ! $this->targetHasCandelaCitations ) {
+			// Add a notice to the user indicating that the Candela Citations could not be cloned
+			\Pressbooks\add_notice( __( 'The source book contained Candela Citations that could not be cloned because the Candela Citations plugin could not be activated on the new book. Contact your network manager to ensure this plugin is available and retry the cloning operation, or find another way to display the missing citations in your book.', 'pressbooks' ) );
```
SteelWagstaff: Notice to display to users if the source had CC but we could not activate Candela Citations in the target book for any reason.

inc/cloner/class-cloner.php
```diff
@@ -847,6 +877,27 @@ public function buildListOfKnownH5P( $url ) {
...
+	 * @return bool True if the source book has at least one post with Candela Citations data
+	 */
+	public function sourceBookHasCandelaCitations() {
+		foreach ( [ 'front-matter', 'chapter', 'back-matter' ] as $post_type ) {
+			if ( empty( $this->sourceBookStructure['_embedded'][ $post_type ] ) ) {
+				continue;
+			}
+			foreach ( $this->sourceBookStructure['_embedded'][ $post_type ] as $post ) {
+				if ( ! empty( $post['meta']['_candela_citation'] ) ) {
+					return true;
```
SteelWagstaff: We don't need to create a list of citations -- as soon as we find one citation, we can exit early because we will need to activate in the target book.

cmurtagh: Could we also first test if the plugin is active on the source book? That might be an easier test. Then if it is active, we look for citations.

SteelWagstaff: Would love to do this, but I don't know of a good way to check for active plugins on a remote network. The target book will not generally have access to the presence and activation status of books on the source network via API, AFAIK.

SteelWagstaff: WordPress does have a plugins endpoint that allows checking status (https://developer.wordpress.org/rest-api/reference/plugins/), but AFAIK it is *not* a public discovery endpoint. Core checks current_user_can( 'activate_plugins' ) before allowing reads of plugin data. I could be wrong, but I don't think our cloning tool has the permissions needed on the source network to make this check.

inc/cloner/class-cloner.php
```diff
@@ -847,6 +877,29 @@ public function buildListOfKnownH5P( $url ) {
...
+	 */
+	public function sourceBookHasCandelaCitations() {
+		foreach ( [ 'front-matter', 'chapter', 'back-matter' ] as $post_type ) {
+			if ( empty( $this->sourceBookStructure['_embedded'][ $post_type ] ) ) {
+				continue;
+			}
+			foreach ( $this->sourceBookStructure['_embedded'][ $post_type ] as $post ) {
+				$meta_value = $post['meta']['_candela_citation'] ?? null;
+				// Ignore null, empty arrays, empty strings, and serialized empty values
+				if ( ! empty( $meta_value ) && $meta_value !== '[]' ) {
```
SteelWagstaff: @arzola @cmurtagh I improved the check step here so that it doesn't consider empty arrays or null values evidence of use. This is based on what Christopher found in database investigation this morning.

richard015ar (approved):
LGTM

PR #4527, opened 2026-07-21
feat: add crawler filter and pressbooks_tracking schema changes

richard015ar:
Issue: https://github.com/pressbooks/private/issues/2629

Download counts in `pressbooks_tracking` include HEAD probes, scripts, and self-announced crawlers — anyone who hits `/open/download?type=…` increments a row. This first step toward https://github.com/pressbooks/private/issues/2596 filters those out at insert time so the numbers reflect actual human-driven downloads. No data is stored about who downloaded what: the user agent is read only to classify the request and is never persisted. Token-gating, dedup, and edge blocking stay out of scope — those land with the homepage-origin/token discussion.

### What changed
- **Insert-time filters in `BookDownload`** — requests are skipped unless they are GET with a non-empty, non-bot user agent. Files still get served to everyone, so only the counting changes and direct links (including OPDS consumers) keep working unchanged.
- **Bot detection via Crawler-Detect** (`jaybizzle/crawler-detect:^1.4`). The original plan proposed `matomo/device-detector` (already in core), but tests proved its bot list misses `python-requests`, `curl`, `Wget`, and other HTTP libraries — the self-announcing scrapers showing up in the access-log evidence. Crawler-Detect covers both named crawlers and HTTP libraries. device-detector stays for the diagnostics page (different purpose).

### How to test
1. `curl -I '<book-url>/open/download?type=pdf'` (HEAD): file headers OK, no row inserted.
2. `curl '<book-url>/open/download?type=pdf' -o /dev/null` (default curl UA): file serves, no row inserted.
3. `curl -A 'Mozilla/5.0 (KHTML, like Gecko) Chrome/126.0' '<book-url>/open/download?type=pdf'`: file serves, row inserted — faked-browser UAs pass the filter; this is a known limitation. 
4. Go to a book home page and download any book format. It should also insert a row.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,8 @@
 		"symfony/process": "^6.0|^7.0",
 		"vanilla/htmlawed": "^2.2",
 		"vlucas/phpdotenv": "^5.4",
-		"h5p/h5p-core": "1.27.0"
+		"h5p/h5p-core": "1.27.0",
+		"jaybizzle/crawler-detect": "^1.4"
 	},
 	"require-dev": {
 		"dms/phpunit-arraysubset-asserts": "^0.5.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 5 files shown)


arzola (approved):
LGTM works as expected

PR #4528, opened 2026-07-22
feat: canonical network title

### LINKED ISSUE #4488: Clarify difference between 'Network Title' and 'Site Name'
Pressbooks is a WordPress multisite. WP multisites have a setting called 'Network Title' which is stored in the wp_sitemeta with metakey 'site_name'. This value is set when the network is first created and can be changed by super admins from the network settings page `https://NETWORK.URL/wp-admin/network/settings.php`. This setting is unavailable to network managers. I believe this value is only used with dynamic registration / app name in LTI registration flows and when sending transactional emails. We also attempt to reference and display this value for Directory fetcher calls (but we actually use the network name from the BI dashboard, AFAIK). This is related to https://github.com/pressbooks/bi-data-manager/issues/663

Image

Each network also has a 'root site' -- this site uses the Aldine theme and displays the network home page. The root site also has a site title. This value is stored in wp_options with metakey 'blogname'. This value is displayed prominently on the network home page. Super admins and network managers can change this value via the 'Site Title' field in the 'Site Identity' option in the Aldine theme customizer `https://NETWORK.URL/wp-admin/customize.php` 

Image

In most cases, the network title and root site title values are similar (at least when the network is created). Network managers can easily change the site title for the root site, but cannot see or change the network title for the network itself. We need to make it more clear what each of these values do and provide a consistent, easy to understand path for changing (and keeping track) of the 'network title' and/or 'site title' for each client network.

There are many reasons why we might want to keep track of a network title for internal purposes. Our Directory displays a network title/name for each source of books (this comes from the networks table in Retool, which is _mostly_ based on the network title, but is not currently updated when network titles are changed.

## Recommendation

Treat the root site 'Site Title' as the canonical 'Network Name'. This is what what we should allow super admins/network managers to adjust (in customizer AND network options page (not currently displayed here), what we should use for transactional emails, what we should use as App name in LTI registration, and in the canonical networks list in BI Dashboard and Directory. We could either deprecate the 'Network title' and not use anywhere OR automatically update it to match the 'site title' value.

TODO:
- [x] Add 'Site Title' to network options page with label 'Network Name'
- [x] Ensure 'Site Title'/'Network name' is used in transactional emails
- [x] Ensure 'Site Title'/'Network name' is used in LTI registration flow
- [x] Ensure 'Site Title'/'Network name' is collected and used by BI Dashboard (networks table)
- [x] Ensure 'Site Title'/'Network name' is used/displayed/updated by the Directory 
- [x] Ensure old 'Network title' is not used/referenced anywhere 
- [ ] Optional: ensure that Network title is automatically updated when 'Network name' is updated?

arzola:
Resolves #4488 

Keeps the network's site name in sync with the main site's title. When you update the main site's blogname, the network `site_name` option is updated to match, the redundant Site Title field on the network settings screen is hidden, and any attempt to save `site_name` directly is overridden to fall back to the main site's blogname — so there's a single source of truth for the network title.

## What changed

- Added `syncSiteName()` on `NetworkSettings`, hooked to `update_option_blogname`, to update the network `site_name` when the main site's blogname changes.
- Added `overrideSiteName()`, hooked to `pre_update_site_option_site_name`, forcing `site_name` to the main site's blogname (falls back to the incoming value when blogname is empty).
- Added `hideSiteTitle()`, hooked to `admin_head-settings.php`, to hide the now-redundant Site Title field on the network settings screen.
- Registered the three hooks in `hooks-admin.php`.
- Added tests covering sync, override, fallback, and the hide-field output.

## How to test

1. In a multisite/network install, go to the main site's **Settings → General** and change the **Site Title**.
2. Visit **Network Admin → Settings** and confirm the network name now matches the new title.
3. Confirm the **Site Title** field is hidden on the Network Settings screen.
4. Try saving Network Settings and confirm `site_name` stays in sync with the main site's blogname.
5. Go to customizer and edit the Site Title, the change should be reflected on Network Settings
6. Run the test suite: `test-networksettings.php`.


richard015ar (approved):
LGTM works as expected.
I edited and checked:
- Appareance > Customize Home Page (customizer)
- Settings > Network Options
- Settings > Root Site General Settings
Also, Settings > Network Settings has no options for Site Name as expected.

Issue #4533, opened 2026-07-29
Cloning description is slightly misleading


SteelWagstaff:
### Description
On our clone a book page, the description below the 'New Book URL' reads, in part: "This URL cannot be changed later, so choose carefully."

https://github.com/pressbooks/pressbooks/blob/63a83201fd50b01427becaec334cb3f9cc4dec36/templates/admin/cloner/page.blade.php#L17

<img width="604" height="316" alt="Image" src="https://github.com/user-attachments/assets/72789128-4ddc-48e7-aaf6-431b3c803ed1" />

This is untrue. Super admins/network managers _can_ in fact change the URL later. It's just regular book admins who cannot. It might be better to omit this sentence or modify it for greater accuracy. Something like: "This URL can only be changed later by a network administrator, so choose carefully."

This change was requested by network managers at the University of Arizona in a recent PUG meeting.

MichSpieler:
Alternative: "As a regular user you will not be able to change this book URL later, so choose carefully."

MichSpieler:
From a Support perspective it could be a good idea to bar network managers from seeing the "Edit book" page (or at least the 'Settings' tab on that page).

PR #4534, opened 2026-07-29
fix: clarify target book URL description

SteelWagstaff:
Update the description for the target book URL input to clarify that changes can only be made later by a network administrator. Fix for https://github.com/pressbooks/pressbooks/issues/4533

### DIFF
--- a/templates/admin/cloner/page.blade.php
+++ b/templates/admin/cloner/page.blade.php
@@ -1,4 +1,4 @@
-<div class="pb-cloner wrap">
+ <div class="pb-cloner wrap">
 	<div class="pb-cloner-section main">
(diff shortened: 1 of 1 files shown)


templates/admin/cloner/page.blade.php
```diff
@@ -1,4 +1,4 @@
-<div class="pb-cloner wrap">
+ <div class="pb-cloner wrap">
```
SteelWagstaff:
```suggestion
<div class="pb-cloner wrap">
```

SteelWagstaff:
@arzola based on today's conversation at bug scrub, support may want to the language to be adjusted. Please hold off review until @MichSpieler and @tw77 have weighed in on desired text.

PR #4538, opened 2026-08-03
fix: z-index google docs picker dialog

arzola:
This address Google Picker overlapping reported by @MichSpieler 

image

Result

image


fdalcin (approved):
LGTM

PR #4540, opened 2026-08-04
fix: use correct sitemap url

cmurtagh:
using home_url() instead of get_option('siteurl') and removing core WP sitemap functionality so it doesn't put the wrong URL in robots.txt


fdalcin (approved):
LGTM

Issue #4523, opened 2026-07-16
Fix Pressbooks for WordPress 7.x compatibility (admin styling + breaking changes)


arzola:
### As a

Network Manager

### I want

Pressbooks to run on WordPress 7.0 without regressions

### So that

I can keep my network on a supported, secure WordPress version

### Context

WordPress 7.x brings admin UI/styling changes and potential breaking API changes. Pressbooks heavily customizes wp-admin (book dashboard, Organize, Export, custom metaboxes) and leans on WP internals, so it needs an audit before networks upgrade.

### Acceptance Criteria

- [ ] Pressbooks admin screens (dashboard, Organize, Export, Themes, custom metaboxes) render correctly on WP 7.x with no visual breakage
- [ ] Deprecated or removed WP APIs used by Pressbooks are identified and updated
- [ ] Webbook front-end and export flows (PDF/EPUB) work on WP 7.x
- [ ] The automated test suite passes against WP 7.0
- [ ] A short compatibility note lists WP 7.x changes that required code updates or remain open risks

### Additional Context

**Technical Notes** — Focus on wp-admin CSS overrides (admin styling has likely shifted), enqueue/hook changes, and any block-editor/REST changes. Check the `pressbooks-aldine` root theme and shared admin assets if affected.

fdalcin:
## WordPress 7.0 Compatibility — Status Update

**Summary:** Pressbooks is compatible with WordPress 7.0. We reviewed the entire Pressbooks suite and the good news is that the large majority works on 7.0 with no changes needed. A small number of minor fixes were required, and those are ready.

### What we reviewed
We checked all Pressbooks components against WordPress 7.0 — the core plugin, every first-party plugin, and both McLuhan/Aldine themes. Each was run against a full WordPress 7.0 test environment.

### What we found
- **3 needed small fixes:**
  - **Core plugin:** WordPress 7.0 changed the look of some admin controls (buttons, form fields, list pagination) and added a few new menu items. We adjusted Pressbooks so the admin keeps its familiar Pressbooks appearance and isn't cluttered with WordPress features that don't apply to us.
  - **Two themes:** minor housekeeping to swap a couple of outdated WordPress functions for their current equivalents.
- **No functionality was broken**, and no blocking issues were found.

### Testing
Automated tests pass across the whole suite on WordPress 7.0. Visual/manual checks of the key admin screens are underway and looking good so far.

### Next steps
- Merge the three sets of fixes (core + two themes).
[...]

PR #4541, opened 2026-08-04
feat: WP 7.x support

fdalcin:
Issue #4523

## TL;DR
Makes the Pressbooks **core** plugin run on **WordPress 7.x** with no regressions. WP 7.0's admin restyle (control sizing, checkbox/radio, buttons, a CSS-custom-property color system) and two new admin pages (Font Library, Connectors) shifted the Pressbooks admin UI; this PR restores the pre-7.0 look and removes the unwanted new pages, plus one PHP null-safety fix. No hard-removed WordPress APIs were found in 7.0.3, so there are no fatal-error fixes — the work is admin-UI restoration plus menu hygiene. Verified live on a WP 7.0.3 network against a 6.9.5 reference and with the unit suite.

## What changed

**PHP / API**
- Made `normalize_css_urls()` null-safe (PHP 8.1+ `preg_replace_callback` deprecation) — with a regression test.

**New WP 7.0 admin pages removed - rendered blank / unwanted in Pressbooks**
- Removed the **Font Library** submenu (Appearance) and blocked direct access to `font-library.php` (403).
- Removed the **Connectors** submenu (Settings) and blocked direct access to `options-connectors.php` (403).
- Both mirror Pressbooks' existing "remove submenu + guard route" pattern, with tests.
- Hid the new admin-bar **Command Palette (⌘K) button** while keeping the ⌘K shortcut functional.

**Admin styling restored to 6.9.x (WP 7.0 restyle)** — all in one removable stylesheet, `assets/src/styles/wp7-controls-compat.scss`:
- Button and input **sizing** (`.button` + `-primary/-secondary/-large/-small/-hero`, text inputs, selects, textareas).
- **Checkbox/radio** styling (6.9.x white-box + brand-red check/dot, 4px radius) — scoped to *not* clobber Pressbooks' custom `.toggle` switches.
- **`.page-title-action`** (nav pills / "Add" buttons) sizing and background.
- **Secondary button** background (solid gray vs 7.0's transparent + theme-tint), with the primary button explicitly excluded so it keeps its red.

**Color scheme (WP 7.0 CSS custom properties)**
- Defined `--wp-admin-theme-color` (+ `--rgb`, `-darker-10/-20` and their `--rgb`) in both Pressbooks schemes (`colors-pb` = `#bb2026`, `colors-pb-a11y` = `#AD0000`). WP 7.0 controls read these; without them, radios/focus rings fell back to WordPress's default blue.

**Compatibility metadata**
- Reconciled the plugin header `Requires at least` (`6.8.3` → `7.0.3`) to match the minimum already enforced in `compatibility.php`. Support floor is held at 7.0.3 (fixes prefer APIs valid across 6.9→7.0).

**Tooling**
- Added a scoped PHPCS ruleset (`phpcs.wp7-deprecations.xml`) for auditing WP deprecations, kept isolated from normal linting.

## Testing
- Unit suite green on WP 7.0.3 (CI). Locally, the only failing tests are environmental — 3 require the external `dev.pressbooks.pub` host (whitelisted on CI) and the Google Docs import tests require their Composer dependencies installed — none are WP 7.0-related.
- Manual QA on a live WP 7.0.3 network compared screen-by-screen against a WP 6.9.5 reference (Dashboard, Organize, Export, Themes, Book Info, network screens, export/webbook flows).


pressbooks.php
```diff
@@ -8,7 +8,7 @@
  * x-release-please-start-version
  * Version: 6.43.0
  * x-release-please-end
- * Requires at least: WordPress 6.8.3
+ * Requires at least: WordPress 6.9.1
```
fdalcin: Do we want to bump this to 7.0 or are we allowing a range of WP versions?

arzola: Good question, I'm not entirely sure

I mean we usually support the latest WP version but this could still work on previous versions

arzola (approved):

PR #4544, opened 2026-08-06
fix: orphan http_import url handler and cleanup

arzola:
Connecting a Google account broke every "Import from URL" flow: the Google Docs field reused `name="import_http"`, so its empty value clobbered the real URL on submit and WXR/EPUB/HTML imports failed. This fixes the collision and drops the URL-handling code left over now that Google Docs imports via the Picker.

### What changed

- Google Docs now uses a dedicated hidden `import_gdoc_id` field instead of sharing `import_http`, so URL imports work again.
- The Picker stores the doc id directly and the server fetches by id; removed the URL round-trip (`OAuthClient::extractDocId` and its test).

### How to test

1. With Google **connected**, import an EPUB/WXR by URL — it should succeed (previously failed).
2. Pick a doc via **Select from Google Drive** and confirm the import runs.
3. Submit Google Docs with no doc picked — expect "Please select a Google Doc to import."

### DIFF
--- a/inc/modules/import/class-import.php
+++ b/inc/modules/import/class-import.php
@@ -531,11 +531,10 @@ abstract class Import {
 	 * @return bool
 	 */
 	static protected function setGoogleDocsImportOptions(): bool {
-		$url = sanitize_text_field( getset( '_POST', 'import_http' ) );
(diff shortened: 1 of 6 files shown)


fdalcin (approved):
LGTM

PR #4556, opened 2026-08-17
fix: epub validation error display

richard015ar:
Issue: https://github.com/pressbooks/pressbooks-lti/issues/485

Authors exporting EPUB sometimes received a contradictory validation email: the summary line reported "Errors: 2, Files affected: 2" while the formatted report below it said "Total Errors: 0 / Files Affected: 0". On top of that, the report replaced real epubcheck messages with canned English strings and threw away the file name, line number, and column — leaving authors with no actionable information to fix their book.

The root cause was two independent parsers that could disagree. The formatted report used a hard-coded whitelist of known error patterns (two specific RSC-005 messages, a handful of other codes), so any unrecognised error — including the `target="_new"` attribute violation that triggered this report — entered the right branch but matched no inner condition, landed in no bucket, and was silently dropped. The summary line counted raw `ERROR(` substrings and
was always right; the report was always wrong for unknown errors.

### What changed

- EPUB validation log parsing and rendering moved into a dedicated `EpubcheckLog` class that recognises any epubcheck severity line (`FATAL`, `ERROR`, `WARNING`, `INFO`, `USAGE`) and preserves the verbatim message, file, line, and column. The report and summary render from the same parsed data, so their numbers can never contradict each other, and nothing is ever silently dropped.
- Both the formatted report and the summary line now derive their numbers from the same parse result, so they can never contradict each other.
- The validation report now shows the real epubcheck output per issue — e.g. `[ERROR RSC-005] Line 15, Col 1045: value of attribute "target" is invalid; must be …` — giving authors what they need to actually fix the book.
- Errors in non-XHTML files (`content.opf`, CSS, etc.) are now visible in the report; previously the file regex was XHTML-only so those errors vanished.
- All report strings are now wrapped in `__()` with the `pressbooks` text domain and translator comments.
- The unused public formatValidationLog() helper was removed; validation reporting now runs entirely through the export's error-logging path.

### How to test

1. Export a book that contains a link with `target="_new"` somewhere in its content (you can add one temporarily to a chapter).
2. After the export, check the validation email that arrives for your user (you can use mailHog in your local, run `lando info` to see where the mailHog host is reachable.
3. Confirm the **Total Errors** count in the formatted report matches the **Errors** figure in the summary line — they should be identical.
4. Confirm the report body shows the actual epubcheck message with a file name, line number, and column number (e.g. `[ERROR RSC-005] Line 15, Col 1045: value of attribute "target" is invalid`), not a generic placeholder string.
5. For a clean book with no validation issues, confirm the email is not sent / the report shows zero across the board.

### DIFF
--- a/inc/modules/export/epub/class-epub.php
+++ b/inc/modules/export/epub/class-epub.php
@@ -2957,132 +2957,13 @@ class Epub extends Export {
 	}
 
 	/**
-	 * Parse an epubcheck validation log into a structured, lossless form.
-	 *
-	 * Handles both literal "\n" separators (as produced by validate()) and real
-	 * newlines. Every FATAL/ERROR/WARNING/INFO/USAGE line is captured with its
-	 * severity, code, file, line, column and verbatim message; lines without a
-	 * recognizable file are grouped under '(package)' so nothing is dropped.
-	 *
-	 * @param string $validation_log Raw validation log string
-	 * @return array{files: array<string, array<int, array{severity: string, code: string, line: int, column: int, message: string}>>, errors: int, warnings: int, files_affected: int}
-	 */
-	protected function parseValidationLog( string $validation_log ): array {
-		$lines = explode( "\n", str_replace( '\n', "\n", $validation_log ) );
-
-		$files = [];
-		$errors = 0;
-		$warnings = 0;
-
-		foreach ( $lines as $line ) {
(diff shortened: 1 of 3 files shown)


arzola (approved):
LGTM and works as expected btw nice and clean implementation

PR #4559, opened 2026-08-18
fix: race condition showing 'No chapters were selected' on import

arzola:
The random "No chapters were selected for import" error was a race between the two requests the import form fires. The form submits a POST that stores the chapter selection in a transient, and an EventSource GET reads it back to run the import but the JS opened the EventSource immediately instead of waiting for the POST (the handler was invoked inline and passed to a nonexistent `done` option). On a slow server the GET could win and read the transient before it was written, so the server saw an empty selection. The EventSource now opens only after the POST completes, which is what the original design intended.

### What changed

- The step-2 import submit opens the EventSource in the POST's `success` callback, so the chapter selection is always stored before the server reads it. This fixes the error for every import type, since the race was upstream of the importers.
- A failed POST now resets the progress UI and shows the connection error instead of leaving the progress bar spinning forever.

### How to test

The race only shows up when the POST is slow, so simulate it:

1. On `dev`, add `sleep( 3 );` just before the `set_transient( 'pressbooks_current_import_POST', ... )` call in `Import::formSubmit()` (STEP 2, `inc/modules/import/class-import.php`).
2. Import any file, select chapters, submit — you get "No chapters were selected for import." every time.
3. Switch to this branch (keep the sleep), repeat the import — it starts ~3 seconds later but imports the selected chapters successfully.
4. Remove the sleep and confirm a normal import still works.


richard015ar (approved):
LGTM

richard015ar:
PR LGTM. I just want to clarify that adding `sleep( 3 );` [here](https://github.com/pressbooks/pressbooks/blob/5b38975f7bab7b719f2949ea174dee87d8cdf08c/inc/modules/import/class-import.php#L262) didn't help to replicate the issue.
Importing with this branch didn't cause any issue, so it still works the same on my local. Hopefully it fixes on our dev and prod environments.

PR #4576, opened 2026-09-17
fix: gtag and h5p.video script conflict

### LINKED ISSUE #4563: H5P Question Set with Video does not display video when Collapse Sections is active
Collapse Sections seems to interfere with the video display in Question Set with a video as part of a Fill in the Blank sub-activity (when there are multiple sections at least).

## Display when Collapse Sections is turned off:
https://wisc.pb.unizin.org/diverserussian/chapter/music-review-vocabulary/

Image

## Display when Collapse Sections is turned on:

Image

We saw this here as well https://github.com/pressbooks/private/issues/2684, but with Multiple Choice as part of a Question Set.

## Browser: 
The issue was not reproducible in Brave
The issue was not reproducible for the network manager in Chrome on Windows 11, except in a private window; it was reproducible for me and the author in Chrome on a Mac
It was reproducible for me in Firefox, Safari on Mac

## Reproducing the issue:
Reproducing the issue on staging was not possible (H5P shortcode create errors)
[Reproducing on university](https://university.pressbooks.pub/michtestbookjuly/chapter/h5p-and-collapse-sections/): no issue with Multiplce Choice or Fill in the Blanks

richard015ar:
Closes: https://github.com/pressbooks/pressbooks/issues/4563

On webbooks where Google Analytics (GA4 / gtag) is enabled, YouTube videos embedded in H5P content — Interactive Video, Question Set with video, etc. — never render. The video area stays blank, no YouTube requests are ever made, and there is no error in the console. This PR makes Pressbooks preload the YouTube IFrame API itself so H5P can always build the player.

### Root cause

A conflict between GA's video engagement tracking and H5P's YouTube loader:

1. `gtag.js` defines `window.onYouTubeIframeAPIReady` early in the page lifecycle, for its own YouTube engagement events.
2. H5P's `H5P.Video` YouTube handler (`loadAPI()` in `h5p-video/scripts/youtube.js`) only injects `https://www.youtube.com/iframe_api` when `window.onYouTubeIframeAPIReady` is `undefined`. Otherwise it assumes "someone else is loading the API" and just hooks into that callback.
3. gtag defines the callback but never loads the API. H5P defers to gtag and also doesn't load it. `window.YT` never becomes defined, `new YT.Player()` is never called, and the video is never built.

Verified with a headless-browser trap: `window.onYouTubeIframeAPIReady` is first set from `googletagmanager.com/gtag/js`, H5P's `loadAPI()` then skips the injection, and blocking GA in the network layer makes the video build immediately (`window.YT = object`, iframe created, YouTube requests fire).

This explains every observation reported:

- **Only some networks** → only books with GA enabled hit it.
- **"It looks version-related"** → newer `H5P.Video` added a `$placeholder.is(':visible')` guard that delays player creation until *after* gtag has already claimed the callback, flipping a race that older `H5P.Video` versions used to win.
- **"It looks like the collapse-sections theme option"** → coincidental. Hiding content delays H5P's `create()` call, but the bug also reproduces without collapse (Question Set builds its videos before they're visible).
- **Not related to** [h5p/h5p-video#177](https://github.com/h5p/h5p-video/issues/177) / [#178](https://github.com/h5p/h5p-video/pull/178) — that bug is about a *built* player losing inline positioning; here the player is never built.

### Fix

On front-end requests where the H5P plugin is active and the current post actually contains an `[h5p]` shortcode, enqueue `https://www.youtube.com/iframe_api` ourselves. This guarantees `window.YT` loads, so H5P can build the player regardless of whether gtag has already grabbed `onYouTubeIframeAPIReady`. GA's video engagement tracking still works — this only ensures the API script actually loads.

The behavior is filterable via `pb_h5p_preload_youtube_api` for books that don't want the third-party request (e.g. privacy reasons, no YouTube-based H5P content).

### Changes

- `inc/interactive/class-content.php` — new `enqueueYouTubeApiForH5P()` method, registered on `wp_enqueue_scripts`. Guards: H5P plugin active, `pb_h5p_preload_youtube_api` filter allows it, and the current post contains an `[h5p]` shortcode.
- `tests/test-interactive-content.php` — unit test covering all three guards.

### How to test

1. On a book with the H5P plugin active, create/import an H5P activity that embeds a YouTube video (e.g. Interactive Video, or QuestionSet with a video). Reference chapter used during development: content ID 3, "Musical instruments" section.
2. Enable Google Analytics on the book.
3. Before this patch: open the chapter, expand the section — video area is blank, no `youtube.com/embed` request in the Network tab.
4. Apply this patch. Reload — page should include a `<script src="https://www.youtube.com/iframe_api">` tag, the video should render, and you should see multiple `youtube.com` requests.
5. Remove the H5P shortcode from the chapter → the API script is no longer enqueued for that page.
6. Add `add_filter( 'pb_h5p_preload_youtube_api', '__return_false' );` → the API script is not enqueued even on H5P chapters.


richard015ar:
There's a pre-existing test failing. I am considering the fix of that out of the scope of this task.

fdalcin (approved):
LGTM

